PR #3, opened 2021-01-04
feat: add NFT standard

### LINKED ISSUE #2: Non-fungible token standard
As per previous discussions ([see Paper doc](https://paper.dropbox.com/doc/Working-group-Stacks-Token-Standard--BCnkvBjpEg~uI25yo8oFT1J2AQ-Hmix9Te1J9x7qYmTvlrqZ)), @friedger will create a proposal for a non-fungible token standard.

I'm creating this to keep track of the conversations

friedger:
This PR adds a SIP for the Non-fungible standard.

It fixes #2


nft-standard.md
```diff
@@ -0,0 +1,64 @@
...
+
+```
+
+(define-trait stacks-token-nft-standard-v1
+  (
+    ;; Name, limited to 32 chars
+    (name () (string-ascii 32)) 
+
+    ;; Icon URL, limited to 2048 chars
+    (icon-url () (string-ascii 2048))
```
friedger, line 35: Should we change this to 400,000 to support base64 encoded images for a size up to 400kb

hstove: I'd say no - too much state, and adds complexity for clients. If anything, it might not even be desirable for many clients to show any arbitrary image, given the potential for abuse.

hstove (commented):
Thanks so much for putting this together @friedger!

I am pretty hesitant to support too many display-oriented functions in this standard. I am much more in favor of a separate trait that does a great job at returning display data for an NFT.

For example, you have a mandatory function for `icon-url (uint)`, but I can imagine many NFTs that wouldn't want this. For example, NFTs that represent ownership of an asset (like a house), or an NFT that denotes membership of something.

Plus, if we wanted to really support easily displaying NFTs, this seems too limited. Why not have `name`, `description`, etc.

I would love a standard that does the job, while supporting flexibility and easy implementation. In my opinion, we only need:

```
get-balance
transfer?
get-owner?
```

I am curious to hear your opinion on defining standard error codes for `transfer?` and `get-owner?`. I believe that could go a long way towards a good experience for clients.

hstove:
@friedger can you rename the file to `sips/sip-009-nft-standard.md`? This is in line with the existing SIPs, which I just ported in #4.

nft-standard.md
```diff
@@ -0,0 +1,75 @@
...
+(define-trait stacks-token-nft-standard-v1
+  (
+    ;; Name, limited to 32 chars
+    (name () (string-ascii 32)) 
+
+    ;; Icon URL for group of NFTs, limited to 2048 chars
+    (icon-url () (string-ascii 2048))
+    
+    ;; Icon URL per NFT , limited to 2048 chars
+    (icon-url (uint) (string-ascii 2048))
```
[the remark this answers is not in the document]
friedger: good point

nft-standard.md
```diff
@@ -0,0 +1,75 @@
...
+
+    ;; Token ID, limited to uint range
+    (last-token-id () uint) 
+
+    ;; number of tokens owned by user
+    (get-balance (principal) uint)
+
+    ;; Owner of given token identifier
+    (get-owner? (uint) (response principal (tuple (kind (string-ascii 32)) (code uint))))
```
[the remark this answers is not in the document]
friedger: yes, `(optional principal)`

nft-standard.md
```diff
@@ -0,0 +1,75 @@
...
+
+    ;; Transfer from to
+    (transfer? (principal principal uint) (response bool (tuple (kind (string-ascii 32)) (code uint))))
+
+    ;; Approve
+    (approve (principal uint) (response bool (tuple (kind (string-ascii 32)) (code uint))))
+
+    ;; Set approval for all
+    (set-approval-for-all (principal bool) (response bool (tuple (kind (string-ascii 32)) (code uint))))
```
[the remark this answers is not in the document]
friedger: I removed it.

friedger:
There is already a discussion about [error types](https://github.com/clarity-lang/reference/issues/33).

For `transfer?`, we could define an error kind for the native transfer:

`{kind: "nft-transfer-failed", code: from-nft-transfer}`

friedger:
@hstove We don't really need `get-balance`, do we?
However, we need `get-last-id` to ensure that it is iteratable

hstove:
@friedger agreed that we don't need `get-balance`, and `get-last-id` makes sense too. I think the error handling part is the only thing that might need some tweaking. I'm in favor of structured errors, but regardless of what we decide as the signature, can we come up with the standard error types for `transfer`?

Another thing that we should do, and make as a best practice for this kind of SIP, is to require an example implementation. A simple repo with the trait and an implementation would be perfect. Bonus points for (simple) JS code, especially to demonstrate error handling.

All opinions my own, of course - would love to get more eyes on this.

xelamade:
@hstove example implementations are a great idea. my thinking would be to create a monorepo under the org that holds all implemented SIP examples, though through drafts they would be held by the implementer.

hstove:
@xelamade Friedger added a reference implementation in this PR: https://github.com/friedger/clarity-smart-contracts/blob/master/contracts/sips/nft-trait.clar

hstove:
Hey @friedger - have been doing some reading around [ERC721](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md#specification) implementations. The standard includes a function `tokenURI` - defined in Solidity as `function tokenURI(uint256 _tokenId) external view returns (string);`.

This returns a URL which includes JSON metadata that clients can use to embed information about the NFT. I believe we should follow a similar practice.

What do you think about adding this function:

```
(metadata-url (uint) (string-ascii 256))
```

Additionally, it would likely be best to build off of the existing JSON schema used in the ERC721 standard:

```json
{
    "title": "Asset Metadata",
    "type": "object",
    "properties": {
        "name": {
            "type": "string",
            "description": "Identifies the asset to which this NFT represents"
        },
        "description": {
            "type": "string",
            "description": "Describes the asset to which this NFT represents"
        },
        "image": {
            "type": "string",
            "description": "A URI pointing to a resource with mime type image/* representing the asset to which this NFT represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive."
        }
    }
}
```

friedger:
@hstove I like `get-metadata-uri` or shorter `get-meta-uri`, similar to `get-owner`. Optionally, we can support `get-meta` that returns the meta data stored on-chain. 

Is the length long enough? Why 256?

Should we limit the asset meta data to images? I suggest using "uri" and "mime-type". 

We discussed immutability, therefore, I suggest to include a hash in the contract `(get-asset-hash (uint) (response (buff 64) uint))`

hstove:
Well, if this is a URI, do you think 256 is not long enough? It's not for storing on-chain data - this method _must_ return a resolvable URI (although it could be on http, ipfs, etc). Supporting `get-meta` is just more tricky, because then we have to specify a very strict format, that can't be easily extended.

For example, the JSON schema I linked is for image-based NFTs, but in ETH there are now more semi-formal standards for other NFTs - like 3d objects, videos, etc.

I also like simply using an off-chain pointer because there is no need to access this data on-chain.

> We discussed immutability, therefore, I suggest to include a hash in the contract (get-asset-hash (uint) (response (buff 64) uint))

This is an interesting idea - this would be a hash of the JSON metadata? I like it - it's light enough and clients can choose how they respond to an invalid hash. However, if you wanted immutability, you could just use an IPFS-resolving URI, which are immutable.

> (This is all about the meta data and should go in a separate section of the standard and it should be optional.)

Agreed that all the JSON schema stuff should be left out of this strict standard. I'm mainly advocating for the URI method, as that'll be quite important to support a decentralized ecosystem of NFT apps.

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,85 @@
+## Preamble
```
jcnelson, line 1: Hey @friedger, sorry for not getting to this sooner.  There's a couple of small formatting style changes that should be made here.

* The top-level section titles (Preamble, Abstract, etc.) should be H1, not H2.
* The Preamble section should not be a bullet list.  Each line should be its own paragraph.

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,85 @@
...
+
+- Sip Number: 009
```
jcnelson, line 3: Small style nitpicks:

* `SIP Number` <-- SIP should be all caps.
* `Created: 2020-12-10` should be `Created: 10 December 2020`.

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,85 @@
...
+
+Tokens are digital assets that are registered on the Stacks blockchain through a smart contract. A non-fungible token (NFT) is a token that is globally unique and that users can identify through its identifier. The smart contract that registers the NFTs defines a name for the group of NFTs.
+
+NFTs are enumerated, the id starts at 1 and the current last id is provided by the smart contract.
+
+## Specification
+
+NFT smart contract shall implement the trait defined at `ST314JC8J24YWNVAEJJHQXS5Q4S9DX1FW5Z9DK9NT.nft-trait` as well as satisfy the additional conditions.
+The trait has three functions:
+* `last-token-id` does not take any arguments and returns the highest number that is used as an identifier for any nft. This is the upper limit when iterating through all nfts.
```
jcnelson, line 34: Small consistency nitpick: NFT should be capitalized consistently in this document.

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,85 @@
...
+    ;; Transfer from to
+    (transfer (uint principal principal) (response bool (tuple (kind (string-ascii 32)) (code uint))))
+  )
+)
+```
+
+### Implementation rules
+
+1. Contracts must use a least one nft asset. A post condition with deny mode and without any nft condition about a changed owner must fail for `transfer` function calls.
+1. After a successfull call to function `transfer` the function `get-owner` must return the recipient of the `transfer` call as the new owner.
```
jcnelson, line 58: Typo: "successful".  Also, is this meant to be an ordered list or an unordered list? I see all items start with `1.`.

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,85 @@
...
+https://eips.ethereum.org/EIPS/eip-721
+https://www.ledger.com/academy/what-are-nft
+
+## Backwards Compatibility
+
+Not applicable
+
+## Activation
+
+This SIP is activated as soon as 5 contracts are deployed that are using the same trait that follows this specification.
```
jcnelson, line 78: Can you add a deadline (in block heights) by which this should happen?  It can be either a Stacks block height or a Bitcoin block height -- it's up to you.

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,85 @@
...
+1. For any call to `get-owner`, resp. `transfer` with an id greater than `last-token-id`, the call should return a response `none`, resp. failed transfer. 
+1. The following error codes are defined
+
+| function | error | description |
+|----------|-------|-------------| 
+|`transfer`|`{kind: "nft-transfer-failed", code: from-nft-transfer}`| Error if the call failed due to the underlying asset transfer. The code `from-nft-transfer` is the error code from the native asset transfer function|
+
+## Related Work
+
+https://eips.ethereum.org/EIPS/eip-721
```
jcnelson, line 69: Could you add a couple sentences here comparing/contrasting ERC-721 with this proposal?  I see that ERC-721 provides a much bigger programmatic interface than this proposal does.  Do you consider methods for querying the token symbol, URL, name, etc. to be out of scope for this SIP?  It's up to you whether or not they are; I'm just asking so that future readers will understand the difference.

sips/sips/sip-009-nft-standard.md
friedger: added tip height # 5000

friedger (commented):
Addresses some comments, more to come

pxydn:
could we consider adding an optional way to store and image or video representing the NFT on-chain? is this possible?

friedger:
> could we consider adding an optional way to store and image or video representing the NFT on-chain rather than a URI? is this possible?
You could use a data uri or a uri that reference onchain data where ever that is stored, maybe in an attachment?

hstove:
FYI I've added `get-token-uri` to the FT standard, which returns `(optional (string-utf8 256))`. I like `get-token-uri` as it's in line with the ERC standards (for both 20 and 721).

I'm not 100% tied to that name or response, but it would be great if both traits had the same spec for that function.

pxydn:
i'll look at potentially adding it to the NFT trait

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,93 @@
...
+https://www.ledger.com/academy/what-are-nft
+
+# Backwards Compatibility
+
+Not applicable
+
+# Activation
+
+This SIP is activated if 5 contracts are deployed that are using the same trait that follows this specification. This has to happen before Stacks tip #5000.
```
jcnelson, line 86: This block height has already passed.  Want to just set this to something really far into the future, like Bitcoin block 700,000, the way @hstove did for SIP 010?

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,93 @@
...
+(define-trait stacks-token-nft-standard-v1
+  (
+    ;; Token ID, limited to uint range
+    (last-token-id () (response uint uint))
+
+    ;; Owner of given token identifier
+    (get-owner (uint) (response (optional principal) uint))
+
+    ;; Transfer from to
+    (transfer (uint principal principal) (response bool (tuple (kind (string-ascii 32)) (code uint))))
```
muneebm, line 58: Should we follow the same order for the parameters as we have in ERC721, (from, to, id) instead of (id, from, to).
ERC721 signature:
` function transferFrom(address _from, address _to, uint256 _tokenId) external payable;`

muneebm:
I think the standard for meta data should also be part of this SIP. Maybe we can define a separate trait for that, something like this:
```
(define-trait sip-009-metadata
  (
    (get-token-uri (uint) (optional (string-utf8 2048)))
  )
)
```

EIP1155 lists some reasons for not including `name()` and `symbol()` in the ERC1155 standard, not sure if it's applicable for us: https://github.com/ethereum/EIPs/issues/1155

> Metadata Choices
> 
> The symbol function (found in the ERC-20 and ERC-721 standards) was not included as we do not believe this is a globally useful piece of data to identify a generic virtual item / asset and are also prone to collisions. Short-hand symbols are used in tickers and currency trading, but they aren't as useful outside of that space.
> 
> The name function (for human-readable asset names, on-chain) was removed from the standard to allow the Metadata JSON to be the definitive asset name and reduce duplication of data. This also allows localization for names, which would otherwise be prohibitively expensive if each language string was stored on-chain, not to mention bloating the standard interface. While this decision may add a small burden on implementers to host a JSON file containing metadata, we believe any serious implementation of ERC-1155 will already utilize JSON Metadata.

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,93 @@
...
+* `get-owner` takes an NFT identifier and returns a response containing the principal owning the NFT for the given identifier. The principal is wrapped as an optional, that means if the corresponding NFT does not exists the response is `(ok none)`, otherwise, e.g. `(ok (some 'ST12...))`. The owner can be a contract principal.
+* `transfer` takes an NFT identifier, a sender principal and a receiver principal. The function changes the ownership of the NFT for the given identifier. The change has to be reflected in the `get-owner` function, for details see implementation rules.
+
+## Trait
+
+```
+(define-trait stacks-token-nft-standard-v1
+  (
+    ;; Token ID, limited to uint range
+    (last-token-id () (response uint uint))
```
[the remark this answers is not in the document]
jcnelson: Two reasons:
* All `(define-public ..)` functions need to return a `response`
* Traits can't require the implementation to be `(define-read-only ..)`, and trait definitions only define public interfaces, so any trait definition must require the method return a `response`.  You can't create a Clarity trait for a function that doesn't return a `response` -- it won't pass analysis.

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,93 @@
...
+
+    ;; Owner of given token identifier
+    (get-owner (uint) (response (optional principal) uint))
```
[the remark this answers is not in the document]
jcnelson: Because this is a trait-defined function, it must return a `response`.  The inner `nft-get-owner?` method returns `(option principal)`, and this method's implementation would almost certainly just return whatever `nft-get-owner?` returns.

jcnelson:
I'll let @friedger decide what's appropriate here, but I'm generally in favor of having a set of "small" traits that are individually well-understood and incrementally deployed, versus a single "large" trait that may contain a lot of unnecessary boilerplate and would take a lot of deliberation to design and deploy.  You're probably right that we'll need something like ERC-1155 at some point, but it's not clear to me that it has to be in this SIP (but again, it's not my choice to make).

sips/sips/sip-009-nft-standard.md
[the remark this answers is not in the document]
muneebm: That makes sense, thanks for clarifying!

pxydn:
- adds better formatting
- grammar errors, typos
- stacks block 5000 was passed, changed to bitcoin block 700,000

uploaded as .txt as GitHub wouldn't allow an md file

[sip-009-nft-standard.txt](https://github.com/stacksgov/sips/files/6136461/sip-009-nft-standard.txt)

hstove:
@pxydn can you make those changes in git? I have no way of knowing what you've changed by simply looking at your `.txt` file. Here's the best way to do that:

- Make a fork of https://github.com/friedger/sips
- Make changes to `sips/sip-009-nft-standard.md`
- Commit those changes, push them to your fork
- Open a PR into _friedger's_ repo

Then, we can easily reason about the changes you've made, and Friedger can easily accept your PR, which will in turn make those changes visible here.

sips/sips/sip-009-nft-standard.md
hstove: I see and appreciate your reasoning here, but it clashes with consistency within Clarity / stacks:

- `nft-transfer?` uses `(identifier from to)`
- `ft-transfer`, `stx-transfer`, and the SIP10 `transfer` use `(amount from to)`

I would rather keep our own standards consistent than worry about strict compatibility with ERC's.

hstove:
Addressing other comments:

- I'd prefer to keep `get-token-uri` in this trait vs. in another trait, especially for NFTs. We've moved to activation for the FT SIP, and it includes `get-token-uri` in the spec. It doesn't add much overhead to implementation - if it's irrelevant to an implementation you can just return `(none)`, but almost every NFT in practice will point to some metadata.
- Removing `symbol` might make sense. Symbol is much less relevant to NFTs. I'm relatively ambivalent here.

sips/sips/sip-009-nft-standard.md
muneebm: I agree, it's better to keep this consistent with our own standards.

friedger:
I merged in @pxydn comments. Thank you!

sips/sips/sip-009-nft-standard.md
friedger: I have added some more details about EIP 721 and meta data decisions.

sips/sips/sip-009-nft-standard.md
```diff
@@ -71,6 +75,8 @@ The trait has three functions:
 |----------|-------|-------------| 
 |`transfer`|`{kind: "nft-transfer-failed", code: from-nft-transfer}`| Error if the call failed due to the underlying asset transfer. The code `from-nft-transfer` is the error code from the native asset transfer function|
 
+1. The methods `get-last-token-id`. `get-token-uri` and `get-owner` can be implemented as read-only functions.
```
hstove, line 78: Should be a comma (not period) after `get-last-token-id`

sips/sips/sip-009-nft-standard.md
```diff
@@ -38,7 +38,8 @@ NFTs are enumerated, the id starts at 1 and the current last id is provided by t
 
 NFT smart contract shall implement the trait defined at `ST314JC8J24YWNVAEJJHQXS5Q4S9DX1FW5Z9DK9NT.nft-trait` as well as satisfy the additional conditions.
 The trait has three functions:
-* `last-token-id` does not take any arguments and returns the highest number that is used as an identifier for any NFT. This is the upper limit when iterating through all NFTs.
+* `get-last-token-id` does not take any arguments and returns the highest number that is used as an identifier for any NFT. This is the upper limit when iterating through all NFTs.
+* `get-token-uri` takes an NFT identifier and returns a response containg a URI pointing to meta data of the given identifier. The URI is wrapped as an optional, which means if the corresponding NFT exists or the contract does not maintain meta data the response is `(ok none)`, otherwise, e.g. `(ok (some "https://example.com"))`. The length of the returned uri is limited to 256. The specification of the meta data should be covered in a separate SIP.
```
hstove, line 42: Metadata is one word

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,101 @@
...
+
+Created: 10 December 2020
+
+License: CC0-1.0
+
+Sign-off:
+
+# Abstract
+
+Non-fungible token are unique digital assets that are registered on the Stacks blockchain through a smart contract with certain properties.
```
kyranjamie, line 23:
```suggestion
Non-fungible tokens are unique digital assets that are registered on the Stacks blockchain through a smart contract with certain properties.
```

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,101 @@
...
+Ethereum has [ERP 721](https://eips.ethereum.org/EIPS/eip-721) that defined non-fungible tokens on the Ethereum blockchain. Notable differences are that the transfer function in EIP 721 uses a different ordering of the arguments ending with the token id. This SIP uses the token id as the first argument which is in line with all transfer functions in Clarity. Furthermore, this SIP only defines a URI pointing to meta data. There are not specifications about schema or other properties of token meta data. These should be specified in a separate SIP.
+
+
+# Backwards Compatibility
+
+Not applicable
+
+# Activation
+
+This SIP is activated if 5 contracts are deployed that are using the same trait that follows this specification. This has to happen before Bitcoin tip #700,000.
```
kyranjamie, line 94:
```suggestion
This SIP is activated if 5 contracts are deployed that use the same trait that follows this specification. This must happen before Bitcoin tip #700,000.
```

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,101 @@
...
+Created: 10 December 2020
+
+License: CC0-1.0
+
+Sign-off:
+
+# Abstract
+
+Non-fungible token are unique digital assets that are registered on the Stacks blockchain through a smart contract with certain properties.
+Users should be able to identify a single non-fungible token. Users should be able to own it and transfer it. Non-fungible tokens can have more properties that are not specified in this standard.
```
kyranjamie, line 24:
```suggestion
Users should be able to identify a single non-fungible token. Users should be able to own it and transfer it. Non-fungible tokens can have additional properties not specified in this standard.
```

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,101 @@
+# Preamble
+
+SIP Number: 009
+
+Title: Standard Trait Definition for Non-Fungibe Tokens
```
pxydn, line 5:
```suggestion
Title: Standard Trait Definition for Non-Fungible Tokens
```

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,101 @@
...
+
+NFTs are enumerated, the id starts at 1 and the current last id is provided by the smart contract.
+
+# Specification
+
+NFT smart contract shall implement the trait defined at `ST314JC8J24YWNVAEJJHQXS5Q4S9DX1FW5Z9DK9NT.nft-trait` as well as satisfy the additional conditions.
+The trait has three functions:
+* `get-last-token-id` does not take any arguments and returns the highest number that is used as an identifier for any NFT. This is the upper limit when iterating through all NFTs.
+* `get-token-uri` takes an NFT identifier and returns a response containg a URI pointing to meta data of the given identifier. The URI is wrapped as an optional, which means if the corresponding NFT exists or the contract does not maintain meta data the response is `(ok none)`, otherwise, e.g. `(ok (some "https://example.com"))`. The length of the returned uri is limited to 256. The specification of the meta data should be covered in a separate SIP.
```
pxydn, line 42:
```suggestion
* `get-token-uri` takes an NFT identifier and returns a response containing a URI pointing to metadata of the given identifier. The URI is wrapped as an optional, which means if the corresponding NFT exists or the contract does not maintain metadata, the response is `(ok none)` - otherwise, e.g. `(ok (some "https://example.com"))`. The length of the returned URI is limited to 256. The specification of the metadata should be covered in a separate SIP.
```

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,101 @@
...
+
+    ;; Transfer from to
+    (transfer (uint principal principal) (response bool (tuple (kind (string-ascii 32)) (code uint))))
+  )
+)
+```
+
+## Implementation rules
+
+1. Contracts must use a least one NFT asset. A post condition with deny mode and without any NFT condition about a changed owner must fail for `transfer` function calls.
```
pxydn, line 68:
```suggestion
1. Contracts must use a least one NFT asset. A postcondition with deny mode and without any NFT condition about a changed owner must fail for `transfer` function calls.
```

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,101 @@
...
+|`transfer`|`{kind: "nft-transfer-failed", code: from-nft-transfer}`| Error if the call failed due to the underlying asset transfer. The code `from-nft-transfer` is the error code from the native asset transfer function|
+
+1. The methods `get-last-token-id`. `get-token-uri` and `get-owner` can be implemented as read-only functions.
+
+# Related Work
+
+NFTs are an established asset class on blockchains. Read for example [here](https://www.ledger.com/academy/what-are-nft).
+
+## EIP 721
+Ethereum has [ERP 721](https://eips.ethereum.org/EIPS/eip-721) that defined non-fungible tokens on the Ethereum blockchain. Notable differences are that the transfer function in EIP 721 uses a different ordering of the arguments ending with the token id. This SIP uses the token id as the first argument which is in line with all transfer functions in Clarity. Furthermore, this SIP only defines a URI pointing to meta data. There are not specifications about schema or other properties of token meta data. These should be specified in a separate SIP.
```
pxydn, line 85:
```suggestion
Ethereum has [ERP 721](https://eips.ethereum.org/EIPS/eip-721) that defined non-fungible tokens on the Ethereum blockchain. Notable differences are that the transfer function in EIP 721 uses a different ordering of the arguments ending with the token id. This SIP uses the token id as the first argument which is in line with all transfer functions in Clarity. Furthermore, this SIP only defines a URI pointing to metadata. There are no specifications about schema or other properties of token metadata. These should be specified in a separate SIP.
```

sips/sips/sip-009-nft-standard.md
```diff
@@ -0,0 +1,101 @@
...
+
+Not applicable
+
+# Activation
+
+This SIP is activated if 5 contracts are deployed that are using the same trait that follows this specification. This has to happen before Bitcoin tip #700,000.
+
+# Reference Implementations
+
+Source code
```
pxydn, line 98:
```suggestion
## Friedger's clarity-smart-contracts
```

pxydn (commented):
Minor changes to grammar & fixes typos.

muneebm:
@friedger opened a [PR](https://github.com/friedger/sips/pull/1) in your repo. Please have a look.

sips/sips/sip-009-nft-standard.md
friedger: :+1:

sips/sips/sip-009-nft-standard.md
friedger: :+1:

sips/sips/sip-009-nft-standard.md
friedger: :+1:

sips/sips/sip-009-nft-standard.md
friedger: :+1:

sips/sips/sip-009-nft-standard.md
friedger: :+1:

sips/sips/sip-009-nft-standard.md
friedger: :+1:

sips/sips/sip-009-nft-standard.md
friedger: :+1:

sips/sips/sip-009-nft-standard.md
friedger: using `post-conditions` now as in docs.blockstack.org

jcnelson (approved):
Looks good to me!  Once you feel ready and all the remaining comments are addressed, let's advance this to `activation-in-progress`.  Once the trait definition is on mainnet, can you add its address to the `Activation` section?

pxydn (approved):

friedger:
@muneebm I try to incorporate your comments. However, it removes the addition implementation conditions. Is this on purpose? I think the specification must also provide requirements for the contract implementation. These could be tested by application developers.

friedger:
The trait has been deployed to testnet and mainnet!

jcnelson:
Awesome! This SIP is now in activation-in-progress status.

hstove:
Given our earlier discussion and https://github.com/stacksgov/Stacks-Grants/issues/73, are we on the same page about only a `uint` error code for `transfer`?

friedger:
@hstove Yes, sip has been updated, new version deployed to mainnet, new beeple nft published.

pxydn (approved):
all looks good. ready for launch! 🚀

friedger:
@314159265359879 The name is not part of the sip, IMHO. You can deploy them using any name and as often as you want. I think in the future, we will have some better referencing to sip trait contracts, e.g. via BNS names.

friedger:
Should this be merged as the sip is activated?

jcnelson:
And it's official!  Congratulations @friedger!  `SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait.nft-trait` is now officially a standard :tada:

jcnelson:
Thanks to everyone who helped review and get this ready!

PR #10, opened 2021-02-11
SIP 011

jcnelson:
This SIP attempts to codify a set of procedures for recovering from _catastrophic errors_ in the blockchain that either cause it to crash, or cause severe safety problems for other peoples' digital assets and code.  I was inspired to write this in light of the recent network outage on 7 February.  The procedures outlined in this SIP are meant to be "game plans" for dealing with future such events, as well as drawing a few lines in the sand as to what's a legitimate reason for following some of the more severe recovery procedures (e.g. forks).

### DIFF
--- a/sips/sip-011/sip-011-catastrophic-failure-recovery.md
+++ b/sips/sip-011/sip-011-catastrophic-failure-recovery.md
@@ -1,0 +1,459 @@
+# Preamble
+
+SIP Number: 011
+
+Title: Catastrophic Blockchain Failures and Recovery
+
+Author: Jude Nelson <jude@stacks.org>
+
+Consideration: Governance
+
+Type: Meta
+
+Status: Draft
+
+Created: 8 February 2021
+
+License: BSD 2-Clause
+
+Sign-off:
+
+Discussions-To: https://github.com/stacksgov/sips
+
+# Abstract
+
+This SIP describes the sanctioned procedures for recovering the state of the
+Stacks blockchain in the event of a _catastrophic_ failure.  A failure qualifies
+as a _catastrophic_ failure if and only if there is no conceivable way for the
+correct nodes in the network to make progress and preserve safety without human
+intervention.
+
+This document describes how to identify a catastrophic failure, and how the
+various parties in the Stacks ecosystem should work together to resuscitate the
+blockchain.  Importantly, these procedures seek to minimize the degree of human
+intervention and code change whenever possible.  Blockchains represent a social
+contract between the people who run the nodes (see SIP-000), so care must be
+taken to preserve the blockchain's expected SIP-described rules and behaviors
+above all else.
+
+# License and Copyright
+
+This SIP is made available under the terms of the BSD-2-Clause license,
+available at https://opensource.org/licenses/BSD-2-Clause.  This SIP’s copyright
+is held by the Stacks Open Internet Foundation.
+
+# Introduction
+
+What does it mean for a blockchain to fail?  They are designed to be resilient
+to node churn, individual node crashes, and (short-lived) network partitions,
+but two broad categories of bugs can cause the whole blockchain network to enter
+an irrecoverable state in which human intervention becomes necessary to recover.
+This document qualifies these failure modes, and describes various procedures
+for recovering from them.
+
+The first class of bugs are _liveness failures_.  These are bugs that prevent
+the blockchain from making progress.  For example, a bug that causes all nodes
+in the network to simultaneously crash is a liveness failure.  As another
+example, a bug that causes all nodes in the network to become unable to respond
+to network requests is a liveness failure.
+
+The second class of bugs are _safety failures_.  These are bugs in which the
+blockchain itself permits state-transitions that are undesirable and
+inconsistent with the design of the system.  For example, a bug that would let
+anyone move anyone else's STX tokens would be a safety failure.  As another
+example, a bug in which some Clarity code's execution led to the wrong result
+(but consistent across the network) would be a safety failure.
+
+This document describes the ways in which Stacks ecosystem participants can
+coordinate to recover from various instances of these failure modes, and when it
+is appropriate to use them.  These procedures include:
+
+* Announcing a new point-release of the Foundation's reference implementation of
+  the Stacks node software
+* Covertly fixing systemically-important nodes to fix security-sensitive bugs
+* Altering or extending the p2p network protocol
+* Soft forks, which constrain what kinds of transactions may be mined on the
+  canonical fork
+* Hard forks, which are backwards-incompatible upgrades
+
+## A Note on Scope
+
+The absence of a feature is not a catastrophic failure.  The blockchain behaving
+consistently with the SIPs that describe its design is not a catastrophic
+failure, even if the behavior is not desired.  This SIP does not describe how to
+remedy these situations; it is only concerned with manual recovery procedures
+and guidelines for restoring the network after a network-wide safety or liveness
+failure is discovered.
+
+# Specification
+
+The following catastrophic failure recovery procedures are presented in order
+from least-disruptive to most-disruptive.  When choosing the recovery procedure
+to execute, ecosystem participants should err on choosing the least disruptive
+procedure that will fix the problem at hand.  A recovery procedure should only
+be taken only once all less-disruptive recovery procedures have been considered
+or tried.
+
+## Procedure for Pushing out a Critical Bugfix Point-Release
+
+In the event that a safety or liveness failure can be readily addressed in
+public _without_ altering the blockchain's social contract (i.e. no changes to
+the networking protocols, mining protocols, or chainstate validation protocols),
+the following procedure will be carried out:
+
+1. The code for the fix will be written and pushed to a Github branch prefixed
+with `fix/` to the Stacks Blockchain reference implementation, hosted at
+https://github.com/blockstack/stacks-blockchain.
+2. The branch will be submitted as a pull-request to the `master` branch, and
+will be reviewed and approved by at least two blockchain engineers, representing
+both the Stacks Foundation and at least one other major Stacks ecosystem entity.
+3. If warranted, an unofficial announcement will be made to various public
+avenues where Stacks miners are known to gather, such as the Stacks Discord
+server or the Stacks forum (https://forum.stacks.org), in order to inform them
+that the source code to the fix is available to be built.
+4. A build process will be triggered to produce pre-built binaries for users to
+download.
+5. Once the binaries are available, a cryptographically-signed email will be
+sent from announce@stacks.org with the cryptographic digests of the binaries and
+download links to them.  The email will detail any special considerations and
+instructions for recovering from the bug.
+
+Anyone can subscribe to receive critical bugfix emails by sending an email to
+announce+subscribe@stacks.org.  Critical bugfix announcements will also be
+posted to the Stacks forum (https://forum.stacks.org).
+
+This procedure is further incorporated into procedures described below, and is
+referenced as the "critical bugfix point-release procedure."
+
+## Procedure for Pushing out Sensitive Bugfixes
+
+Depending on the severity of the bug, it may be necessary for ecosystem
+participants to quietly fix and/or reboot their nodes with a bugfix before
+making any announcement.  For example, if the bug is an as-of-yet-unexploited
+safety failure that could lead to loss or theft of digital assets, it may be
+more prudent to silently upgrade vulnerable nodes before announcing the
+vulnerability.
+
+If it is necessary to do this, then the Foundation or an agent of its choosing
+will reach out to vulnerable ecosystem participants with instructions on how to
+proceed.  In order to minimize the chance that the details of an unexploited but
+critical bug will be leaked, the Foundation (or its agent) may take any means
+necessary to ensure its secrecy in order to protect unknowingly-vulnerable
+users.  This may include, but is not limited to, requiring ecosystem
+participants so contacted to sign a legally-binding non-disclosure agreement.
+
+Once the vulnerability no longer poses a systemic threat to the Stacks ecosystem
+(or once the bug gets exposed, whichever happens first), the procedure for
+pushing out a critical bugfix point-release will be followed.
+
+To encourage users who discover such sensitive blockchain bugs to report them
+while keeping them secret, the Stacks Foundation will a bug-bounty program that
+will be set up once this SIP activates.
+
+## Procedure for Altering Network Protocols
+
+If a catastrophic failure can be traced to a bug in the Stacks blockchain's
+network protocols (see SIP-003), and if it can be fixed without changing the
+protocol(s) in an incompatible way, then it will be fixed through the critical
+bugfix point-release procedure above.
+
+The Stacks network protocols offer nodes a lot of leeway in how they are allowed
+to behave, so it is unlikely that the protocol itself will need to be altered in
+an incompatible way in order to recover from a catastrophic failure.  But that
+does not by itself preclude the possibility of a catastrophic failure
+originating from the network's design and/or implementation.  In all likelihood,
+failures in the network protocol will be liveness failures.
+
+Broadly speaking, a critical bugfix to the network protocols can take one or
+both of the following two forms:
+
+* Adding _new_ network messages whose handling does not interfere or interact
+  with code for handling existing messages.
+* Making an incompatible change to how the system handles _existing_ messages.
+
+### Adding New Network Messages
+
+If the catastrophic failure can be remedied by adding a new type of network
+message which old nodes will ignore, then the procedure for rolling out the
+bugfix is as follows:
+
+1. If the message types are part of the p2p network protocol, a new bit pattern
+must be defined the `services` bitfield in the p2p message preamble that
+distinguishes nodes with the fix applied from nodes that do not have the fix.
+This, in turn, will allow the fixed nodes to know when it is appropriate to use
+the new message types.
+2. The Foundation and other ecosystem entities will deploy new versions of their
+nodes, which use the new `services` bitfield.
+3. Once the Foundation believes that there are enough publicly-routable nodes
+with the fix applied, it will begin executing the critical bugfix rollout
+procedure per above.
+
+In this procedure, the network versioning information is preserved.  Therefore,
+this procedure is only appropriate if backwards-compatibility with other nodes
+and downstream clients can be preserved.
+
+This procedure should be executed instead of changing the networking protocol if
+at all possible.
+
+### Changing Existing Network Messages
+
+If the catastrophic failure can only be remedied by changing the way nodes
+handle an existing message, such that the change renders the node incompatible
+with the network (i.e. unable to effectively participate in block
+synchronization, relaying, and/or mining), then the bugfix will require rolling
+out two concurrent versions of the network that share the same chainstate for a
+time in order to allow people running old nodes a chance to upgrade.  The
+deadline to upgrade (and the deadline at which the old, buggy code-paths will
+cease to be supported) will be enforced programmatically by new nodes, and will
+be measured in burnchain block heights.
+
+This bugfix procedure consists of the following steps.
+
+1. If at all possible, a version of the node software with the buggy code paths
+disabled will be released in order to provide partially-degraded service while
+booting up the new, incompatible version of the network with the fix applied.
+This release may have certain features intentionally disabled, or may introduce
+performance regressions, as long as doing so does not introduce any other
+liveness or safety failures.
+2. The Foundation and other ecosystem entities will run versions of the degraded
+but functional node software from step 1, if applicable.  If a pre-built binary
+needs to be generated to encourage its mass deployment, then it will be
+announced via a signed email from announce@stacks.org as a critical bug-fix
+point release, with instructions on how to proceed in the coming network
+upgrade.
+3. Once the immediate liveness or safety bug no longer manifests (even if it
+means operating the blockchain in a degraded state), the Foundation will procure
+a _testnet_ release of the node software with the incompatible fix applied.
+Power users and ecosystem entities are encouraged to run this testnet release in
+order to verify that it solves the problem and works correctly.
+4. Once the Foundation and at least one other significant ecosystem entity are
+satisfied that the new network version solves the problem, they will produce a
+second critical bugfix point-release for mainnet.  To the greatest extent
+possible, the new networking code will remain compatible with the old but broken
+network's message formats and interfaces.  The Foundation will provide a
+detailed explanation of how old nodes will interact with new nodes as part of
+this release's email from announce@stacks.org.
+5. Once the new version of the network is running, each ecosystem entity,
+including the Foundation, may independently decide when to shut down their
+degraded nodes in favor of their new incompatible nodes.  It is highly
+recommended that both versions are kept alive for as long as possible, so that
+the blockchain can be resuscitated with the new network version while providing
+service to downstream dependencies that are not yet compatible with the new
+version.
+6. The new software release in step 5 will set a reasonable deadline, measured
+in burnchain block heights, at which point any legacy support for the old,
+replaced networking code paths will be dropped.  This deadline will be no longer
+than six months into the future.  The exact block height is chosen at the
+discression of the Foundation, with advice from affected ecosystem entities.
+
+The Foundation will keep its old nodes online as long as it believes that there
+is a systemic risk to shutting them down, and will send an email to
+announce@stacks.org to indicate when support for the old network version will be
+terminated.  This deadline may exceed the built-in deadline, at the Foundation's
+discression.
+
+Note that if this procedure is used, ecosystem participants should expect no
+fewer than three separate emails from announce@stacks.org -- one in step 2, one
+in step 4, and one in step 5.  The email in step 5 will list the burnchain block
+height at which support for any legacy, broken networking code will be dropped
+(if applicable).
+
+The engineers who procure the new incompatible network version should take steps
+to retain backwards-compatibility with old clients and old nodes to the greatest
+extent possible until the upgrade deadline passes.  This is because at a
+minimum, the new incompatible node code will need to boot off of an old node's
+chainstate.
+
+The incompatible node implementation must change any network version signaling
+information (e.g. HTTP prefixes, p2p preamble version bytes, etc.) so that old
+versions of the network do not attempt to contact it.
+
+## Procedure for Soft Forks
+
+A _soft fork_ is a backwards-compatible change to the blockchain's consensus
+rules.  Soft forks work by constraining the set of state-transitions that may be
+applied to the canonical chain tip.  Old nodes will remain capable of validating
+the Stacks blockchain data without the new rules applied, and will still be able
+to calculate the same canonical fork.
+
+Soft forks are a potentially-contentious form of upgrade, and should be avoided
+in favor of any of the above procedures if at all possible.  Therefore,
+executing a soft fork necessarily requires a high degree of cooperation between
+ecosystem participants in order to make sure a significant majority are onboard
+with the changes.
+
+A soft fork may be an appropriate course of action if healing a catastrophic
+failure requires the following kinds of changes, among others (this list is
+exemplary, and not exhaustive):
+
+* Adding supplementary block data (and validation rules for it) that gets
+  considered and stored "outside of" blocks, but which correct miners must
+nevertheless produce in order to mine a valid block on the canonical fork.  For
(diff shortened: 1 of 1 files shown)


sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,459 @@
...
+in burnchain block heights, at which point any legacy support for the old,
+replaced networking code paths will be dropped.  This deadline will be no longer
+than six months into the future.  The exact block height is chosen at the
+discression of the Foundation, with advice from affected ecosystem entities.
+
+The Foundation will keep its old nodes online as long as it believes that there
+is a systemic risk to shutting them down, and will send an email to
+announce@stacks.org to indicate when support for the old network version will be
+terminated.  This deadline may exceed the built-in deadline, at the Foundation's
+discression.
```
whoabuddy, line 253: typo: discretion

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,459 @@
...
+liveness or safety failure.  A subsequent SIP may codify other types of soft
+forks and their applications in different contexts.  Due to the comparatively
+low barrier-to-entry for mining on the Stacks chain, miner-activated soft forks
+are expected to offer an acceptable measurement of the ecosystem's sentiments on
+a pending soft fork.
+
+A miner-activated soft fork procedure for healing a catastrophic failure works
+as follows:
+
+1. The Foundation and one other significant ecosystem entity will approve and
```
whoabuddy, line 322: entity --> entities

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,459 @@
...
+permanently at the end of the voting period.  If mining cannot be resumed, then
+this step is skipped.
+4. A new version of the node software will be released via the critical bugfix
+point-release procedure.  The node software will resume from the old chainstate
+at the last block mined before the catastrophic failure was manifested (as
+determined by the burnchain block height).
+
+Step 3 should be executed if possible, because it gives the network a chance to
+vote on the new rules.
+
```
whoabuddy, line 423: Add to match the other sections:

Note that in this procedure, at least two emails from announce@stacks.org will
be sent -- one in step 1 and one in step 4.

whoabuddy:
Thank you for this! Without knowing much about soft and hard forks on the technical side, this SIP gave me a great understanding of what capabilities exist to mitigate a catastrophic failure. Each possible action contained a clear description and action plan, with emphasis on using least-disruptive method possible.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,459 @@
...
+the networking protocols, mining protocols, or chainstate validation protocols),
+the following procedure will be carried out:
+
+1. The code for the fix will be written and pushed to a Github branch prefixed
+with `fix/` to the Stacks Blockchain reference implementation, hosted at
+https://github.com/blockstack/stacks-blockchain.
+2. The branch will be submitted as a pull-request to the `master` branch, and
+will be reviewed and approved by at least two blockchain engineers, representing
+both the Stacks Foundation and at least one other major Stacks ecosystem entity.
+3. If warranted, an unofficial announcement will be made to various public
```
whoabuddy, line 110: When would this not be warranted? Given:

> A failure qualifies as a _catastrophic_ failure if and only if there is no conceivable way for the correct nodes in the network to make progress and preserve safety without human intervention.

Two options that could happen automatically:

- setup GitHub to announce any releases for stacks-blockchain automatically
- setup Discord to accept and relay the message from announce@stacks.org

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,459 @@
...
+participants so contacted to sign a legally-binding non-disclosure agreement.
+
+Once the vulnerability no longer poses a systemic threat to the Stacks ecosystem
+(or once the bug gets exposed, whichever happens first), the procedure for
+pushing out a critical bugfix point-release will be followed.
+
+To encourage users who discover such sensitive blockchain bugs to report them
+while keeping them secret, the Stacks Foundation will a bug-bounty program that
+will be set up once this SIP activates.
+
```
HaroldDavis3, line 152: **typo:** ..Foundation will start*? bug-bounty program. 

Would love to explore this more, may be relevant to COC?
The incentive weights of malicious actors is also interesting to play with in bug-bounties mentioned presumably?
 
Ty kindly for all this, this kind of SIP goes a long way to opening up maintenance know how in the space 💯  agree with @whoabuddy's sentiments there.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: All the more-disruptive procedures use this procedure as a step.  Some of them, like the one below it, require a public embargo on talking about the bug because the act of making the bug public knowledge would make the problem worse.  For example, if someone discovered a bug that would let an attacker steal anyone's STX, we would _not_ announce the bug until the fix was already deployed.

I like the idea of relaying all messages from announce@stacks.org to Discord.  Is there a bot that can do this?  Not too familiar with Discord (I'm old-school), but I'd love to have a dedicated #announce channel which just relayed these messages.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: I really appreciate your feedback @HaroldDavis3.  Ty kindly for taking the time to read through this SIP :)

> The incentive weights of malicious actors is also interesting to play with in bug-bounties mentioned presumably?
If at all possible, I think we'll want to make it more profitable for malicious actors who find bugs that trigger catastrophic failures to just tell us about the bug privately instead of exploit it.  We'd want to do some research and figure out what the ROI would be for exploiting various kinds of serious bugs, and see if we can offer some kind of ROI-matching bug bounty (it's a tricky calculation -- just because the attacker can steal funds on-chain, for example, doesn't necessarily mean that they can cash them out without consequences).

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: Fixed in 3a9ec98

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: Fixed in 3a9ec98

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: Fixed in 3a9ec98

jcnelson:
Would y'all like to talk about this some more at the governance call this week?

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,464 @@
...
+inconsistent with the design of the system.  For example, a bug that would let
+anyone move anyone else's STX tokens would be a safety failure.  As another
+example, a bug in which some Clarity code's execution led to the wrong result
+(but consistent across the network) would be a safety failure.
+
+This document describes the ways in which Stacks ecosystem participants can
+coordinate to recover from various instances of these failure modes, and when it
+is appropriate to use them.  These procedures include:
+
+* Announcing a new point-release of the Foundation's reference implementation of
```
diwakergupta, line 71: [nit] Should we note somewhere at the top that "Foundation" references in this doc specifically mean the "Stacks Foundation"?

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,464 @@
...
+In the event that a safety or liveness failure can be readily addressed in
+public _without_ altering the blockchain's social contract (i.e. no changes to
+the networking protocols, mining protocols, or chainstate validation protocols),
+the following procedure will be carried out:
+
+1. The code for the fix will be written and pushed to a Github branch prefixed
+with `fix/` to the Stacks Blockchain reference implementation, hosted at
+https://github.com/blockstack/stacks-blockchain.
+2. The branch will be submitted as a pull-request to the `master` branch, and
+will be reviewed and approved by at least two blockchain engineers, representing
```
diwakergupta, line 108: [nit] s/"blockchain engineers"/"developers with write access to the stacks-blockchain repo"

Alternatively: "two members of the Stacks Core Developers group"?

sips/sip-011/sip-011-catastrophic-failure-recovery.md
diwakergupta: Why even have a separate (3) step? IMO step (5) could very well be re-worded as "Availability of new binaries will be communicated via standard channels per established norms of the project -- currently via the announce@ email list and Discord"

In particular, is there some specific advantage an "unofficial" announcement has? If Github releases are a source of truth, availability of a new release is automatically a "formal" announcement IMO.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,464 @@
...
+If a catastrophic failure can be traced to a bug in the Stacks blockchain's
+network protocols (see SIP-003), and if it can be fixed without changing the
+protocol(s) in an incompatible way, then it will be fixed through the critical
+bugfix point-release procedure above.
+
+The Stacks network protocols offer nodes a lot of leeway in how they are allowed
+to behave, so it is unlikely that the protocol itself will need to be altered in
+an incompatible way in order to recover from a catastrophic failure.  But that
+does not by itself preclude the possibility of a catastrophic failure
+originating from the network's design and/or implementation.  In all likelihood,
```
diwakergupta, line 164: We could point to the recent bug with block limits, or the Atlas issues as examples?

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,464 @@
...
+needs to be generated to encourage its mass deployment, then it will be
+announced via a signed email from announce@stacks.org as a critical bug-fix
+point release, with instructions on how to proceed in the coming network
+upgrade.
+3. Once the immediate liveness or safety bug no longer manifests (even if it
+means operating the blockchain in a degraded state), the Foundation will procure
+a _testnet_ release of the node software with the incompatible fix applied.
+Power users and ecosystem entities are encouraged to run this testnet release in
+order to verify that it solves the problem and works correctly.
+4. Once the Foundation and at least one other significant ecosystem entity are
```
diwakergupta, line 229: Is it worth qualifying "significant ecosystem entity" for this document? Perhaps "entities operating one or more publicly routable nodes" or "entities operating default bootstrap nodes"?

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,464 @@
...
+2. Once the catastrophic bug has been addressed, and a reliable majority of
+blocks on the canonical fork are produced by miners running the new software, a
+SIP describing the soft fork will be written and published describing the new
+rules and why they are necessary.  This SIP must then be advanced in the SIP
+process to "activation-in-progress" status by the relevant advisory boards and
+technical steering committee.  Among other things, this SIP's activation
+criteria must require the following:
+
+   * At least 80% of all miners must _explicitly and continuously vote to
+     accept_ the soft fork rules for at least _two consecutive_ whole PoX reawrd
```
diwakergupta, line 340: [typo] "reawrd"

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,464 @@
...
+new release is the beginning of a soft-fork roll-out.
+2. Once the catastrophic bug has been addressed, and a reliable majority of
+blocks on the canonical fork are produced by miners running the new software, a
+SIP describing the soft fork will be written and published describing the new
+rules and why they are necessary.  This SIP must then be advanced in the SIP
+process to "activation-in-progress" status by the relevant advisory boards and
+technical steering committee.  Among other things, this SIP's activation
+criteria must require the following:
+
+   * At least 80% of all miners must _explicitly and continuously vote to
```
diwakergupta, line 339: How do we measure this? Is this 80% of nodes on the network that have "miner=true"?

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,464 @@
...
+     accept_ the soft fork rules for at least _two consecutive_ whole PoX reawrd
+cycles. Additional, stricter criteria are permitted.
+   * An _activation window_, defined by a _start activation height_ and an _end
+     activation height_, represented as burnchain block heights, must be
+stipulated in the activation criteria.  The aforementioned miner vote must take
+place and run to completion in-between these two block heights.  The block
+heights _must_ correspond to the beginnings of PoX reward cycles, and must be
+between 14 reward cycles (1 maximum Stacking period plus two reward cycles) and
+25 reward cycles (about 1 year) apart.
```
diwakergupta, line 348: Is this too long? I'm imagining a situation where say two soft-forks are needed, 6-months apart. How do the start/end activation heights work in this case (they would overlap)?

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,464 @@
...
+   * A soft-fork-specific bit pattern must be specificied, which miners _must_
+     include in their coinbase transactions in order to indicate unambiguous
+support for activating the new rules.
+
+3. Once a release with the soft fork voting rules and activation window is
+written and adequately tested (possibly including rolling it out to one or more
+public testnets first), it will be announced via a signed email from
+announce@stacks.org.  This release announcement must occur before the activation
+window begins.
+4. If at least 80% of all miners vote to accept the soft fork rules for two
```
diwakergupta, line 358: How do miners vote? Is it by simply continuing to mine with the version that introduces the soft fork?

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,464 @@
...
+participants.
+
+A hard fork is only necessary for repairing catastrophic failures if all other
+procedures prove inadequate.  It is not to be used in any other circumstance.
+Therefore, the kinds of catastrophic failures that necessitate hard forks are
+artifacts of a design bug in the Stacks blockchain that renders it unsustainable
+or unable to operate as intended.
+
+Some examples of when a hard fork is the appropriate course of action include,
+but are not limited to:
```
diwakergupta, line 387: I would also add something like "Introducing new features or capabilities that the ecosystem is overwhelming in favor of". For instance, improvements to Stacking would fall into this category.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,464 @@
...
+vote on the new rules.
+
+Note that in this procedure, at least two emails from announce@stacks.org will
+be sent -- one in step 1 and one in step 4.  However, given the severity of any
+bug that can only be resolved with a hard fork, there may be many more
+communications sent.
+
+# Related Work
+
+This section is a work in process.
```
diwakergupta, line 431: Which of the WIP sections are blockers for this SIP? Do you need help fleshing those out?

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,464 @@
...
+
+Bitcoin has a codified process for soft fork votes.  It should be cited here.
+
+Tezos has an even more extreme version of soft-forking than we support, but it's
+still a related concept to be cited.
+
+Ethereum is a good example of how _not_ to do hard forks -- they just happen
```
diwakergupta, line 438: I think it's fair to provide an objective commentary of the landscape. Maybe reframe as "The author(s) of this SIP find hard-forks in Ethereum to be not as rigorous as they would like"?

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,464 @@
...
+willy-nilly once enough "core developers" are on board.
+
+# Activation
+
+This section is a work in progress.  Feedback is welcome.
```
diwakergupta, line 443: Neither options are binding nor enforceable. 

Option 1 is no different than the entities (or their verified representatives) publicly expressing support as a comment on this PR?

I like the idea of capturing the support on-chain. Perhaps we can combine the two options by having a smart-contract that determines activation based on contract-calls from a set of well-known, whitelisted addresses representing the entities here?

diwakergupta (changes requested):
Thanks @jcnelson -- this is a great starting point, left bunch of minor comments!

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: > In particular, is there some specific advantage an "unofficial" announcement has? If Github releases are a source of truth, availability of a new release is automatically a "formal" announcement IMO.

The point is to let node operators know when the source code for the fix is available for scrutiny and testing (which is mentioned at the end of (3)).

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: 80% of all blocks produced within the time interval.  The set of miners (really, their mining power) is fungible.  Will clarify.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: Multiple soft forks can happen concurrently if they don't conflict or aren't dependent on one another.  This sentence is only a statement about the possible values the activation window block heights can take.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: It's mentioned just above this line:

```
   * A soft-fork-specific bit pattern must be specificied, which miners _must_
     include in their coinbase transactions in order to indicate unambiguous
     support for activating the new rules.
```

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: If the ecosystem is overwhelmingly in favor of doing something with the urgency you suggest, then a SIP is neither going to stop them nor give them permission.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: This section is basically comprised of notes jotted down.  They're not even close to being up to snuff.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: I'm going to remove that term, since it's so hard to qualify.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: This SIP mainly affects the behaviors of Stacks Core Developers and the Stacks Foundation, since it codifies their response to catastrophic failures.  It's meant to be informational to all other ecosystem participants -- i.e. to make sure they know what to expect in the event of a catastrophic failure.  So, that makes me think that neither of these are good options.  I think it might be sufficient to just have approval from the Stacks Core Developers who will be responsible for fixing these problems if they come up.  Of course, other ecosystem participants are free to do what they like -- nothing about this SIP can be legally or programmatically enforced.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: Addressed in 3832ae2

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: Ah, I see Github rendered the quoted text in the above comment.

I don't think the set of people who have write access to the stacks blockchain repo accurately reflects the set of people will be called up on to fix catastrophic bugs.  I addressed this in 3832ae2 by defining the Stacks Core Developers as a list of developer names and contacts in a supplementary file, which can be added to or removed from by the Steering Committee.  To be clear, this is just the list of folks who will be called upon to execute these procedures; this isn't meant to be an exclusionary list of who's-who in the blockchain or some other social club or clique.  It's more like a "who is it okay to email at 3am on a Saturday if the blockchain halts and catches fire" list.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: (Github didn't render your comment in the appropriate place for some reason.  Meant to post this reply here).

The point is to let node operators know when the source code for the fix is available for scrutiny and testing (which is mentioned at the end of (3)).

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: Addressed in 3832ae2.  Since this section pertains to how we're going to repair a broken network, I think it's safe to say that the list of folks who need to sign off on (or at least not actively oppose) a solution to the breakage are simply the folks who showed up in step 2 to help tide the system over from the broken and degraded network to the new network.  These are exactly the folks who will be well-positioned to make the call as to whether or not the new incompatible network has fixed the problem.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: Addressed in 3832ae2.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: I clarified this a bit more in 3832ae2.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: I just need to sit down and finish them.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: I went ahead and dropped options 1 and 2 in 3832ae2 in favor of just ratifying this SIP by having the Stacks Core Developers who will be responsible for running these procedures sign off on it.  One way or another, other ecosystem participants are going to be looking to these folks (and their employers) for guidance if the blockchain encounters a catastrophic error, so I don't think it's necessary to get their sign-off (they already expect us to help them).

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: Addressed in 3832ae2

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: Addressed in 3832ae2

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,473 @@
...
+     accept_ the soft fork rules for at least _two consecutive_ whole PoX reward
+cycles. Additional, stricter criteria are permitted.
+   * An _activation window_, defined by a _start activation height_ and an _end
+     activation height_, represented as burnchain block heights, must be
+stipulated in the activation criteria.  The aforementioned miner vote must take
+place and run to completion in-between these two block heights.  The block
+heights _must_ correspond to the beginnings of PoX reward cycles, and must be
+between 14 reward cycles (1 maximum Stacking period plus two reward cycles) and
+25 reward cycles (about 1 year) apart.
+   * A soft-fork-specific bit pattern must be specificied, which miners _must_
```
jcnelson, line 361: I'm actually not sure this is the best we can do.  It might be a lot safer and more reliable to use a particular bit-pattern in the remaining 5 bits of a block-commit's `memo` field to indicate which soft-forks are under consideration.  There could be at most 5 in parallel; the node software would need to know which bit in which reward cycle corresponded to which soft-fork.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: added

sips/sip-011/sip-011-catastrophic-failure-recovery.md
jcnelson: added

jcnelson:
@kantai I've added a set of case studies to demonstrate how each of these catastrophic error recover procedures can be used.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,763 @@
...
+3. Once a release with the soft fork voting rules and activation window is
+written and adequately tested (possibly including rolling it out to one or more
+public testnets first), it will be announced via a signed email from
+announce@stacks.org.  This release announcement must occur before the activation
+window begins.
+4. If at least 80% of all sortitions in two consecutive, whole PoX prepare
+   phases vote to accept the soft fork rules within the activation window, then the new
+rules will take effect starting in the next whole reward cycle (i.e. right after
+the second prepare phase finishes).  All new releases of the Stacks node will
+adhere to the soft fork rules, since they are now part of the block validation rules.
```
kantai, line 418: In this soft-fork process, would the node released in step 1 _immediately_ begin enforcing the soft-fork rules (i.e., ignoring blocks that do not follow the soft fork)? Then is step 4 just a consolidation of this rule (i.e., the declaration that all future releases will also follow the soft fork rule)?

jcnelson: Not necessarily.  This SIP only requires that miners who vote for the soft-fork do so through this signaling procedure, and that the new rules must take effect _no later than_ when this activation threshold is met.  The SIP intentionally does not specify anything about how early miners can start deciding when to ignore blocks they would no longer consider valid.  In fact, per point 2, the provision that "stricter criteria are permitted" is meant to allow a particular soft-fork upgrade to impose additional requirements for miners to begin activating the new rules.  For example, in the third case study below, a particular soft fork to repair a bug in smart contract processing could require miners to begin orphaning blocks in which the bug manifests ASAP.

sips/sip-011/sip-011-catastrophic-failure-recovery.md
```diff
@@ -0,0 +1,763 @@
...
+software would work on a set of forks in which no subsequent smart contracts and
+contract-calls execute.
+
+In addition, the Stacks Core Developers would coordinate with the Foundation to
+release a version of the node software that included a fix for the bug, as well
+as code to re-activate smart contracts under the condition that
+the vast majority of miners have _rejected_ all the forks in
+which this bug's exploits have occurred.  In other words, smart contracts would
+only re-activate if a fork that does _not_ descend from any block in which an
+exploit occurs becomes the dominant fork.
```
kantai, line 530: Could this itself be a hard-fork? Repairing these kinds of bugs could easily create a hard-fork, but is this class of solution only considering soft-fork-able fixes?

jcnelson: So, admittedly this example is a bit contrived because it assumes that there's a bug severe enough that anyone can take anyone else's STX, but at the same time, we can somehow identify each occurrence of the exploit each STX's true owner.  This maybe isn't the best example.

The point of this section is to explain how to use a chain fork to "undo" any active exploits, assuming they could be identified as such.  The real-world example I had in mind of this was Bitcoin's integer overflow bug whereby 2**64 - 1 BTC got minted.  This was handled by a soft fork then, and we could do something similar if there was a STX minting bug or some other kind of bug that resulted in the number of liquid STX increasing unexpectedly.

whoabuddy:
Note: the first goal for the Governance CAB will be to review and comment on SIP-011 by the next governance meeting on 2021/09/16, after which we can discuss any comments and help move this toward being ratified!

AcrossfireX:
Further advancement of this particular SIP is pending more real world stress testing of this SIP being used in DR scenarios. As this has been successfully utilized in recent Stacks updates that should be noted and consideration for moving to Accepted should be considered

PR #14, opened 2021-02-24
Fix misspelled word

larrysalibra:
Originally from: https://github.com/blockstack/stacks-blockchain/pull/2445


jcnelson:
Thanks @larrysalibra!

PR #15, opened 2021-03-04
Fix field ID values in SIP-005

jecassis:
- Fix field ID values in spending field body for signatures in SIP-005
 - Correct typos and grammar in SIP-005


jcnelson (approved):
lol wee kant splel ;)

Thanks for fixing all the typos!

PR #18, opened 2021-03-17
Metadata for NFTs

friedger:
This PR
* adds a SIP about how metadata for NFTs are defined


hstove:
For this SIP, I would ask "why"? I definitely agree that a standard for metadata is necessary. However, why should we keep it on-chain?

One reason to keep it on-chain is so that other contracts can consume this information. But, for metadata, why do they need to? Would an external contract ever need to do some logic based on this contract's `mime-type`, or `image-url`? Perhaps there is a registry for only JPEG NFTs. But, in that case, you cannot guarantee that the actual type of the image is JPEG - that's just what the contract says.

Another reason might be to provide some guarantee as a client that the information returned from an HTTP request to `get-token-uri` is not changing, or is not returning different data to different clients. In that case, storing the hash of the result of that HTTP request would be sufficient.

friedger:
superseded by #44

PR #19, opened 2021-03-18
Update activation rule to make sip 9 and sip 10 compatible

friedger:
This PR
* updates the activation rule to make the error type of the transfer function in sip 9 and sip 10 compatible.

Discussions and recommendations should happen on https://github.com/clarity-lang/reference/issues/33


jcnelson:
@hstove Does this change work for you?

hstove:
There is some ongoing discussion in the #clarity channel on Discord, and it seems that we have three devs in favor of `uint` error codes, vs your preference of a tuple with a string. I'm not saying we have to definitely do the error code, but I don't think it's right to simply force this change through

friedger:
@hstove This was only about adding an activation rule that both trait use the same error type whatever it is.

friedger:
This PR is no longer relevant.

PR #5, opened 2021-01-26
feat: proposed fungible token trait

hstove:
Included is a proposed standard and trait for fungible tokens on the Stacks blockchain.

This addresses #1 and https://github.com/stacksgov/Stacks-Grants/issues/44

cc @psq - almost all of this code comes from your work, and you deserve all the credit for this. Please request any changes you'd like to see, and I will include them.

PR #25, opened 2021-05-12
SIP10 addition: add spec for memo

lgalabru:
This PR is a proposal for augmenting the transfer function with an optional memo field.
I'd also recommend going with a naming more explicit for this trait - something like `sip-010-trait` would be more descriptive: this trait is the first, but we will probably see other kind of fungible token in the future.

### DIFF
--- a/sips/sip-010/sip-010-fungible-token-standard.md
+++ b/sips/sip-010/sip-010-fungible-token-standard.md
@@ -43,15 +43,15 @@ The Stacks blockchain has a native fungible token: the Stacks token (STX). In ad
 
 # Specification
 
-The fungible token trait, `ft-trait`, has a 6 functions:
+The fungible token trait, `sip10-trait`, has 7 functions:
 
 ## Trait functions
 
 ### Transfer
 
-`(transfer ((amount uint) (sender principal) (recipient principal)) (response bool uint))`
+`(transfer ((amount uint) (sender principal) (recipient principal) (memo (optional (buff 34)))) (response bool uint))`
 
-Transfer the fungible token from the sender of this transaction to the recipient. The `amount` is an unsigned integer. It is recommended that implementing contracts use the built-in `ft-transfer` Clarity method. If the sender does not have enough tokens to complete the transaction, the transaction should abort and return an `(err uint)`.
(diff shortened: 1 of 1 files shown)


jcnelson (approved):
This LGTM! Just need @hstove's sign-off as the author.

sips/sip-010/sip-010-fungible-token-standard.md
```diff
@@ -66,6 +66,18 @@ When returning an error in this function, the error codes should follow the same
 | u3         | `amount` is non-positive                        |
 | u4         | `sender` is not the same as `tx-sender`         |
 
+
+Contract implementers should take note that in Stacks 2.0, the memo field won't be included in the event emitted by successful `ft-transfer?` operations. As a consequence, if compliance with exchanges is a requirement, it is recommended to emit an event including the memo, by adding a `print` statement if the `ft-transfer?` is successful.
+
+Exemple:
```
hstove, line 72: Typo here

hstove, line 72:
```suggestion
Example:
```

hstove (approved):
Thanks Ludo!

hstove:
@jcnelson this PR is a bit off because the SIP repo branch for `SIP-10` does not include the latest commits from my PR, and so this PR does not include those commits. I'm happy with merging this, but we should first sync my branch, then fix the one typo here, and then we can merge

sips/sip-010/sip-010-fungible-token-standard.md
lgalabru: Updated, thanks!

lgalabru:
Is this ball in my court, @hstove's, or @jcnelson's?

jcnelson:
Either @hstove or @lgalabru -- it's the SIP authors' job to get the SIP into the right format.

hstove:
Maybe a miscommunication - what I meant is that this PR is trying to merge into the `hstove-feat/sip-10-ft` branch, but the open PR (#5 ) has more recent commits than that branch. So this PR is not quite in sync with the latest of my PR.

Either way, it's all good - I've cherry-picked and pushed @lgalabru's commits to #5. So, I think we can close this PR!

PR #34, opened 2021-09-07
chore: fix sip10 readme url

aulneau:
This updates the URL from the README.md to point to the correct markdown document. previously it was pointing to an outdated version.


jcnelson:
Thanks @aulneau!

Issue #17, opened 2021-03-17
Metadata for NFTs


friedger:
There should a standard that defines a schema for non-fungible tokens using SIP-009

friedger:
I recommend to follow https://docs.opensea.io/docs/metadata-standards

dantrevino:
This is a derivation of opensea's metadata standard and ERC-1155. IMHO it is simpler and easily extendable.
```
{
"name": "test 3",
"description": "asdfasdfasd ",
"properties": {
"series_name": {
"display_type": "",
"trait_type": "string",
"value": "test 3"
},
"series_size": {
"display_type": "number",
"trait_type": "integer",
"value": "100"
},
"date_created": {
"display_type": "date",
"trait_type": "date",
"value": 1629737029184
},
"external_url": {
"display_type": "url",
"trait_type": "string",
"value": ""
},
"animation_url": {
"display_type": "url",
"trait_type": "string",
"value": ""
}
},
"localization": {
"uri": "",
"default": "en",
"locales": [
"en",
"de",
"ar"
]
},
"image": "ipfs://bafybeiej6org5hquxstziqoiqwk65335vf2vqp7v6jfqukowxm7ldubxmy/1024px-Bitcoin.svg.png"
}```

markmhx:
What's the next step here? @dantrevino @friedger might one of you want to open a PR with all necessary elements?

Note Dan's relevant comment [here](https://github.com/blockstack/stacks-blockchain-api/issues/743#issuecomment-916464848) as well:

> The following still needs to be managed:
> 
> 1. metadata protocol. Is this service prepared to handle ipfs (ipfs://)/arweave (ar://) as well as http(s)?
> 2. metadata format. CSV? JSON? Excel? Clarity tuple? No standard is specified by sip009.
> 3. metadata properties. Are images named "img"? "image"? "imageUrl"? "imagen"? Is the name "name" or "title" or > something else?
> 4. what metadata properties must be provided?

friedger:
@radicleart might be a good contributor

radicleart:
I've gone with the open sea standards for props - 'name', 'image', 'attributes' etc - as in https://docs.opensea.io/docs/metadata-standards.

But props like 'image' and 'animation_url' could be better formalised for future proofing - i.e. what type of object is the NFT? Should it support physical as well as digital objects?

[...]

PR #42, opened 2021-10-19
SIP-013 Semi-Fungible Token standard

MarvinJanssen:
This SIP proposes a standard for semi-fungible tokens. Semi-Fungible Tokens, or SFTs, are digital assets that sit between fungible and non-fungible tokens. Fungible tokens are directly interchangeable, can be received, sent, and divided. Non-fungible tokens each have a unique identifier that distinguishes them from each other. Semi-fungible tokens have both an identifier and an amount.

The standard is still in draft stage but a PR is the best way to request comments and feedback. Copying some text from my [reference repo](https://github.com/MarvinJanssen/stx-semi-fungible-token) for brevity.

## Uses

Semi-fungible tokens can be very useful in many different settings. Here are some examples:

**Art**

Art initiatives can use them to group an entire project into a single contract and mint multiple  series or collections in a single go. A single artwork can have multiple editions that can all be expressed by the same identifier. Artists can also use them to easily create a track-record of their work over time. Curation requires tracking a single contract instead of a new one per project.

**Games**

Games that have on-chain economies can leverage their flexibility to express their full in-game inventory in a single contract. For example, they may express their in-game currency with one token ID and a commodity with another. In-game item supplies can be managed in a more straightforward way and the game developers can introduce new item classes in a transparent manner.

## SFTs and post conditions

Post conditions are tricky because it is impossible to make assertions based on custom `print` events as of Stacks 2.0. Still, native events can be utilised to safeguard SFT actions in different ways. The [reference SFT implementation](https://github.com/MarvinJanssen/stx-semi-fungible-token/blob/a2504bb0eb0b4b340e30029e6823bb9e98ac97dc/contracts/semi-fungible-token.clar) defines a fungible token using `define-fungible-token` to allow for post conditions asserting the amount of tokens transferred. It enables the user to state "I will transfer exactly 50 semi-fungible tokens of contract A". I am still exploring options in which the user can also assert the type of token. There are definitely ways in which an NFT defined with `define-non-fungible-token` can be used.

Options I am considering:
- Mint and burn a native NFT with the provided token ID on every SFT action.
- Define a native NFT with a more complex token identifier and mint these to the contract when an event takes place. The challenge is creating something that is unique whilst still making it easy enough to create assertions for. For example:
  ```clarity
  (define-non-fungible-token sft-events {token-id: uint, amount: uint, sender: principal, recipient: principal, nonce: uint})

  (define-private (sft-event (token-id uint) (amount uint) (sender principal) (recipient principal))
  	(nft-mint? sft-events {token-id: token-id, amount: amount, sender: sender, recipient: recipient, nonce: (next-event-nonce)} (as-contract tx-sender))
  )
  ```
The second option makes it so we can also possibly do away with custom `print` events. A downside is that it creates an ever-growing NFT collection on the contract itself.

### DIFF
--- a/sips/sip-013/sip-013-semi-fungible-token-standard.md
+++ b/sips/sip-013/sip-013-semi-fungible-token-standard.md
@@ -10,13 +10,13 @@ Consideration: Technical
 
 Type: Standard
 
-Status: Accepted
+Status: Ratified
 
 Created: 12 September 2021
 
 License: CC0-1.0
 
-Sign-off: 
+Sign-off: Jude Nelson <jude@stacks.org>, Brice Dobry <brice@hiro.so>
 
 Layer: Traits
 


sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,232 @@
...
+```
+## Events
+
+Semi-fungible token contracts should emit custom events in certain situations via `print`. These events should be emitted after any built-in token events (such as those emitted by `ft-transfer?`) and before the memo in the case of `transfer-memo` and `transfer-many-memo`.
+
+| Event name           | Tuple structure                                                                                       | Description                          |
+|----------------------|-------------------------------------------------------------------------------------------------------|--------------------------------------|
+| `sft_transfer_event` | `{type: "sft_transfer_event", token-id: uint, amount: uint, sender: principal, recipient: principal}` | Emitted when tokens are transferred. |
+| `sft_mint_event`     | `{type: "sft_mint_event", token-id: uint, amount: uint, recipient: principal}`                        | Emitted when new tokens are minted.  |
+| `sft_burn_event`     | `{type: "sft_burn_event", token-id: uint, amount: uint, sender: principal, recipient: principal}`     | Emitted when tokens are burned.      |
```
LNow, line 181:
```suggestion
| Event name           | Tuple structure                                                                                       | Description                          |
|----------------------|-------------------------------------------------------------------------------------------------------|--------------------------------------|
| `sft_transfer` | `{type: "sft_transfer", token-id: uint, amount: uint, sender: principal, recipient: principal}` | Emitted when tokens are transferred. |
| `sft_mint`     | `{type: "sft_mint", token-id: uint, amount: uint, recipient: principal}`                        | Emitted when new tokens are minted.  |
| `sft_burn`     | `{type: "sft_burn", token-id: uint, amount: uint, sender: principal}`     | Emitted when tokens are burned.      |
```

I think we there is no need to keep `_event` in events names nor types.
`sft_burn` don't need `recipient`.

Also probably we should keep these events as small as possible, so maybe we should replace `sender` and `recipient` in transfer event with `from` and `to` ?

radicleart (commented):
at line 88 get-token-uri. as opposed to to string-ascii in 009? - seems a shame to diverge unless there a good reason? Urls dont support non ascii characters so specifying ascii-string is unambiguous as to whether you need to escape non ascii characters in the url.

MarvinJanssen:
That's right. Perhaps using UTF8 could save some space in comparison to punycode. What do you think?

Here's SIP009:
```clojure
(get-token-uri (uint) (response (optional (string-ascii 256)) uint))
```

And here's SIP010:
```clojure
(get-token-uri () (response (optional (string-utf8 256)) uint))
```

radicleart:
I see and yes utf8 is neater than using punycode but would there need to be utf8 characters in the url or is this for future proofing?

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,232 @@
...
+
+Title: Standard Trait Definition for Semi-Fungible Tokens
+
+Author: Marvin Janssen <https://github.com/MarvinJanssen>
+
+Consideration: Technical
+
+Type: Standard
+
+Status: Draft
```
friedger, line 13: A draft does not have a sip number.

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Decimals
+
+`(get-decimals ((token-id uint)) (response uint uint))`
+
+Returns the decimal places of a token type. This is purely for display reasons, where external applications may read this value to provide a better user experience. The ability to specify decimals for a token type can be useful for applications that represent different kinds of assets using one SFT. For example, a game may have an in-game currency with two decimals and a fuel commodity expressed in litres with four decimals.
+
+### Token URI
+
+`(get-token-uri ((token-id uint)) (response (optional (string-utf8 256)) uint))`
```
friedger, line 88: Why string-utf8 here? It can cause some problems, see e.g. https://www.w3.org/International/wiki/IRIStatus
I propose to follow sip-009 instead of sip-010.

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,232 @@
...
+
+Returns an optional UTF8 string that is a valid URI which resolves to this token type's metadata. These files can provide off-chain metadata about that particular token type, like descriptions, imagery, or any other information. The exact structure of the metadata is out of scope for this SIP and may be defined in a future SIP. However, the metadata file should be in JSON format and should include a `version` property containing a string:
```
friedger, line 90: The metadata SIP is in accepted status by now.

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,232 @@
...
+
+```JSON
+{
+	"version: "v1"
```
friedger, line 94: it should be "version": 1

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,232 @@
...
+`(transfer ((token-id uint) (amount uint) (sender principal) (recipient principal)) (response bool uint))`
+
+Transfer a token from the sender to the recipient. It is recommended to leverage Clarity primitives like `ft-transfer?` to help safeguard users. The function should return `(ok true)` on success or an `err` response containing an unsigned integer on failure. The failure codes follow the existing conventions of `stx-transfer?` and `ft-transfer?`.
+
+| Error code | Description                                      |
+|------------|--------------------------------------------------|
+| `u1`       | The sender has insufficient balance.             |
+| `u2`       | The sender and recipient are the same principal. |
+| `u3`       | Amount is `u0`.                                  |
+| `u4`       | The sender is not the `tx-sender`.               |
```
friedger, line 112: `u4` is only for tokens that require the sender to be the tx-sender. This is a very restrictive property that is not wanted for all tokens, see e.g. `operable` trait.

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Bulk transfers
+
+`(transfer-many ((transfers (list 200 {token-id: uint, amount: uint, sender: principal, recipient: principal}))) (response bool uint))`
+
+Transfer many tokens in one contract call. Each transfer should follow the exact same procedure as if it were an individual `transfer` call. The whole function call should fail with an `err` response if one of the transfers fails.
+
+### Bulk transfers with memos
+
+`(transfer-many-memo ((transfers (list 200 {token-id: uint, amount: uint, sender: principal, recipient: principal, memo: (buff 34)}))) (response bool uint))`
```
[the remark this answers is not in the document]
friedger: No, but let's add the tx costs for a basic implementation (with Stacks 2.05+ cost functions)

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,232 @@
...
+	)
+)
+```
+## Events
+
+Semi-fungible token contracts should emit custom events in certain situations via `print`. These events should be emitted after any built-in token events (such as those emitted by `ft-transfer?`) and before the memo in the case of `transfer-memo` and `transfer-many-memo`.
+
+| Event name           | Tuple structure                                                                                 | Description                          |
+|----------------------|-------------------------------------------------------------------------------------------------|--------------------------------------|
+| `sft_transfer`       | `{type: "sft_transfer", token-id: uint, amount: uint, sender: principal, recipient: principal}` | Emitted when tokens are transferred. |
```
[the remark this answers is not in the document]
friedger: maybe `action`

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Use of post conditions
+
+The Stacks blockchain includes a feature known as Post Conditions. By defining post conditions, users can create transactions that include pre-defined guarantees about what might happen in a contract. These post conditions can also be used to provide guarantees for custom fungible and non-fungible tokens that were defined using built-in Clarity primitives.
+
+However, since there are no Clarity primitive counterparts for semi-fungible tokens, post conditions can only safeguard users on a basic level. Semi-fungible token contracts that are implemented using the Clarity primitive `define-fungible-token` give users the ability to make assertions against the total number of tokens transferred in any single call. It does not, however, provide any securities as to the type of token transferred.
+
+Future research may be conducted to see if a custom non-fungible definition using `define-non-fungible-token` can help on this front, perhaps by minting and burning in one go, or by defining a more complex internal token ID. Another solution may be if a future Stacks version allows users to define post conditions for custom `print` events.
```
friedger, line 209: This paragraph should be not part of the sip, maybe reference some thought in related work.

saad-s:
[Here](https://github.com/tintash/clarity-examples/tree/master/006-sft/coc) I created a game contract as a reference implementation, loosely based on Clash of Clans using this standard.

sips/sip-013/sip-013-semi-fungible-token-standard.md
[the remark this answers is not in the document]
MarvinJanssen: I took a "worst case" scenario with a list of 200 items, all with a 34 byte buffer and calculated the costs with `clarinet 0.20.0`. It looks like we exhaust `read_count`:

```
+----------------------+-----------+------------+
|                      | Consumed  | Limit      |
+----------------------+-----------+------------+
| Runtime              | 117399567 | 5000000000 |
+----------------------+-----------+------------+
| Read count           | 12003     | 7750       |
+----------------------+-----------+------------+
| Read length (bytes)  | 873618    | 100000000  |
+----------------------+-----------+------------+
| Write count          | 8000      | 7750       |
+----------------------+-----------+------------+
| Write length (bytes) | 870000    | 15000000   |
+----------------------+-----------+------------+
```

So `(7750-3)/60` puts us at a maximum list length of `129.1166666667`, let us say `128` to be safe.

I'll do the same for the `send-many` function without memo and see how far I get.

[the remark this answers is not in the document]
friedger: A transaction should not take more than 80% of the block limit, otherwise it will be rejected, I think.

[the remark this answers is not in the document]
MarvinJanssen: Source on this? If so, then `((7750*0.8)-3)/60 = 103.2833333333` or simply `100` to be safe.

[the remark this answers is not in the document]
MarvinJanssen: Also, we could stick with a larger number to future-proof the SIP. I imagine costs will be lowered again at some point in the future. What do you think?

MarvinJanssen:
Updated the SIP013 document based on feedback.

@fiftyeightandeight would appreciate your feedback since your project is heavily leveraging SIP013. Did you run into any issues or limitations? Also note that the trait has changed a little; namely, `string-ascii` for `get-token-uri` and a new limit of `100` instead of `200` for the `send-many` functions.

fiftyeightandeight:
We have a number of contracts implementing SIP013 (an example of which is [here](https://github.com/alexgo-io/alex-v1/blob/main/clarity/contracts/yield-token/yield-usda.clar)).
It helps maintain our code base compact (by aggregating similar contracts into a single contract), which is immensely helpful.
The proposal currently implements `transfer` (with/out memo) slightly different from SIP010, which may be debated (I am in favour of the approach SIP013 is taking).
It also has, necessarily perhaps, some compatibility issues with SIP010 (for example, `get-balance` vs. `get-overall-balance`), but I suppose this is more to do with wallet support, rather than the design of the proposal.
It would be interesting if Clarity would ever support polymorphism, but I assume this is not going to happen.

MarvinJanssen:
Thanks for the quick response @fiftyeightandeight, immensely insightful. Would you mind giving a high level explanation of how ALEX utilises this SIP and what the upsides are in a few sentences? I think it is a nice implementation that warrants some background for other curious developers.

fiftyeightandeight:
Hi @MarvinJanssen , we at ALEX offer fixed-rate / fixed-term lending/borrowing. The rate of such lending/borrowing is determined collectively and dynamically by the ALEX community (by buying/selling the underlying contract). 
The term of such lending/borrowing is also determined collectively by the ALEX community, but is fixed at the time of such contract being created. For example, we may create a lending contract which expires on Jan 1, 2022 (i.e. fixed term), whose lending rate to expiry (i.e. fixed rate to expiry at a given point of time) is determined dynamically by the community.

From the Clarity contract perspective, this means you have a number of substantially similar contracts (e.g. lending contract) with slightly different configuration (i.e. expiry).

Before we adopted this SIP, we represented each of these contracts as a standalone contract, essentially copying and pasting an older contract, changing its name and updating it with a new expiry. You can see [here what we had before we adopted this SIP](https://github.com/alexgo-io/alex-v1/tree/legacy/clarity/contracts/yield-token). It is, put it mildly, not great.

After we adopted this SIP, we can aggregate all similar contracts (for example, `yield-usda-xxxx.clar`) into a single contract (for example, `yield-usda.clar`) without losing information about the expiry (in our case, `token-id` == `expiry` given its uniqueness). This immediately makes the code base much more compact while vastly reducing the margin of error. You can see [how we now handle `yield-token` here](https://github.com/alexgo-io/alex-v1/tree/main/clarity/contracts/yield-token).

MarvinJanssen:
@fiftyeightandeight thanks for the explanation. Once (if) this SIP settles, I will update the reference repository with a list of interesting implementations like yours.

0xAsteria:
This is super cool. I'm working on Fractal - an NFT fractionalisation protocol; blog [here](https://asteria.id/posts/fractals), testnet v1.5 [here](https://explorer.syvita.org/txid/0x27d6bf69ffaa3c32914521e2913920ce28da90cda48e29af1582ee02785035b7?chain=testnet). Could we make some changes to this standard around memos? I don't think we need two functions for with/without memo. Wasn't aware of this proposal before working on this, so v2.0 of Fractal would support this standard.

LNow:
Just like you can transfer STX without providing memo, you should be able to do so with both FT and NFT, therefore I think separate functions should stay.

On a side note - @0xAsteria you have a huge security hole in your contract (unguarded trait + as-contract).

0xAsteria:
> you have a huge security hole in your contract (unguarded trait + as-contract).


haha funny you pointed this out, I noticed it this morning when reworking a few things for 2.0. haven't worked out a fix though

MarvinJanssen:
As @LNow said (also, welcome back!), I also think it should stay as separate functions. Effectively, most token transfers do not include a memo and the naming also mirrors the upcoming Stacks 2.1 `stx-transfer-memo?` complement to `stx-transfer?`. However, if you have a compelling case we have not considered then feel free to share it.

Fractional NFTs are basically a type of SFT. I created two (untested) example contracts on how it can be done with SIP013:

1. Wrapping SIP009, where the owner of the original NFT can split it into fractions that can only be recombined by owning all fractions. The original NFT is held by the SFT contract in the mean time.

https://github.com/MarvinJanssen/stx-semi-fungible-token/blob/main/contracts/examples/fractional-sip009-sft.clar

2. A SIP013 SFT contract that takes the SIP009 out of the equation. If an NFT creator foresees that owners might in the future decide to split their NFTs, might as well have that functionality built-in. A SFT token-id/supply combination is effectively an NFT if the supply is 1. Thus, SFTs are initially minted with a supply of 1 in this case, and whoever owns the total supply of a particular token ID can (re)fractionalise it to any supply of his or her choosing.

https://github.com/MarvinJanssen/stx-semi-fungible-token/blob/main/contracts/examples/fractional-nft.clar

Was also going to point out the unguarded trait reference but @LNow beat me to it 😁.

LNow:
> > you have a huge security hole in your contract (unguarded trait + as-contract).
> 
> haha funny you pointed this out, I noticed it this morning when reworking a few things for 2.0. haven't worked out a fix though. any ideas there?

Whitelisting or sandboxing.
Whitelisting is easy to implement and used by most (if not all) NFT marketplaces build on Stacks.
Sandboxing is more complicated (only tiny bit) and requires multiple contracts to be deployed, but at the end of the day it allows to build permission less systems.
https://github.com/LNow/clarity-notes/blob/main/security/as-contract.md

LNow:
@MarvinJanssen have you though about using both NFT and FT at the same time?
NFT could be tied directly to `type` and would represent ownership of particular `type.
And FT would represent amount.

Transferring `X` amount of `Y` type would result in:
 - transferring `X` amount of FT to recipient
 - transferring NFT associated with type `Y` to contract
 - burning NFT if user transferred all or transferring NFT back to user

That way user will see on the transaction confirmation screen that `X` units will be transferred out of user wallet, NFT `Y` will be transferred to contract and eventually returned.

MarvinJanssen:
I assume your last sentence should read:

>burning NFT if user transferred all or transferring NFT back to **[contract]**

Yes I thought of a few different ways of including an NFT—even introducing a more complex NFT type definition with a tuple. It would be really useful to be able to do post conditions but I couldn't come up with one that didn't have downsides or would clash in one situation or another. Can you share a bit more on what you think it could look like?

LNow:
No. NFT is transferred back to user.

Main logic is the same as in your examples. FT represents amount, balance per type is stored in map.
The only difference is that users receives unique NFT that represents ownership of one specific token type. And this NFT is not transferable. 

For example user would have one NFT for gold, one for sword, one for helmet, one for viking helmet etc.
ID of each NFT is unique. My sword NFT will have different ID then your sword NFT.

At first I thought that this idea is stupid and doesn't bring any value to what you proposed. But at some point in the future wallets might be able to display basic NFT metadata. And when that happen users will see exactly whether what they are transferring is exactly what they wanted to transfer or not.

In this PoC NFT is always returned back to sender, so it can be treated as mark that you have or had tokens of a specific type.

```clojure
(define-constant CONTRACT_ADDRESS (as-contract tx-sender))

(define-fungible-token sft-value)
(define-non-fungible-token sft-token uint)

(define-constant ERR_INSUFFICIENT_BALANCE (err u1000))
(define-constant ERR_UNAUTHORIZED (err u1001))

(define-data-var lastNftId uint u0)


(define-map TypeSupply
    uint ;; TypeId
    uint ;; Supply
)

(define-map OwnerTypeToToken
    { owner: principal, typeId: uint }
    { tokenId: uint, balance: uint }
)

(define-read-only (get-balance (typeId uint) (who principal))
    (ok (get-balance-uint typeId who))
)

(define-read-only (get-balance-uint (typeId uint) (who principal))
    (get balance (default-to {balance: u0} (map-get? OwnerTypeToToken { owner: who, typeId: typeId})))
)

(define-read-only (get-overall-balance (who principal))
    (ok (ft-get-balance sft-value who))
)

(define-read-only (get-total-supply (typeId uint))
    (ok (get-total-supply-uint typeId))
)

(define-read-only (get-total-supply-uint (typeId uint))
    (default-to u0 (map-get? TypeSupply typeId))
)

(define-read-only (get-overall-supply)
    (ok (ft-get-supply sft-value))
)

(define-read-only (get-decimals (typeId uint))
    (ok u0)
)

(define-read-only (get-token-uri (typeId uint))
    (ok none)
)

(define-public (transfer (typeId uint) (amount uint) (sender principal) (recipient principal))
    (let
        (
            (tokenOwnedBySender (unwrap! (map-get? OwnerTypeToToken { owner: sender, typeId: typeId}) ERR_INSUFFICIENT_BALANCE))
            (tokenOwnedByRecipient (map-get? OwnerTypeToToken { owner: recipient, typeId: typeId}))
        )
        (asserts! (is-eq tx-sender sender) ERR_UNAUTHORIZED)
        (asserts! (>= (get balance tokenOwnedBySender) amount) ERR_INSUFFICIENT_BALANCE)

        ;; transfer NFT to contract
        (try! (nft-transfer? sft-token (get tokenId tokenOwnedBySender) sender CONTRACT_ADDRESS))
        
        ;; transfer FT to recipient
        (try! (ft-transfer? sft-value amount sender recipient))

        ;; update recipient balance and mint NFT if necessary
        (match tokenOwnedByRecipient token
            (map-set OwnerTypeToToken
                { owner: recipient, typeId: typeId }
                { tokenId: (get tokenId token), balance: (+ (get balance token) amount) }
            )            
            (let
                ((newNftId (+ (var-get lastNftId) u1)))
                (try! (nft-mint? sft-token newNftId recipient))
                (map-set OwnerTypeToToken
                    { owner: recipient, typeId: typeId }
                    { tokenId: newNftId, balance: amount }
                )
                (var-set lastNftId newNftId)
            )        
        )

        ;; update owner balance
        (map-set OwnerTypeToToken 
            { owner: sender, typeId: typeId }
            { tokenId: (get tokenId tokenOwnedBySender), balance: (- (get balance tokenOwnedBySender) amount) }
        )    

        (try! (as-contract (nft-transfer? sft-token (get tokenId tokenOwnedBySender) CONTRACT_ADDRESS sender)))


        (ok true)
    )
)

(define-public (transfer-memo (typeId uint) (amount uint) (sender principal) (recipient principal) (memo (buff 34)))
    (begin
		(try! (transfer typeId amount sender recipient))
		(print memo)
		(ok true)
	)
)

(define-public (mint (typeId uint) (amount uint) (recipient principal))
    (let
        (
            (tokenOwnedByRecipient (map-get? OwnerTypeToToken { owner: recipient, typeId: typeId}))
            (totalSupply (get-total-supply-uint typeId))
        )
        ;; not guarded by anything special, as this is only a PoC

        ;; update balance and mint NFT if necessary
        (match tokenOwnedByRecipient token
            (map-set OwnerTypeToToken
                { owner: recipient, typeId: typeId }
                { tokenId: (get tokenId token), balance: (+ (get balance token) amount) }
            )            
            (let
                ((newNftId (+ (var-get lastNftId) u1)))
                (try! (nft-mint? sft-token newNftId recipient))
                (map-set OwnerTypeToToken
                    { owner: recipient, typeId: typeId }
                    { tokenId: newNftId, balance: amount }
                )
                (var-set lastNftId newNftId)
            )        
        )
        ;; mint requested amount
        (try! (ft-mint? sft-value amount recipient))
        (map-set TypeSupply typeId (+ totalSupply amount))

        (ok true)
    )
)
```

MarvinJanssen:
Right, and how would you know the internal token ID so you can construct your post condition? Read it first via a read-only call? Not insurmountable but seems inconvenient. 

Are these roundtrip NFTs fine for post conditions? I assume so but never tried.

We could also define the NFT as this, perhaps:

```
(define-non-fungible-token sft-token {token-id: uint, owner: principal})
```

Then it is easier to determine what the token ID will be for post conditions. Every transfer would require the user to send that NFT to the contract which it will later get back. Or it could a burn-and-mint in one go, instead of a transfer. Also never tried how post conditions respond to that.

radicleart:
Could the fractionalisation be done another way - e.g. using Judes nftree to fractionalise a single NFT file - in a similar way to how dropbox builds a merkle tree from blocks of a single file in order to make the download more efficient. Each chunk/block is already backed by an FT (tickets) equating to the value of that chunk w.r.t. the whole?

I think this in similar way to the Myceanas project which tokenised high value artworks to try to democratise artwork ownership?

LNow:
> We could also define the NFT as this, perhaps:
> 
> ```
> (define-non-fungible-token sft-token {token-id: uint, owner: principal})
> ```

Yes, good point. As this NFT is not a normal NFT that you can trade with other user it doesn't have to be compliant with SIP-009, therefore custom/complex ID should not be a big deal. And it eliminates extra read-only call from the workflow. :+1: 

With regards to post-conditions - they will work perfectly fine.

MarvinJanssen:
@LNow would it be that simple? https://github.com/MarvinJanssen/stx-semi-fungible-token/commit/309a9a9fb6be96543aec3ca36d5ccb0c6a32648e

I always naturally assumed something like this is meaningless because the final state doesn't change and that is what post conditions check. Nonetheless, I never tried it out. We could deploy something on testnet and play with it. (Side-note: the explorer really needs post condition support.)

LNow:
Yes it is that simple.

> I always naturally assumed something like this is meaningless because the final state doesn't change and that is what post conditions check.

Post conditions are very weird and they don't work exactly as they are described in the documentation.

* For burn+mint in one TX the only valid post condition is single `DoesNotOwn`that covers burn operation. 
[example contract](https://explorer.stacks.co/txid/0x43bff72678d1f6707aa7ed2905a1125b19df3c6eb6e6f3d91a6b9ff3d033dd77?chain=testnet)

* For transfer+transfer in one TX (like in my PoC) - you need 2 `DoesNotOwn` post conditions (one for user and one for contract). [example contract](https://explorer.stacks.co/txid/STCAQ4RCYJ30BYKJ9Z6BRGS3169PWZNN892NSS0H.patient-amber-constrictor?chain=testnet)

Explorer, Wallet and documentation needs proper post-condition support.

LNow:
@MarvinJanssen looks ok.
To increase security it would be also good to change how wallet displays NFT post conditions. Right now NFT id is not displayed on TX confirmation screen. So every time you do something with NFT you can be tricked by UI to transfer/swap different (more expensive) token.

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,248 @@
...
+
+Sign-off: 
+
+Layer: Traits
+
+Discussions-To: https://github.com/stacksgov/sips
+
+# Abstract
+
+Semi-Fungible Tokens, or SFTs, are digital assets that sit between fungible and non-fungible tokens. Fungible tokens are directly interchangeable, can be received, sent, and divided. Non-fungible tokens each have a unique identifier that distinguishes them from each other. Semi-fungible tokens have both an identifier and an amount. This SIP describes the SFT trait and provides a reference implementation.
```
jcnelson, line 27: Formatting nit -- can you break lines into 80 columns?  It will make merging patches less conflict-prone.

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Transfer with memo
+
+`(transfer-memo ((token-id uint) (amount uint) (sender principal) (recipient principal) (memo (buff 34))) (response bool uint))`
+
+Transfer a token from the sender to the recipient and emit a memo. This function follows the exact same procedure as `transfer` but emits the provided memo via `(print memo)`. The memo event should be the final event emitted by the contract. (See also the events section of this document below.)
+
+### Bulk transfers
+
+`(transfer-many ((transfers (list 100 {token-id: uint, amount: uint, sender: principal, recipient: principal}))) (response bool uint))`
```
jcnelson, line 132: If there's going to be a cap of 100 transfers here, why not just call this `transfer-100`?

jcnelson: Same for `transfer-many-memo` below.

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,248 @@
...
+
+The recommended native asset primitives to use:
+
+- `define-fungible-token`
+- `ft-burn?`
+- `ft-get-balance`
+- `ft-get-supply`
+- `ft-mint?`
+- `ft-transfer?`
+- And their NFT equivalents.
```
jcnelson, line 204: Might as well list them too ;)

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,248 @@
...
+
+All of the functions in this trait return the `response` type, which is a requirement of trait definitions in Clarity. However, some of these functions should be "fail-proof", in the sense that they should never return an error. The "fail-proof" functions are those that have been recommended as read-only. If a contract that implements this trait returns an error for these functions, it may be an indication of a faulty contract, and consumers of those contracts should proceed with caution.
+
+## Use of post conditions
+
+The Stacks blockchain includes a feature known as Post Conditions. By defining post conditions, users can create transactions that include pre-defined guarantees about what might happen in a contract. These post conditions can also be used to provide guarantees for custom fungible and non-fungible tokens that were defined using built-in Clarity primitives.
+
+There are no Clarity primitive counterparts for semi-fungible tokens, but contract developers can leverage a combination of post conditions to achieve the same result.
+
+There are two factor that should be checked by post conditions:
```
jcnelson, line 218: "factors"

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,248 @@
...
+
+1. The amount of semi-fungible tokens being transferred.
+2. The token ID of the semi-fungible token being transferred.
+
+To that end, it is recommended that developers use both Clarity primitives in their design. Semi-fungible token contracts can achieve complete post condition coverage by using both `define-fungible-token` and `define-non-fungible-token`.
+
+For strategies on how to best guard a semi-fungible token contract with post conditions, see the reference implementation at the end of this document.
```
jcnelson, line 225: Can you also include a tarball of the source code as a supplemental file, per SIP-000?

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,248 @@
...
+
+# Related work
+
+## Ethereum ERC1155
+
+- [EIP-1155](https://eips.ethereum.org/EIPS/eip-1155)
```
jcnelson, line 231: Can you briefly compare/contrast the approach taken in EIP-1155 to this approach?  Is there more to it than the fact that a particular factoring of SFTs can support post-conditions?

sips/sip-013/sip-013-semi-fungible-token-standard.md
jcnelson: Also, I think it's in-scope how to represent SFTs as a combination of fungible and non-fungible tokens.  Even though we can't enforce this via a trait, wallets can require that the AST of the contract or even certain function implementations have certain properties (for example, this SIP could require that the core SFT mint/transfer/burn functionality adhere to a code _template_ which the wallet could inspect).  In addition, wallet implementations will still need to know exactly how to generate post-conditions on SFTs, and this is the perfect place to specify this.

jcnelson:
Thanks for sending this in @MarvinJanssen!  Left feedback.

sips/sip-013/sip-013-semi-fungible-token-standard.md
MarvinJanssen: I was mirroring the naming convention of the `send-many` STX tool. I can see how adding the total can be useful so that a `transfer-N` can be introduced by a future SIP. Having said that, I want to see if there are any contracts on mainnet that already implement this draft SIP in its current form.

Additionally, I decided to change it back to its original value of `200`. While it might not be possible to call it with a full list right now, if we anticipate this to become a standard and used for many years to come, one can assume that we can see another runtime cost adjustment at some point in the future.

sips/sip-013/sip-013-semi-fungible-token-standard.md
MarvinJanssen: Yep that sounds great. I was actually looking at providing a `clarinet run` script that checks if a contract adheres to such a template. It looks like `clarinet` no longer exposes the AST which makes it a bit more tricky.

sips/sip-013/sip-013-semi-fungible-token-standard.md
MarvinJanssen: Will do.

sips/sip-013/sip-013-semi-fungible-token-standard.md
jcnelson: Yeah, you could make this number arbitrarily large for forward-compatibility.  The block limit probably won't increase by more than a factor of 100, but you never know.

sips/sip-013/sip-013-semi-fungible-token-standard.md
jcnelson: Wait really?  That's annoying.  The node will report the contract's interface [1] right now, but not the AST.  I suppose the wallet could do some simple template-matching simply by fetching the contract source code and matching function bodies to a regex, but would get brittle really fast (but maybe that's what we want -- if the code isn't *exactly* what a regex says, then it's assumed nefarious).

[1] https://hirosystems.github.io/stacks-blockchain-api/#operation/get_contract_interface

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,248 @@
...
+
+The `transfer` function should emit a special transfer event, as detailed in the Events section of this document.
+
+### Transfer with memo
+
+`(transfer-memo ((token-id uint) (amount uint) (sender principal) (recipient principal) (memo (buff 34))) (response bool uint))`
+
+Transfer a token from the sender to the recipient and emit a memo. This function follows the exact same procedure as `transfer` but emits the provided memo via `(print memo)`. The memo event should be the final event emitted by the contract. (See also the events section of this document below.)
+
+### Bulk transfers
```
jcnelson, line 130: Thinking more about `transfer-many` and `transfer-many-memo`, I'm not 100% persuaded that they need to be part of the trait definition.  It might make more sense to publish a supplemental contract that takes a `sip-013` trait implementation and implements `transfer-many` and `transfer-many-memo` by iteratively calling the `transfer` function.  Then, each SIP 013 contract doesn't have to re-implement the same boilerplate each time (which exposes them to bugs).

sips/sip-013/sip-013-semi-fungible-token-standard.md
MarvinJanssen: Yes. Would be pretty cool if the wallet can do it as well. I opened an issue for Clarinet here https://github.com/hirosystems/clarinet/issues/268. The mere ability to check if `ft-transfer?` is called at some point would already be a large benefit.

sips/sip-013/sip-013-semi-fungible-token-standard.md
MarvinJanssen: I would prefer that too if we had a `contract-call?` that does not reload the target contract into memory each time. Adding a `transfer-many` to the token contract is a lot more efficient at this point.

MarvinJanssen:
>I unresolved a couple comments because it helps me remember which ones need to be addressed before advancing this to Accepted status. Hope you don't mind ;)

Sure, just makes it a little harder for me to figure out which ones I had already addressed but not yet published. 😁

sips/sip-013/sip-013-semi-fungible-token-standard.md
jcnelson: I opened an issue for this here: https://github.com/stacks-network/stacks-blockchain/issues/3061

I agree that this is a problem, and it should ultimately be solved in the Stacks node by only charging the transaction once per loaded contract (or even charging the *block* once per loaded contract).  But, do you think the SIP could outline a migration plan so that `transfer-many` and `transfer-many-memo` could be deprecated in favor of this approach once Stacks 2.1 incorporates this fix?

LNow: What if we define two separate traits?
One for all base SFT function, and another one for `transfer-many` and `transfer-many-memo`.

That way we'll be able to:
- support only base functions using `SIP-013-trait` 
- support base functions and transfer-many (all in one contract) using `SIP-013-trait` and `SIP-013-transfer-many-trait` 
- support base functions in token contract and transfer-many in different contract

MarvinJanssen: That's exactly what I was thinking too. We can do the same for SIP010.

MarvinJanssen:
Happy to have been working on this since moving to the Clarity Lab.

Updated per discussions and comments above @jcnelson @friedger @radicleart @LNow.

- This SIP now contains two trait definitions, the main `sip013-semi-fungible-token-trait` and a new optional `sip013-send-many-trait`.
- The document now refers to a tar file `SIP-013-001.tar.gz`, which will contain the reference repository once we move into the accepted stage.
- 80 column rule applied.
- For those that didn't see it yet, the reference repository contains some more example implementations, including two "fractional NFT" contracts. https://github.com/MarvinJanssen/stx-semi-fungible-token/tree/main/contracts/examples

MarvinJanssen:
Great, done!

I do have a few more open points / questions:

1. In the current version, Post Condition coverage sounds more advisory than mandatory. I am not sure if I prefer making the mechanisms in the reference implementation mandatory; specifically (a) the fungible token transfer to cover the amount, and (b) the mint-and-burn to cover the token ID. It would be good to get some more input and weigh the pros and cons. Also to add there is no good way to enforce these mechanisms. Traits obviously do not cover them and Clarinet would need something like https://github.com/hirosystems/clarinet/issues/268 to make such deep analysis possible.

2. Suggestions on good Activation requirements that is clearly falsifiable. To me the standard seems "obvious" but is "_N_ number of SIP013 token contract deployed by block height _H_" sufficient?

friedger:
1) Instead of enforcing post conditions, I would define a  behavioural specification between the functions, like the balance function must return the same amount until transfer, transfer-memo was called. Tokens can be implemented with many native assets in various ways, therefore a specification for post conditions would not be too restrictive.

2) My concern would be that, currently, there is not a lot of demand for semi fungible tokens. Therefore, activation is not a question of number of deployed contracts. You could add that no alternative SIP draft was published after the deployment of a compliant contract and after some many blocks.

jcnelson:
> You could add that no alternative SIP draft was published after the deployment of a compliant contract and after so many blocks.

The risk here is that we pick a timeout that's too low, and then we're stuck with it.  What if we did a combination?

* There are N contracts by block B that implement this trait
* There are no revisions to this SIP's trait after C blocks

We pick N, B, and C accordingly, where C is significantly bigger than B.

Also, I think N should be at least 1.  I think that if no one implements this trait -- not even the SIP authors -- then there's no reason to have the SIP at all.

radicleart:
Sounds good. 

ALEX already have one implementation and Clarity Lab has another on testnet so N should be minimum 3.

fiftyeightandeight:
> Sounds good.
> 
> ALEX already have one implementation and Clarity Lab has another on testnet so N should be minimum 3.

Yes, our [implementation](https://github.com/alexgo-io/alex-v1/blob/dev/clarity/contracts/yield-token/yield-alex.clar) is expected to go live later this month.

314159265359879:
> > You could add that no alternative SIP draft was published after the deployment of a compliant contract and after so many blocks.
> 
> The risk here is that we pick a timeout that's too low, and then we're stuck with it. What if we did a combination?
> 
>     * There are N contracts by block B that implement this trait
>     * There are no revisions to this SIP's trait after C blocks
> 
> We pick N, B, and C accordingly, where C is significantly bigger than B.
> 
> Also, I think N should be at least 1. I think that if no one implements this trait -- not even the SIP authors -- then there's no reason to have the SIP at all.

I like this idea of Jude and Mike suggestion for N, minimum of 3. 

**Little history of SIP010 and SIP009**
Block C should be a couple months in the future in my view. I see SIP010 (FT trait) was flagged as activation in progress on [March 12th 2021](https://github.com/stacksgov/sips/pull/5#issuecomment-797636750), and the last revisions were made [May 29th](https://github.com/stacksgov/sips/commit/4dd02c8be1e83ee385ea47275f2bd1ca09df9871) (changing the March 12th trait for the third/last time). Activation threshold (just N contracts deployed with trait) was met [August 31st 2021](https://github.com/stacksgov/sips/pull/5#issuecomment-909436016). 
For SIP009 (NFT trait), activation was on [March 17th 2021](https://github.com/stacksgov/sips/pull/3#issuecomment-801232263), last changes made on March 19th and it was activated on May 20th 2021.

Allowing 2-3 months for revisions, is that reasonable? In my view SIP010 (and SIP009) were special because they were made in a very hectic time when Stacks 2.0 was just launched and a lot was happening, this also lead to some _senior developers_ noticing inconsistencies later in the process. Suggestions/discussion for block B and block C

Block B, (six months) 25920 BTC blocks **after activation**, or should this be closer to C? generally before or after C?
Block C, three months **after activation**.

MarvinJanssen:
I'm happy to move this forward. I think we can keep the post condition stuff as-is. However, we need to add activation criteria before we put it up for review. I'll sleep on it and submit the update tomorrow.

radicleart:
@MarvinJanssen adding activation criteria is still outstanding.

MarvinJanssen:
@radicleart updated Bitcoin block heights for activation and submitted.

rafaelcr:
@MarvinJanssen is there a way for a caller to get the last token ID similar to SIP-009's `last-token-id`? If not, could it be added to this SIP? If I'm not mistaken, `get-overall-supply` would *not* give the caller the number of token classes but rather the sum of all token class supplies, right?

This would be very useful for token metadata indexing services so they could iterate through all the available token classes for a SIP contract

jcnelson:
Thanks for adding the `Activation` section!  Are you ready for a review from the technical CAB?

MarvinJanssen:
> Thanks for adding the `Activation` section! Are you ready for a review from the technical CAB?

Yes it is ready for review @jcnelson. Missed this one.

jcnelson:
@kantai Can you take a look at this SIP when you get a free moment? Thanks!

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,388 @@
...
+not what is expected.
+
+Since it is possible for smart contracts to own tokens, it is recommended to
+check for both `tx-sender` and `contract-caller` as it allows smart contracts to
+transfer tokens it owns without having to resort to using `as-contract`. Such a
+guard can be constructed as follows:
+
+```clarity
+(asserts! (or (is-eq sender tx-sender) (is-eq sender contract-caller)) (err u4))
+```
```
kantai, line 183: My understanding is that contracts _should_ use `as-contract` to issue a transfer: the goal of `as-contract` was exactly to allow contracts to behave like normal principals when trying to invoke other contracts. Is there a reason that `as-contract` should be avoided?

sips/sip-013/sip-013-semi-fungible-token-standard.md
```diff
@@ -0,0 +1,388 @@
...
+- `define-fungible-token`
+- `ft-burn?`
+- `ft-get-balance`
+- `ft-get-supply`
+- `ft-mint?`
+- `ft-transfer?`
+- `define-non-fungible-token`
+- `nft-burn?`
+- `nft-mint?`
+- `nft-transfer?`
```
kantai, line 263: Can you add a minimal example to the text here for using the `ft-transfer?` and `nft-transfer?` methods to implement a `transfer` method?  I know that the reference implementations will include this, but I think a minimal example inline will help readers see how they can use native asset functions to preserve post-condition functionality.

sips/sip-013/sip-013-semi-fungible-token-standard.md
radicleart: Yes. Its explained in a new SIP suggestion for safe transfer of FT/NFT asset - [see this PR](https://github.com/stacksgov/sips/pull/84). 

The gist is that forcing the marketplace contract to use `as-contract` to call into the NFT contract leaves it exposed to malicious behaviour by the NFT contract (I think @LNow first pointed this out) - e.g. transferring other assets owned by the calling contract. Safe transfer requires check to be against either tx-sender or contract-caller.

kantai: Oh, yes, I remember now. Thanks!

sips/sip-013/sip-013-semi-fungible-token-standard.md
radicleart: @MarvinJanssen  I think you have control of the fork - am not sure how to push a commit to this PR?

MarvinJanssen: @kantai just pushed an update. Is it what you are looking for?

MarvinJanssen:
Pinging @kantai, is it good to move forward?

MarvinJanssen:
Pinging @kantai, is it good to move forward?

Hero-Gamer:
> Pinging @kantai, is it good to move forward?

Hi @MarvinJanssen on the weekly SIP call #17 was said that Aaron is out of action until January afaiu.
On that call, Brice volunteered to review this for you guys @obycode while Aaron is away. :)

sips/sip-013/sip-013-semi-fungible-token-standard.md
obycode: As mentioned, I'm filling in for @kantai here. This looks good to me.

obycode:
> > Pinging @kantai, is it good to move forward?

> Hi @MarvinJanssen on the weekly SIP call https://github.com/stacksgov/sips/issues/17 was said that Aaron is out of action until January afaiu.
On that call, Brice @obycode kindly volunteered to review this for you guys while Aaron is away. :)

Looks good to me. It seems Aaron's concerns were addressed.

friedger:
Let's move to ratified!

MarvinJanssen:
It appears that all activation criteria have been met, so can we move straight to `Ratified`? @friedger @obycode?

jcnelson:
Yup!

jcnelson:
Please switch the status to `Ratified` and add both myself and @obycode's contact information to the `Sign-off` section, and I'll merge it.  Great job everyone!

MarvinJanssen:
@jcnelson done!

PR #4, opened 2021-01-04
fix: adds existing SIPs from stacks-blockchain

hstove:
This PR adds all existing SIPs from https://github.com/blockstack/stacks-blockchain . No modifications have been made.

Issue #1, opened 2021-01-04
Fungible token standard


agraebe:
As per previous discussions ([see Paper doc](https://paper.dropbox.com/doc/Working-group-Stacks-Token-Standard--BCnkvBjpEg~uI25yo8oFT1J2AQ-Hmix9Te1J9x7qYmTvlrqZ)), @psq will create a proposal for a fungible token standard.

I'm creating this to keep track of the conversations

friedger:
Defined and ratified. Can be closed.

PR #44, opened 2021-10-20
SIP-016 Metadata for Tokens

friedger:
This PR
* adds a sip for metadata of token, this includes NFTs that are compliant to SIP-009, FTs that are compliant to SIP-010, or semi-fungible tokens.
* addresses #17


friedger:
The SIP contains a JSON schema definition written in JSON. 
Makers of $65m NFT "First 5000 days" by Beeple mixed schema definition with schema values. Metadata should **NOT** look like this. While valid according to the sip, providing values in "description" fields is counter-intuitive.

```json
{"title": "EVERYDAYS: THE FIRST 5000 DAYS", 
"name": "EVERYDAYS: THE FIRST 5000 DAYS", 
"type": "object", 
"imageUrl": "https://ipfsgateway.makersplace.com/ipfs/QmZ15eQX8FPjfrtdX3QYbrhZxJpbLpvDpsgb2p3VEH8Bqq", 
"description": "I made a picture from start to finish every single day from May 1st, 2007 - January 7th, 2021.  This is every motherfucking one of those pictures.", "attributes": [{"trait_type": "Creator", "value": "beeple"}], 
"properties": 
   {"name": {"type": "string", "description": "EVERYDAYS: THE FIRST 5000 DAYS"}, 
    "description": {"type": "string", "description": "I made a picture from start to finish every single day from May 1st, 2007 - January 7th, 2021.  This is every motherfucking one of those pictures."}, 
    "preview_media_file": {"type": "string", "description": "https://ipfsgateway.makersplace.com/ipfs/QmZ15eQX8FPjfrtdX3QYbrhZxJpbLpvDpsgb2p3VEH8Bqq"}, 
    "preview_media_file_type": {"type": "string", "description": "jpg"}, "created_at": {"type": "datetime", "description": "2021-02-16T00:07:31.674688+00:00"}, "total_supply": {"type": "int", "description": 1}, 
    "digital_media_signature_type": {"type": "string", "description": "SHA-256"},
    "digital_media_signature": {"type": "string", "description": "6314b55cc6ff34f67a18e1ccc977234b803f7a5497b94f1f994ac9d1b896a017"}, 
     "raw_media_file": {"type": "string", "description": "https://ipfsgateway.makersplace.com/ipfs/QmXkxpwAHCtDXbbZHUwqtFucG1RMS6T87vi1CdvadfL7qA"}}}
```

Jamil:
Right now stxnft.com goes with a format that looks something like this:
```
{
  "collection": "StacksArmy",
  "name": "StacksArmy Captain #9",
  "image": "ipfs://ipfs/QmPPQBNKohiDYAnjKkdrQcf2LupqixFMEpyvnNWQXz7ozq/StacksArmy%20Captain%20%239.png",
  "attributes": [
    {
      "trait_type": "class",
      "value": "Captain"
    }
  ]
}
```

This is similar to the OpenSea metadata format, with the addition of the additional field `collection`. 

This is because right now SIP-009 doesn't support "collection-level" metadata (`ERC721Metadata` for example, specifies `name()` and `symbol()` functions). I'm not terribly opinionated on whether this should be done at the contract level with an additional function to get the name/symbol, or whether it should be handled in `get-token-uri` as we do now.

To me, the OpenSea metadata format works just fine with this collection-level metadata caveat. I'd support either (a) an extension to SIP-009 allowing for NFT contracts to return a name and symbol (e.g., `get-name` and `get-symbol` functions which take no arguments), or adopting a metadata format that is OpenSea + some field to specify the collection name.

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,163 @@
...
+
+Created: 20 October 2021
+
+License: CC0-1.0
+
+Sign-off:
+
+# Abstract
+
+Non-fungible tokens or NFTs for short are digital assets registered on blockchain with unique identifiers and properties that distinguish them from each other. SIP-009 defines the trait how ownership of an NFT is managed. This SIP aims to provide a flexible standard to attach meta data to NFTs, like descriptions or urls to digital files.
```
kyranjamie, line 23:
```suggestion
Non-fungible tokens or NFTs for short are digital assets registered on blockchain with unique identifiers and properties that distinguish them from each other. SIP-009 defines the trait for how ownership of an NFT is managed. This SIP aims to provide a flexible standard to attach metadata to NFTs, like descriptions or urls to digital files.
```

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,163 @@
...
+
+```
+{
+    "title": "Token Metadata",
+    "type": "object",
+    "required": [],
+    "properties": {
+        "name": {
+            "type": "string",
+            "description": "Identifies the asset to which this token represents"
```
kyranjamie, line 60: Should there be a max length described?

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,163 @@
...
+
+## JSON scheme of Metadata
+
+The JSON blob resolved through the token uri must follow the following JSON schema. If the string `{id}` exists in any JSON value, it MUST be replaced with the actual token id in decimal format, by all client software that follows this standard.
+
+```
+{
+    "title": "Token Metadata",
+    "type": "object",
+    "required": [],
```
kyranjamie, line 56: No properties are required? How about additional properties?

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,163 @@
...
+
+The resolved data must be a JSON blob.
+
+## JSON scheme of Metadata
+
+The JSON blob resolved through the token uri must follow the following JSON schema. If the string `{id}` exists in any JSON value, it MUST be replaced with the actual token id in decimal format, by all client software that follows this standard.
+
+```
+{
+    "title": "Token Metadata",
```
kyranjamie, line 54:
```suggestion
```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "title": "Token Metadata",
```

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+}
+```
+
+### Examples
+
+# Using NFTs in applications
+
+Before presenting metadata to users, application developers should verify whether the metadata is compliant with the applications guidelines.
```
kyranjamie, line 112:
```suggestion
Before presenting metadata to users, application developers should verify whether the metadata is compliant with the application's guidelines.
```

friedger:
@dantrevino properties and attributes should be swapped, shouldn't it? At least that is how opensea defines it: https://docs.opensea.io/docs/metadata-standards#metadata-structure

**attributes** consists of trait_type, display_type and value.

**properties** can be anything.

dantrevino:
> @dantrevino properties and attributes should be swapped, shouldn't it? At least that is how opensea defines it: https://docs.opensea.io/docs/metadata-standards#metadata-structure
> 
> **attributes** consists of trait_type, display_type and value.
> 
> **properties** can be anything.

yup got that backwards

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,295 @@
...
+      "value": "red",
+    },
+    {
+      "trait_type": "strength",
+      "display_type": "number",
+      "value": 99,
+    },    
+  ],
+  "properties": {
+    {
```
Jamil, line 141: is this supposed to be nested or instead:
```
"properties": {
    "collection":  "Foo Collection",
    "collection_size":  "10000" 
}
```

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,295 @@
...
+# License and Copyright
+
+This SIP is made available under the terms of the Creative Commons CC0 1.0 Universal license, available at https://creativecommons.org/publicdomain/zero/1.0/
+This SIP’s copyright is held by the Stacks Open Internet Foundation.
+
+# Introduction
+
+Tokens are digital assets registered on blockchain through a smart contract. A non-fungible token (NFT) is a token that is globally unique and can be identified through its unique identifier. In blockchains with smart contracts, including the Stacks blockchain, developers and users can use smart contracts to register and interact with non-fungible tokens.
+
+Some use cases of NFTs are name registration, digital art, certification, media and enternainment, real-estate. They all require that users associate certain content with an NFT. In general, it is helpful for the users to have a name, sound, image that represents this content.
```
Jamil, line 34: nit: "entertainment"

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,295 @@
...
+
+# Specification
+
+Every SIP-X compliant smart contract in Stacks blockchain must implement one or more functions that return a resolvable/retrievable uri referencing metadata. The metadata provide information for displaying a digital asset to users. This type of function is named "metadata uri functions".
+
+Appendix A contains a list of trait functions that must meet the following requirements for the return value. The appendix can be extended without changing the ratification status of this SIP.
+
+## Return Value of Metadata URI Functions
+
+The return value must be a `some` value if and only if the metadata reference an existing token, otherwise the value must be `none`. Appendix A specifies the exact meaning of "existingt" for each function.
```
Jamil, line 44: nit: "existing"

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,295 @@
...
+
+If a metadata uri function expects a parameter of type `uint` that identifies a token and the resulting strings contains `{id}`, then the `{id}` part must be replaced by the identifier in decimal format given in the function call.
+
+The resolved data of the URI must be a JSON blob.
+
+## JSON scheme of Metadata
+
+The JSON blob resolved through the uri must follow the following JSON schema. 
+
+If metadata were retrieved by a function call containing a token identifier and the string `{id}` exists in any JSON value, it MUST be replaced with the actual token id in decimal format, by all client software that follows this standard.
```
Jamil, line 56: I like this part :)

sips/sip-x/sip-x-nft-metadata.md
friedger: addressed in 661ba2a073b48aa4a45e30251182adff9a2f4057

sips/sip-x/sip-x-nft-metadata.md
friedger: :+1:

sips/sip-x/sip-x-nft-metadata.md
friedger: addressed in 661ba2a073b48aa4a45e30251182adff9a2f4057

sips/sip-x/sip-x-nft-metadata.md
friedger: You are correct. Addressed in 661ba2a073b48aa4a45e30251182adff9a2f4057

friedger:
@Jamil Thank you for your review!

zone117x:
Thanks for initiating this SIP @friedger. Something I'd like to point out is that this SIP does not include requirements needed for implementing NFT metadata support in the current Stacks Blockchain API. See issue https://github.com/hirosystems/stacks-blockchain-api/issues/846

For example, the SIP does not include anything related to making the metadata easily (and at scale) accessible for a Stacks Node event observer such as the API. That would entail something like a `print` (AKA contract log event) requirement to output the NFT metadata URI (possibly more) on NFT mints. It also doesn't address a concern that has been brought up a few times around "refreshing" NFT metadata, for example, a metadata "reveal" sometime after an NFT is minted. That would likely also entail something like a contract log event to inform consumers to "re-resolve" the metadata for a given NFT.

These concerns may be out of scope for this SIP, but I figured it would be good to bring up in case anyone was planning on API support related to this SIP.

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -187,16 +187,42 @@ pt-BR.json
...
+}
+```
+
+Example:
+
+```
+{
+    "type": "string", 
+    "description": "Address of custodian key holder", 
+    "value": "Casa Inc., P.O. Box 20575, Charleston, S.C. 29413, United States."
```
dcsan, line 223: good this is extensible this far, and not too complex... but i would rather focus on digital assets first.

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -187,16 +187,42 @@ pt-BR.json
...
+        "type": {
+            "type": "string",
+            "description": "type of the property"
+        },
+        "description": {
+            "type": "string",
+            "description": "description of the property"
+        },
+        "value": {
+            "type": any,
```
dcsan, line 210: does `type: any` have complications if we wanted to bring this metadata into a contract?
esp given how strongly typed clarity is? you can't even cast a num to an int there.

dcsan:
@friedger 
re raw media - things like real world addresses, maybe thats too much feature creep? personally I'd like to nail digital assets first without adding the kitchen sink.

Did you see [this example](https://github.com/stacksgov/sips/issues/17#issuecomment-968191073) from the solana ecosystem just allowing files to be added. isn't this similar to the spec for [files and folders for IPFS?](https://docs.ipfs.io/concepts/file-systems/#mutable-file-system-mfs)

dcsan:
one other comment, this might not be the place.. but when minting an NFT on a gas-cheap chain like STX, i'd like to add metadata to allow customization of NFTs, and keeping some of that data on-chain.

[SIP009](https://github.com/stacksgov/sips/blob/main/sips/sip-009/sip-009-nft-standard.md) sets out the `transfer` fn but it doesn't take any extra optional parameters.

so while this spec allows for richer metadata, i'm wondering where that metadata lives. 
For simple text metadata, we might want to look at eventually removing the need for external URLs and just exposing it directly from the contract (much like solidity allows for SVG data on-chain). Hence standardized functions for getting data in, and reading it back out, and also a very simple and type-limited version (no 'any') of the spec since Clarity is so strict on types.

sips/sip-x/sip-x-nft-metadata.md
kyranjamie: Is `any` a valid type? https://stackoverflow.com/a/16826141/1141891

sips/sip-x/sip-x-nft-metadata.md
dantrevino: In general disagree with focusing on digital assets only.

dantrevino (approved):

dantrevino:
Thx @friedger other than the `any` comment from @kyranjamie looks good

zone117x:
@friedger you mentioned this PR supersedes https://github.com/stacksgov/sips/pull/18 -- is NFT class-level metadata intended to be out of scope? The previous PR specified a `(get-nft-meta () ... )`.

The API has some initial support for resolving and indexing NFT class metadata (similar to FT metadata), but it's disabled by default, and was pending ratification of the previous NFT metadata PR. See issue https://github.com/hirosystems/stacks-blockchain-api/issues/742

dantrevino:
Actually @zone117x reminds of a good point ... 

One thing that should be added to the SIP, imho, is that a protocol is **_not_** specified.  

https://, ar://, ipfs://, sia://, etc, etc, as well as any future protocols should all be valid.

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,341 @@
...
+| Name | Type | Description |
+|------|------|-------------|
+| `collection` | `string` | collection name the token belongs to. See also Appendix A. |
+| `decimals` | `integer` | number of decimals. See also Appendix A. |
+| `id` | `integer` | identifier for NFTs. See also Appendix A. |
+| `created` | `integer` | creation date of the token in unix timestamp | 
+| `symbol`  | `string` | token symbol |
+| `total_supply`| `integer` | number of total supply, e.g. minted tokens |
+| `ip_document_uri` | `string` | link to document about intellectual property (IP) rights |
+| `external_url` | `string` | url that will view the token on an external site |
```
dcsan, line 337: another requested field is the `animation_url` which is often for a plain file type, but it can also support a URL to a full html page which is usually embedded in an iframe:

https://docs.opensea.io/docs/metadata-standards

Animation_url also supports HTML pages, allowing you to build rich experiences and interactive NFTs using JavaScript canvas, WebGL, and more. Scripts and relative paths within the HTML page are now supported. However, access to browser extensions is not supported.

dcsan (commented):
comment re animation_url

sips/sip-x/sip-x-nft-metadata.md
friedger: This is only an example, even for digital assets there might be meta data that we don't want to specify here like `camera` or `iso`, ...

sips/sip-x/sip-x-nft-metadata.md
friedger: Has been added

sips/sip-x/sip-x-nft-metadata.md
friedger: Has been corrected

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,389 @@
...
+  }
+}
+```
+
+de.json
+
+```
+{
+    "version": 1,
+    "attritbutes: [
```
Jamil, line 173: another minor nit: typo here :)

friedger:
Section about uri schemes has been added @dantrevino 

Metadata function have their own section now @zone117x 
As the metadata is not stored on-chain the api node is maybe not the right place to cache the meta data.

I have added more properties to make the metadata compatible with opensea. I even removed the requirement for a `value` of properties. I refused to add `youtube_url` because in a few years nobody knows what youtube was. I added a generic `files` property taken from metaplex (solana). @dcsan

sips/sip-x/sip-x-nft-metadata.md
Jamil: and below

friedger: fixed in 2bc81e987b43de79a4c91c16272947849e0d142c

dcsan:
I came across this project which is one of the first "on chain" STX generative art projects I've seen:
https://stxplates.xyz/ 

discussion:
https://discord.com/channels/621759717756370964/877879077569171456/911219717610545193

This uses a 'bytes' field to expose some metadata from the contract, used for rendering the generative art. 

Since clarity requires the exact string len, emitting metadata directly from the contract in the tokenURI seems very impractical.
But this `bytes` field approach where the seed for the artwork is always the same length is very practical. This is similar to artblocks which just passes in the last mint hash. It makes the artwork reproducible which is critical for value - not just random stuff.

the contract metadata for Plates is like below, but perhaps this `bytes` field (or `seed` maybe?) could go elsewhere in the final layout.

contract:
https://explorer.stacks.co/txid/0x88ccc25f363d443316befe51b56ceac4a9b8a1c4283b3ce0ed70da93a53722ee?chain=mainnet

returns:
`(ok (some "ipfs://Qmd4zgFFhe85roe9e9zDVzP5GquaRAgbV7KaS524mMVQaz/624.json"))`

so view it here:
https://ipfs.io/ipfs/Qmd4zgFFhe85roe9e9zDVzP5GquaRAgbV7KaS524mMVQaz/624.json

gives 

```json
{
"external_url": "https://stxplates.xyz",
"image": "ipfs://QmNNfMNaH7wJkhhkfpyDyiLy5VEKbncoQ7oknv6ZGqSN1d/624.png",
"description": "STXPLATES is a collection of 10,000 randomly generated, unique, on-chain NFTS for the Stacks community. This is plate 44L5GHE6R",
"name": "44L5GHE6R",
"attributes": [
{
"trait_type": "Text",
"value": "44L5GHE6R"
},
{
"trait_type": "Text Color",
"value": "#F90F1B"
},
{
"trait_type": "Background Color",
"value": "#B5E3FC"
},
{
"trait_type": "Border Color",
"value": "#303649"
}
],
"artist": "@txp0st",
"collectionSize": 10000,
"index": 624,
"bytes": "666666666666666666666688888888888888888886688822888822882888886688282888282882888886682882882882882888886682222282222282888886688882888882882222286688888888888888888886682222288222282888286682888882888882888286682222882882282888286688888282888282222286682222888222282888286688888888888888888886682222288222282222886682888882888882888286682228882222882888286682888882888282222886682222288222882888286688888888888888888886666666666666666666666",
"collectionName": "STXPLATES"
}
```

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,389 @@
...
+    },
+    {
+      "trait_type": "strength",
+      "display_type": "number",
+      "value": 99,
+    },
+  ],
+  "properties": {
+      "collection":  "Foo Collection",
+      "collection_size":  "10000"
```
dantrevino, line 158: The example here uses "collection_size", but below, under common properties, "total_supply" is specified. I would like some clarity here. My expectation is that we will use "total_supply" for collection size.

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,389 @@
...
+| `files`                         | `array`   | list of all associated files, represented as `{uri: string, type: string, signature: string, signature_type: string}`.                                                                                                                                                                                                                                              |
+| `creators`                      | `array`   | list of creators and their shares, represented as `{address: string, share: number}`.                                                                                                                                                                                                                                                                               |
```
dantrevino, line 389: How is "share" specified?  as a uint?  as a percentage?  Is the context for royalties?  Should royalties be speciifed separately?

sips/sip-x/sip-x-nft-metadata.md
```diff
@@ -0,0 +1,389 @@
...
+The JSON blob resolved through the uri must follow the following JSON schema.
+
+If metadata were retrieved by a function call containing a token identifier and the string `{id}` exists in any JSON value, it MUST be replaced with the actual token id in decimal format, by all client software that follows this standard.
+
+```
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Token Metadata",
+    "type": "object",
+    "required": ["version", "name"],
```
dantrevino, line 65: isnt "image" required?

friedger: I removed the requirement for image, because `properties.image_data` could be used or `image_data` directly. One of these three needs to be defined. I don't know how to specify that in the schema.

sips/sip-x/sip-x-nft-metadata.md
friedger: clarification has been added

sips/sip-x/sip-x-nft-metadata.md
friedger: fixed

friedger:
This SIP has been moved to accepted by the author. Please take the next steps @jcnelson

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+If metadata were retrieved by a function call containing a token identifier and the string `{id}` exists in any JSON value, it MUST be replaced with the actual token id in decimal format, by all client software that follows this standard.
+
+```
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Token Metadata",
+    "type": "object",
+    "required": ["version", "name"],
+    "properties": {
+        "version": {
```
friedger, line 69: Should this be `version` or `schema_version`? I couldn't find any general guidelines.

Here is a reference from databases to use `schema_version`: 
https://docs.mongodb.com/manual/tutorial/model-data-for-schema-versioning/

dantrevino: I prefer schema_version, just so that 'version' isnt confused with the NFT.  I have no good authoritative source to back up using 'schema_version' though.

friedger: Do you have a preference @Jamil @mardi00 @MarvinJanssen ? Everybody?

MarvinJanssen: I'm fine with either but prefer `version` because it is shorter. In any case, I will update SIP013 to be compliant.

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+    ]
+}
+```
+
+pt-BR.json
+
+```
+{
+    "version": 1,
+    "attritbutes: [
```
MarvinJanssen, line 194: Typo `attritbutes` -> `attributes`.

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+
+An array of localized `attributes` overwrites the whole list of default `attributes`.
+
+A localized properties with partial data overwrites only the provided properties, the remaining default properties remain as default values.
+
+# Using metadata in applications
+
+Before presenting metadata to users, application developers should verify whether the metadata is compliant with the application's guidelines.
+
+We remind implementation authors that the empty string for the token uri is a valid response. We also remind everyone that any smart contract can use the same metadata as other NFT contracts. It is out of the scope of this standard to define how a client may determine which smart contracts are is the original, well-known, canonical one.
```
MarvinJanssen, line 269: Typo `are is`.

MarvinJanssen (approved):
I've also been working on a generative NFT (from on-chain data) that will be implementing this standard. The JSON metadata files are generated by a server off-chain. I would like to get your opinion on it when the time comes.

sips/sip-16/sip-016-token-metadata.md
Jamil: It's not something I have a strong opinion on personally; agree with Marvin that brevity is generally preferable.

friedger: What about `sip` similar to `openapi` in https://swagger.io/specification/#openapi-object

sips/sip-16/sip-016-token-metadata.md
friedger: Thank you. Fixed

sips/sip-16/sip-016-token-metadata.md
friedger: I updated the spec to use `sip` instead of `version`

sips/sip-16/sip-016-token-metadata.md
friedger: Thank you, fixed.

friedger:
The spec now includes a property `sip` that is the number of the sip that specifies the json schema for the meta data. In this case `16`. Property `version` has been removed.

MarvinJanssen:
Not saying I am against the `sip` field in favour of `version` but I wonder if it may be too restrictive. (Still trying to think of scenarios.)

However, I like that it is nice and short. I'll bring it into the next draft for SIP013.

dantrevino:
> Not saying I am against the `sip` field in favour of `version` but I wonder if it may be too restrictive. (Still trying to think of scenarios.)
> 
> However, I like that it is nice and short. I'll bring it into the next draft for SIP013.

I think its more descriptive than 'version' ... the version really does refer to a sip, and the version would not change without a new sip ... for that reason, I think its actually more clear.

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+
+License: CC0-1.0
+
+Sign-off:
+
+Layer: Traits
+
+# Abstract
+
+Non-fungible tokens - NFTs for short - are digital assets registered on blockchain with unique identifiers and properties that distinguish them from each other. SIP-009 defines the trait for how ownership of an NFT is managed. Fungible tokens - FTs for short - are digital assets where each token can be replaced by another token (see SIP-010). Semi-fungible tokens are digital assets where each token has a unique identifier and is dividable into fungible parts (see SIP-013). This SIP aims to provide a flexible standard to attach metadata to NFTs, like descriptions or urls to digital files. The same standard is applicable to fungible tokens.
```
jcnelson, line 25: Small formatting nit -- can you make it so that lines are 80 columns?  It will make managing patches a lot less conflict-prone.  Thanks!

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+
+Appendix A contains a list of trait functions that must meet the following requirements for the return value. The appendix can be extended without changing the ratification status of this SIP.
+
+## Return Value of Metadata URI Functions
+
+The return value must be a `some` value if and only if the metadata reference an existing token, otherwise the value must be `none`. Appendix A specifies the exact meaning of "existing" for each function.
+
+For existing tokens, the inner value of the return value must be a string representing a resolvable URI.
+
+The schema of the resolvable URI is not specified and should be a well-known schema like `https`, `ar`, `ipfs`, `sia`. A `data` URI is also valid, however, the length is limited by this SIP.
```
jcnelson, line 50: Minor formatting nit -- can you use either `URI` or `uri` consistently?  Thanks!

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+
+```
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Token Metadata",
+    "type": "object",
+    "required": ["sip", "name"],
+    "properties": {
+        "sip": {
+            "type": number,
```
jcnelson, line 70: Typo: `"type": "number",`

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+            "type": "string",
+            "description": "A URI pointing to a resource with mime type image/* representing the asset to which this token represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive. If the token represents a media file of different mime-type or of higher quality defined in property 'raw_media_file_uri', then this image should be used as preview image like a cover for music, or an low-res image."
+        },
+        "attributes": {
+            "type": "array",
+            "description": "Arbitrary attributes. Values may be strings, numbers, object or arrays."
+            "items: {
+                "type": "object",
+                "required": ["value"],
+                "properties": {
```
jcnelson, line 91: If this field is meant to contain arbitrary attributes, then why are items in this array required to be objects that contain special type restrictions for `display_name` and `trait_name`?

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+{
+    "type": "string",
+    "description": "Address of custodian key holder",
+    "value": "Casa Inc., P.O. Box 20575, Charleston, S.C. 29413, United States."
+}
+```
+
+### Attributes
+
+Attributes describe addition elements of tokens that are "observable", usually represented in the image of the token.
```
jcnelson, line 247: Possible typo -- did you mean "additional"?

jcnelson: Also, can `attributes` and `properties` be code-formatted?

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+
+## Localization
+
+The localized data follow the same JSON schema with property `sip` as required and all other properties as optional.
+
+The localized data overwrite data provided in the default meta data JSON. The localized data can provide only partial data.
+
+An array of localized `attributes` overwrites the whole list of default `attributes`.
+
+A localized properties with partial data overwrites only the provided properties, the remaining default properties remain as default values.
```
jcnelson, line 263: Typo --> "...provided properties; the remaining..." (semicolon)

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+The [badges contract](https://explorer.stacks.co/txid/0xb874ddbb4a602c22bb5647c7a2f8bfcafbbca7c0c663a175f2270ef3665f33de?chain=mainnet) defines metadata for nfts.
+
+The function signatures for metadata are:
+
+- `(get-badge-meta () {uri: (string-ascii 78111)})` and
+- `(get-meta? (uint) (optional {user: principal}))`
+
+# Backwards Compatibility
+
+This SIP defines meta data so that metadata for existing NFTs on other blockchains like Ethereum, Solana or WAX can be re-used for NFTs on the Stacks blockchain.
```
jcnelson, line 318: "meta data" <-- did you mean "metadata"?

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+
+# Introduction
+
+Tokens are digital assets registered on blockchain through a smart contract. A non-fungible token (NFT) is a token that is globally unique and can be identified through its unique identifier. In blockchains with smart contracts, including the Stacks blockchain, developers and users can use smart contracts to register and interact with non-fungible tokens.
+
+Some use cases of NFTs are name registration, digital art, certification, media and entertainment, real-estate. They all require that users associate certain content with an NFT. In general, it is helpful for the users to have a name, sound, image that represents this content.
+
+# Specification
+
+Every SIP-X compliant smart contract in Stacks blockchain must implement one or more functions that return a resolvable/retrievable uri referencing metadata. The metadata provide information e.g. for displaying a digital asset to users. This type of function is named "metadata uri functions".
```
jcnelson, line 40: You can use `SIP-016` here instead of `SIP-X` :)

Also, typo -- "smart contract in **the** Stacks blockchain"

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+
+Appendix A contains a list of trait functions that must meet the following requirements for the return value. The appendix can be extended without changing the ratification status of this SIP.
```
jcnelson, line 42: That's fine, but can you also add a requirement that any changes will be explicitly noted in a changelog subsection for this SIP?

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+
+## Return Value of Metadata URI Functions
+
+The return value must be a `some` value if and only if the metadata reference an existing token, otherwise the value must be `none`. Appendix A specifies the exact meaning of "existing" for each function.
+
+For existing tokens, the inner value of the return value must be a string representing a resolvable URI.
+
+The schema of the resolvable URI is not specified and should be a well-known schema like `https`, `ar`, `ipfs`, `sia`. A `data` URI is also valid, however, the length is limited by this SIP.
+
+If a metadata uri function expects a parameter of type `uint` that identifies a token and the resulting strings contains `{id}`, then the `{id}` part must be replaced by the identifier in decimal format given in the function call.
```
jcnelson, line 52: Typo: "resulting strings contain" (not "contains").

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+            "type": "string",
+            "description": "Identifies the asset which this token represents"
+        },
+        "description": {
+            "type": "string",
+            "description": "Describes the asset which this token represents"
+        },
+        "image": {
+            "type": "string",
+            "description": "A URI pointing to a resource with mime type image/* representing the asset to which this token represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive. If the token represents a media file of different mime-type or of higher quality defined in property 'raw_media_file_uri', then this image should be used as preview image like a cover for music, or an low-res image."
```
jcnelson, line 83: Minor nit: should be "MIME type."

sips/sip-16/sip-016-token-metadata.md
jcnelson: Ah, I see this is answered below.  But nevertheless, maybe a comment on why these two fields exist would be appropriate for the `"description"` field, to minimize reader confusion on the first pass?

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+        },
+        "animation_url": {
+            "type": "string",
+            "description": "URL to a multi-media attachment for the item. Deprecated. Use `properties.animation_url`."
+        },
+    }
+}
+```
+
+The lengths of string values is not restricted. Nowadays, clients should be smart enough to deal with values of different lengths.
```
jcnelson, line 136: Is there a reasonable limit we can advise here?  Like, maybe no longer than 1024 characters is recommended?

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+
+### Attributes
+
+Attributes describe addition elements of tokens that are "observable", usually represented in the image of the token.
+
+In contrast, properties describe elements of tokens that are more abstract.
+
+An attribute consists of a `trait_type` defining the name of the trait, `value` is the value of the trait, and `display_type` is a field indicating how you would like it to be displayed.
+
+Appendix B describes type of attributes
```
jcnelson, line 253: After reading and re-reading this section, I still don't really know the difference between `attributes` and `properties`.  Can you provide a deeper explanation?  I also still don't really understand what `trait_type` and `display_type` do, and why they're necessary.

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+
+The localized data overwrite data provided in the default meta data JSON. The localized data can provide only partial data.
+
+An array of localized `attributes` overwrites the whole list of default `attributes`.
+
+A localized properties with partial data overwrites only the provided properties, the remaining default properties remain as default values.
+
+# Using metadata in applications
+
+Before presenting metadata to users, application developers should verify whether the metadata is compliant with the application's guidelines.
```
jcnelson, line 267: Can you speak more to what these application's guidelines are, and where they could be found?  Is there a real-world example you can reference?

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+
+We remind implementation authors that the empty string for the token uri is a valid response. We also remind everyone that any smart contract can use the same metadata as other NFT contracts. It is out of the scope of this standard to define how a client may determine which smart contracts are the original, well-known, canonical ones.
+
+## Graphical representation
+
+The metadata of a token contain several properties that can be used to visually represent the token. It is recommended to use the first defined property of the following list:
```
jcnelson, line 273: Given that some of these fields are marked as deprecated, it might make more sense to use this section simply to describe the order in which a set of fields that describe the same metadata should be considered.  Maybe this should be an appendix?

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+2. `animation_url`
+
+# Out of Scope
+
+Accessiblity of content is not covered by the standard.
+
+Properties other than resolvability of the token uri are out of scope. This implies that metadata might change over time (stability).
+
+# Metadata functions
+
```
jcnelson, line 291: Can this section speak more to the extent that metadata functions should be relied upon, now that this SIP exists?  For example, does this SIP supersede all metadata functions?  Does this SIP require that certain fields have the same values as those returned by corresponding metadata functions?  I'm not sure how to relate these two things.

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+| ------------------------------- | --------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `collection`                    | `string`  | collection name the token belongs to. See also Appendix A.                                                                                                                                                                                                                                                                                                          |
```
jcnelson, line 375: Can this table be sorted by `Name`?

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+| `artist_name`                   | `string`  | name of the artist, mainly used as attribution.                                                                                                                                                                                                                                                                                                                     |
+| `seed`                          | `string`  | hex string represented the DNA of the NFT. The seed is usually stored on-chain, it might be contained in the metadata for convenience.                                                                                                                                                                                                                              |
```
jcnelson, line 394: Is there a reference implementation of a metadata validator somewhere that this SIP can include as supplemental data?

jcnelson:
Hey @friedger @dantrevino, thanks for being patient -- I didn't see the original github notification from December.  I left my initial feedback pass.  Hope it's helpful!

sips/sip-16/sip-016-token-metadata.md
friedger: fixed

sips/sip-16/sip-016-token-metadata.md
friedger: I added more details.

sips/sip-16/sip-016-token-metadata.md
friedger: Not sure who to improve the section. I think it is important to define which properties should be used to display a token.

sips/sip-16/sip-016-token-metadata.md
friedger: I tried to improve this section.

sips/sip-16/sip-016-token-metadata.md
friedger: fixed

sips/sip-16/sip-016-token-metadata.md
friedger: re-ordered

sips/sip-16/sip-016-token-metadata.md
friedger: Not yet... as far as I know.

friedger (commented):
@jcnelson I have tried to address your comments. Thank you for your time to review the SIP!

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,476 @@
...
+
+Images of NFTs often have a limited set of traits and each trait has a limited
+number of possible values. These values are represented as attributes in the
+metadata. They can be used to calculate a score for each NFT in the collection
+that could define the rarity of the NFT.
+
+An `attribute` consists of a `trait_type` defining the name of the trait, e.g.
+"hair". The `value` is the value of the trait, e.g. "red". The `display_type` is
+a field indicating how the trait value should be displayed, e.g. on a
+marketplace.
```
jcnelson, line 296: If `trait_type` and `display_type` are required fields for items in the `attributes` array, then should this be reflected in the JSON schema?

friedger: fixed.

sips/sip-16/sip-016-token-metadata.md
jcnelson: Thanks! Looks good.

sips/sip-16/sip-016-token-metadata.md
jcnelson: I see that.  Thanks!

jcnelson:
Right now, the consideration advisory board for technical-track SIPs has no members besides myself [1].  That said, I'm happy to advance this to `Recommended` status on the board's behalf if there are no other objections or feedback.  Once at least one mainnet contract is using this trait, the SIP can be transitioned to `Activation-in-Progress`.  Feel free to add my email to the `Signed-off: ` section in the preamble.

[1] I want to change this -- the governance group is happy to take new members.  The list of all consideration advisory boards and their members can be found here: https://github.com/stacksgov/sips/tree/main/considerations

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,479 @@
...
+```
+{
+    "sip": 16,
+    "attributes: [
+        {
+          "trait_type": "cabelos",
+          "value": "vermelho",
+        },
+        {
+          "trait_type": "amido",
```
Dulb26, line 235: if this was supposed to be strength in brazilian portuguese the correct would be "força" 
amido is starch

friedger: Thank you. Fixed!

dcsan:
>  `seed`                          | `string`  | hex string represented the DNA of the NFT. The seed is usually stored on-chain, it might be contained in the metadata for convenience.

I would be tempted to make the seed a string, unless there's a real benefit in terms of gas or other on-chain storage.
Some people might use it to just store a text message, like a title or a couple of words.
As (hex) data the stacks API functions had problems automatically detecting types when returning as JSON, mixed with other fields, and also this required conversions before storing in a DB etc. Overall it makes things much more cumbersome. It reminds me of the days before unicode of all kinds of ugly encoding errors.

friedger:
> > `seed`                          
> I would be tempted to make the seed a plain/not hex string, unless there's a real benefit in terms of gas or other on-chain storage for using hex. Some people might use it to just store a text message, like a title or a couple of words. As (hex) data the stacks API functions had problems automatically detecting types when returning as JSON, mixed with other fields, and also this required conversions before storing in a DB etc. Overall it makes things much more cumbersome. It reminds me of the days before unicode of all kinds of ugly encoding errors.

My idea was to provide a well-defined machine processable field. If it is not useful, could you please provide a better description, maybe also a better name instead of `seed`? Maybe we want to define two different fields. @dcsan 

If the API has a problem then we should fix the API not the standard :-)

dcsan:
`seed` is fine as a name, i think its over-complex to have two fields for this too.
the API is fine when it's reading just one data type but when it's a mixed struct it fails. 
But it's unavoidable that this also means the client has to encode data specially for this seeding, which is just extra potential for error, and in fact occurred for me. What worked in the test suite didn't work when run through (hiros) API contract functions.

So, i'd suggest to keep it simple so we can see and confirm what's getting posted in, and what we get out - as strings, without adding a double-dutch sandwich of encoding/decoding for little purpose, that (currently) mangles the data.

friedger:
@jcnelson The schema has been validated using JsonValidator.  See https://github.com/Light-Labs/media-metadata-schemas
The found errors has been fixed.

jcnelson:
Just checking in.  Is there a mainnet contract using this trait yet?

jcnelson:
@friedger Feel free to advance the status of this SIP to `Activation-in-Progress`, per my earlier comment.  I can't commit directly to your PR since it's in your repo.

sips/sip-16/sip-016-token-metadata.md
```diff
@@ -0,0 +1,479 @@
...
+                "required": ["trait_type", "value"],
+                "properties": {
+                    "display_type": {"type": "string"},
+                    "trait_type": {"type": "string"},
+                    "value": {"anyOf": [{"type": "object"}, {"type": "string"}, {"type": "number"}, {"type": "integer"}, {"type": "boolean"}, {"type": "array"}]}
+                }
+            }
+        },
+        "properties": {
+            "type": "object",
```
rafaelcr, line 131: @friedger shouldn't this be an `array` of `object`s (following the schema specified in **Properties**)? Similar to `attributes` above

Hero-Gamer:
Hi all just wanna make sure this is captured officially. 

During last Friday's weekly SIP call https://github.com/stacksgov/sips/issues/79#issuecomment-1212993156 I raised this point. 

According to the activation criteria:
_"This SIP is activated if 10 contracts are deployed that follows this
specification. This must happen before Bitcoin tip **#750,000**."_

We are now at bitcoin block 749,565.
**435 blocks left** as of writing, which is approx. **3 days.** 

So I'm now working with the NFT marketplaces to fetch data to confirm 10 contracts have been deployed already.
Will update all how it goes.

dantrevino:
There's a miss here @friedger ... this is not a contract sip, its a metadata sip.  @Hero-Gamer there is no way to look at a contract and determine if the metadata supports sip-016.  The contract only has a link to metadata, but, for instance, the boom-300 "contract" was deployed 10 months ago, before the sip was complete, and initially we created nfts that were not sip-016 compliant.  However, now it is deploying nfts that *are* sip-016 compliant.  Same contract.

https://explorer.stacks.co/txid/SP1QK1AZ24R132C0D84EEQ8Y2JDHARDR58R72E1ZW.boom-nft-300?chain=mainnet

both minted from the same contract:
non-sip016 nft, created jan 17: https://bafkreie3abg3pp4573zkjmxcgaosifpyof5hi5gxelnqutnyhelihjqeqq.ipfs.nftstorage.link/
sip016 nft, created recently: https://bafkreicg63zfryk2cvrucztb5bztg4vh6hj2e65yxcr7nld2rtryobigcy.ipfs.nftstorage.link/

Jamil:
Our Create Portal contracts are almost SIP-16 compliant but not quite, since we don't have the required "sip" field in the JSON. Working with our developers to add that for all future contracts, but not sure 10 contracts like this will be deployed by Wednesday. What is the process if this SIP does not meet the requirements by then, and we have to re-start?

dartman10:
I found these with `"sip": 16` in metadata:

name | contract | metadata_url
-- | -- | --
Boombox [10th Edition] | SP1QK1AZ24R132C0D84EEQ8Y2JDHARDR58R72E1ZW.boomboxes-cycle-32 | https://cloudflare-ipfs.com/ipfs/bafkreigzei42ktazjl3b6ounitwazbmcz7zzn5vb7poevx2xoroau7ftnm
Boombox [12th Edition] - Bloombox by Grace Hye | SP1QK1AZ24R132C0D84EEQ8Y2JDHARDR58R72E1ZW.boomboxes-cycle-36 | ipfs://bafkreidowklqh7nub2dxrh2mj4tu46inqtapqfidqnwk5taown5w4li6pq
Boombox [13th Edition] | SP1QK1AZ24R132C0D84EEQ8Y2JDHARDR58R72E1ZW.boomboxes-cycle-38 | ipfs://bafkreibduyxkdcp3djswzjjoexoi6hcvk2suaoyri2uqpiuyxbhkh6bwxe
Boombox [14th Edition] - STX of Toys | SPMS4E9RQ4GCGG68R6D15PKV01TYNCBPYZG1ZMFE.boomboxes-cycle-40 | ipfs://bafkreieanobksr6qetgqupx7zpn2aebw742siwsqa4ef6i7aybxvnku2ze
Boombox [8th Edition] - Moonbox | SP1QK1AZ24R132C0D84EEQ8Y2JDHARDR58R72E1ZW.boomboxes-cycle-28 | https://cloudflare-ipfs.com/ipfs/bafkreic3emy7vy4azvm56sy3dwsuny4653uzi7ubb36ui63w424kagzfym
Boombox [8th Edition] - Moonbox | SP1QK1AZ24R132C0D84EEQ8Y2JDHARDR58R72E1ZW.boomboxes-cycle-28-moonbox | https://cloudflare-ipfs.com/ipfs/bafkreic3emy7vy4azvm56sy3dwsuny4653uzi7ubb36ui63w424kagzfym
Boombox [9th Edition] - Mushroom by Flower | SP1QK1AZ24R132C0D84EEQ8Y2JDHARDR58R72E1ZW.boomboxes-cycle-30 | https://cloudflare-ipfs.com/ipfs/bafkreibszx62vqkce7pefr6xtcqvyec7lo5xdjy6c7r3dhfd5bkl3wyvku

jcnelson:
Technically yes, you would need to restart in general since it indicates that the SIP doesn't have enough interest.  But it seems that in this specific case, the only reason this SIP hasn't activated was due to a bug in the implementation (e.g. a missing `sip` field in the JSON), but there is otherwise sufficient interest.  In this case, it would be fine by me to just combine the whole "withdraw, re-submit, accept, recommend, activation-in-progress" sequence into simply the act of increasing the activation-in-progress block height to whatever you need (a week?).  The effect is the same either way; the SIP text wouldn't be changing and the activation-in-progress section only needs the steering committee's sign-off (which right now is just me).

friedger:
Here is another one: https://explorer.stacks.co/txid/SP1QK1AZ24R132C0D84EEQ8Y2JDHARDR58R72E1ZW.boom-nft-300?chain=mainnet Makes it 8.

dartman10:
No. 9 - a new one deployed a few hours ago:
SPQE8N8BHMT462W2XPK028GDM4RMQBSHAAY8D37G.crashpunks-punkettes
https://explorer.stacks.co/txid/0x95c03cb957324a9983430b4419c4d4180d13ca5b067cda7e23d314df4f762443?chain=mainnet
ipfs://QmcUiKLAeYbx78UZWoCuyKdm5GhZH6Jhto9yhKt1oPvWHY/

Hero-Gamer:
153 bitcoin block time left, approx. 1 day. (current block time 749,847)
Who else/where else can we ask to find the last contract required?
almost there!

Jamil:
The contract above (`SPQE8N8BHMT462W2XPK028GDM4RMQBSHAAY8D37G.crashpunks-punkettes`) was deployed with the Gamma Create Portal, which now has the `sip` field in the JSON. 

The next contract deployed with the Gamma Portal will be the 10th :)

dartman10:
No. 10?  This NFT has a dynamic metadata.  The latest mint has a SIP-016 metadata, prior mints were not.

- United States of Bitcoin
- SP1JCPNPAMAQJ364AFHPTW3HY7X0HYZ3TJ0ZDGWZH.artificial-satoshi

Latest mint:
- ipfs://QmcNwUgnJJ4W5SEpuZ3f4MVjky69RuoVm8rG3DRXvtLvzD/json/1.json
- https://explorer.stacks.co/txid/0xd64cfe3e53b0d1291596b2bce6c613a168e8a9fd20d4fec7a4d6a6a9332abc61?chain=mainnet

Earlier mint example:
- ipfs://QmYi2633t2QRiWUwA5RpQ8QPEqbh8yAb5jSeLzd6D8WFZC/json/1.json
- https://explorer.stacks.co/txid/0x819fae14fa1b0633e75ade07002f6df5017805615168ba7f9f89d38d25f05dec?chain=mainnet

Hero-Gamer:
If all contracts checks out:
Current Bitcoin block height: **749,876**
https://mempool.space/block/0000000000000000000894d96480628dcd38fd9424750e9d2638aa75233e19f0

jcnelson:
Awesome to see!  I will take a look as soon as I am able, and if it all checks out, it will be my pleasure and privilege to advance this to Ratified :tada:

jcnelson:
> Here is another one: https://explorer.stacks.co/txid/SP1QK1AZ24R132C0D84EEQ8Y2JDHARDR58R72E1ZW.boom-nft-300?chain=mainnet Makes it 8.

@friedger I'm not sure where to look for the metadata URL here?  This appears to be a contract deploy from 6 months ago.

dartman10:
Perhaps Xan Ditkoff can help. :)



name | contract | metadata_url
-- | -- | --
Xan Ditkoff #1 | SP2VG7S0R4Z8PYNYCAQ04HCBX1MH75VT11VXCWQ6G.the-built-on-bitcoin-podcast | ipfs://QmQBr6yGuiJmsWTmkYSbgkU1GX41BJpNB19mKNiLeFh3Lq/json/1.json
Xan Ditkoff #2 | SP2VG7S0R4Z8PYNYCAQ04HCBX1MH75VT11VXCWQ6G.the-built-on-bitcoin-podcast | ipfs://QmQBr6yGuiJmsWTmkYSbgkU1GX41BJpNB19mKNiLeFh3Lq/json/2.json
Xan Ditkoff #3 | SP2VG7S0R4Z8PYNYCAQ04HCBX1MH75VT11VXCWQ6G.the-built-on-bitcoin-podcast | ipfs://QmQBr6yGuiJmsWTmkYSbgkU1GX41BJpNB19mKNiLeFh3Lq/json/3.json
Xan Ditkoff #4 | SP2VG7S0R4Z8PYNYCAQ04HCBX1MH75VT11VXCWQ6G.the-built-on-bitcoin-podcast | ipfs://QmQBr6yGuiJmsWTmkYSbgkU1GX41BJpNB19mKNiLeFh3Lq/json/4.json
Xan Ditkoff #5 | SP2VG7S0R4Z8PYNYCAQ04HCBX1MH75VT11VXCWQ6G.the-built-on-bitcoin-podcast | ipfs://QmQBr6yGuiJmsWTmkYSbgkU1GX41BJpNB19mKNiLeFh3Lq/json/5.json
blobby | SP1CSHTKVHMMQJ7PRQRFYW6SB4QAW6SR3XY2F81PA.continuous-007 | ipfs://QmRD1wKK6cMmfrgPBAzSR9LYj7ifKauETCEARJu1EpMZ9C/json/2.json

jcnelson:
@dartman10 thanks!  Looks good to me.

It is my pleasure to inform everyone that this SIP has met all of the criteria for ratification!  Please transition this SIP to "Ratified" and I'll merge it :tada:

friedger:
Great work everybody!

@jcnelson I changed the status to Ratified. Anything else to do?

jcnelson:
Nope; going to merge

PR #50, opened 2021-11-19
SIP for Authentication Protocol

friedger:
This SIP defines a authentication protocol used by Stacks apps.

The current version has (hopefully) all the required information about the protocol as it is currently used. 

I changed three properties of the auth response: `hubUrl` -> `hub_url` and `associationToken` -> `association_token`. `profile.stxAddress`-> `stx_address`.

I added `state` to the auth messages as defined in OAuth 2.0.

It is recommended to use  `did:stacks:v2` instead of `did:btc-addr`

For the public profile, this spec uses the Verifiable Credential model. The [VC spec](https://www.w3.org/TR/vc-data-model) was chosen because it now has W3C Recommendation status.

### DIFF
--- a/sips/sip-x/sip-x-authentication-protocol.md
+++ b/sips/sip-x/sip-x-authentication-protocol.md
@@ -31,9 +31,9 @@ This SIP’s copyright is held by the Stacks Open Internet Foundation.
 
 Decentralized application do not want to store credentials of their users. Instead users should be able to login using some kind of cryptographic proof that they control a public key.
 
-The private key for that public key is guarded by a so-called authenticator. When a users visits the app, the app needs to communicate with the authenticator. In the authenticator, the user can control which public key should be shared with the application. 
+The private key for that public key is guarded and managed by a so-called authenticator. When a users visits the app, the app needs to communicate with the authenticator. In the authenticator, the user can control which public key should be shared with the application. 
 
-In addition, to the public key more information can be shared like email address or profile pictures. In particular, a private key is derived by the authenticator specific for the application and for the user. This private key can be used by the application to access for example decentralized storage or sign messages in the name of the user of the application.
+In addition to the public key, more information can be shared like email address or profile pictures. In particular, a private key is derived by the authenticator specific for the application and for the user. This private key can be used by the application to access for example decentralized storage or sign messages in the name of the user of the application.
 
 # Specification
 
--- a/sips/sip-x/sip-x-authentication-protocol.md
+++ b/sips/sip-x/sip-x-authentication-protocol.md
@@ -61,24 +61,38 @@ The payloads for the authentication request and the response are specified in th
 
 ## Authentication Request
 
-The authentication request is a JWT created by the application. It is signed by a private key that is usually freshly generated. The key is called app transit key. The payload schema is defined as follows:
+The authentication request is a JWT created by the application. It is signed by a private key that must be freshly generated. The key is called app transit key.
+
+### Signed JWT
 
 The payload must contain the following claims:
 
-| Claim name             | Type   | Description                                                                                                   |
-| ---------------------- | ------ | ------------------------------------------------------------------------------------------------------------- |
-| jti                    | string | As defined in RFC7519                                                                                         |
-| iat                    | string | As defined in RFC7519                                                                                         |
-| exp                    | string | As defined in RFC7519                                                                                         |
-| iss                    | string | Decentralized identifier defined in DID specification representing the user's account.                        |
-| public_keys            | array  | Single item list with the public key of the signer                                                            |
-| domain_name            | string | The url of the application with schema.                                                                       |
-| manifest_uri           | string | The url of the application manifest, usually domain_name + "/manifest.json"                                   |
-| redirect_uri           | string | The url that should receive the authentication response                                                       |
-| do_not_include_profile | bool   |                                                                                                               |
-| supports_hub_url       | bool   |                                                                                                               |
-| scopes                 | array  | list of strings specifying the requested access to the user's account. See Appendix A for full list of scopes |
-| version                | string | must be "2.0.0"                                                                                               |
+| Claim name             | Type   | Description                                                                                                                               |
+| ---------------------- | ------ | ----------------------------------------------------------------------------------------------------------------------------------------- |
+| jti                    | string | As defined in RFC7519                                                                                                                     |
+| iat                    | string | As defined in RFC7519                                                                                                                     |
+| exp                    | string | As defined in RFC7519                                                                                                                     |
+| iss                    | string | Decentralized identifier defined in DID specification representing the user's account. See Appendix C for list of well-known DID methods. |
+| public_keys            | array  | Single item list with the public key of the signer                                                                                        |
+| domain_name            | string | The url of the application with schema.                                                                                                   |
+| manifest_uri           | string | The url of the application manifest, usually domain_name + "/manifest.json"                                                               |
+| redirect_uri           | string | The url that should receive the authentication response                                                                                   |
+| do_not_include_profile | bool   |                                                                                                                                           |
+| supports_hub_url       | bool   |                                                                                                                                           |
+| scopes                 | array  | list of strings specifying the requested access to the user's account. See Appendix A for full list of scopes                             |
+| version                | string | must be "2.0.0"                                                                                                                           |
+
+### Verification
+
+Authenticators should verify that the request has the following properties:
+
+- expiration date (`exp`) is not in the past
+- issuance date (`iat`) is in the past
+- public keys length is 1
+- public key (`public_keys[0]`) is same as the signer's key
+- public key's stacks address is the same as the issuer
+- manifest url is same origin as the app domain
+- redirect url is same origin as the app domain
 
 ## User authorization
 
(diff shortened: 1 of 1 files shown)


sips/sip-x/sip-x-authentication-protocol.md
```diff
@@ -111,10 +151,9 @@ https://github.com/PravicaInc/wise-js
 
 ## Authenticators
 
-
 ### Hiro Wallet
 
-Hiro Wallet is an wallet that handles authentication and transaction signing.
+Hiro Wallet is an browser extension that handles authentication and transaction signing.
```
314159265359879, line 156: Hiro Wallet is **a** browser extension...

friedger: @314159265359879 Thank you fore reading the sip and finding these. Fixed

friedger:
After the description of the current protocol 1.3.1, I have updated the spec to 2.0.0 in https://github.com/stacksgov/sips/pull/50/commits/7b35de622fceae1d7d31634985cfcc5ee5115f14 using verifiable credentials and better definition of the issuers.

sips/sip-x/sip-x-authentication-protocol.md
```diff
@@ -0,0 +1,397 @@
...
+In addition to the public key, more information can be shared like email address or profile pictures. Some data can be shared publicly, other only with the application. In particular, a private key is derived by the authenticator that is specific to the application and to the user. This private key can be used by the application for example to access decentralized storage detailed in the response or sign messages in the name of the user of the application.
+
+# Specification
+
+The basic flow of the authentication between the application and the authenticator (aka wallet or agent of the user) is as follows:
+
+1. Application creates app transit private key, signs an auth request with that key and sends the request to the Authenticator.
+2. In the Authenticator, User authorizes sharing of public key, Authenticator derives app private key from request and updates the user's public profile if required.
+3. Authenticator creates response with authorized data and sends response to the Application.
+4. Application verifies signature against the app transit private key.
```
aulneau, line 45:
```suggestion
**Client (app):**
1. Client generates ephemeral private/public key pair: `transitPrivateKey`
2. Client creates an auth request payload (`authRequestPayload`) with details about the application, eg:
    ```ts
    export interface AuthRequestPayload {
      scopes: AuthScope[];
      redirect_uri: string;
      public_keys: string[];
      domain_name: string;
      appDetails: AppDetails;
    }
    ```
3. Client signs the auth request payload (a JWT) with the `transitPrivateKey`, with signature format `ES256k`
4. Client sends the JWT to an authenticator (eg: Hiro Web Wallet)

**Authenticator (wallet):**
1. Authenticator validates the signature against the included public key(s)
2. Authenticator generates an account index by creating a SHA256 hash of the domain of the application that generated the `authRequestPayload`, and the wallet's salt
3. Authenticator then uses that index to derive a hardened child private key (`appPrivateKey`) from the root seed phrase (BIP32).
4. Authenticator generates an `authResponse` which includes the `appPrivateKey`
5. Authenticator encrypts the JWT with the public key provided in the `authRequestPayload`

**Client (app):**
1. Client receives the encrypted `authResponse` and decrypts it with the `transitPrivateKey`
2. Client is now authenticated
```

sips/sip-x/sip-x-authentication-protocol.md
```diff
@@ -0,0 +1,397 @@
...
+### Signed JWT
+
+The payload must contain the following claims:
+
+| Claim name             | Type   | Description                                                                                                                               |
+| ---------------------- | ------ | ----------------------------------------------------------------------------------------------------------------------------------------- |
+| jti                    | string | As defined in RFC7519                                                                                                                     |
+| iat                    | string | As defined in RFC7519                                                                                                                     |
+| exp                    | string | As defined in RFC7519                                                                                                                     |
+| iss                    | string | Decentralized identifier defined in DID specification representing the user's account. See Appendix C for list of well-known DID methods. |
```
aulneau, line 75: This seems to be a combination of both an authResponsePayload and an authRequestPayload. How can the authRequest include the DID of an address its not yet aware of?

sips/sip-x/sip-x-authentication-protocol.md
```diff
@@ -0,0 +1,397 @@
...
+- issuance date (`iat`) is in the past
+- public keys length is 1
+- public key (`public_keys[0]`) is same as the signer's key
+- public key's stacks address is the same as the issuer
+- manifest url is same origin as the app domain
+- redirect url is same origin as the app domain
+
+## User authorization
+
+The authenticator manages the user' private keys. The protocol requires that keys are created from a deterministic wallet using [BIP-39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki).
```
aulneau, line 100: should be bip32?

sips/sip-x/sip-x-authentication-protocol.md
```diff
@@ -0,0 +1,397 @@
...
+- Update public profile (depending on requested scopes)
+- Create authentication response
+
+## Application Private Key
+
+If the authentication request contains the scope `storage_write`, then the authenticator must create a private key that is specific for the requesting application. It must use the following algorithm:
+
+1. Wallet salt: create the sha256 hash of the hex representation of the public key of derivation path (`m/888'/0'`)
+1. Create sha256 hash of concatinated string of the `domain_name` from the request and the hex representation of the wallet salt.
+1. Hash code: Create a hash code (as defined in Java and Javascript) from the hex representation of the hash, then apply bitwise `and` to the hash code and 0x7fffffff.
```
aulneau, line 115: This is highly prone to collisions, and generally a bad implementation. Unfortunately, it has been used for years. I have created what I think is a better alternative [here in micro-stacks](https://github.com/fungible-systems/micro-stacks/blob/8d9e2493e0ed751bb64a7a7cd574b692c068f8d7/src/wallet-sdk/account/get-app-private-key.ts#L22-L37).

```ts
  // Rather than using the `hashCode` function to derive an index for the child
  // which makes it so there's a unreasonable high probability that someone can
  // log into a new app (website) and it accidentally shares data with another one
  //
  // we want to use the `appDomain` + `salt` to generate a sha256 hash that is
  // then used as the chaincode for the new app keychain
  const chainCode = hashSha256(utf8ToBytes(`${appDomain}${account.salt}`));
  const rootNode = HDKeychain.fromBase58(account.appsKey);
  const appKeychain = await HDKeychain.fromPrivateKey(rootNode.privateKey, chainCode).deriveChild(
    0,
    true
  );
  ```

With the formalization of this standard in a SIP, would it make sense to explore this (or another) alternative?

Comment (aulneau) [reply sips/sip-x/sip-x-authentication-protocol.md:115]: cc @zone117x @kantai for any thoughts you might have on this

**Review by aulneau — COMMENTED** [review]
@friedger, thank you so much for doing this. I added a few comments, curious what you think :)

File: sips/sip-x/sip-x-authentication-protocol.md (hunk 4 above)
Comment (friedger) [reply sips/sip-x/sip-x-authentication-protocol.md:100]: BIP32 would be sufficient. However, I would like to enforce BIP39 for usability for humans and portability?

File: sips/sip-x/sip-x-authentication-protocol.md (hunk 3 above)
Comment (friedger) [reply sips/sip-x/sip-x-authentication-protocol.md:75]: It is a did representation of the public transit key. I have updated to description.

File: sips/sip-x/sip-x-authentication-protocol.md (hunk 5 above)
Comment (friedger) [reply sips/sip-x/sip-x-authentication-protocol.md:115]: Interesting. We should fix that!

File: sips/sip-x/sip-x-authentication-protocol.md [hunk 6]
```diff
@@ -0,0 +1,397 @@
...
+| public_keys       | array   | Single item array containing the public key of the selected account in hex representation.                                                                                                                          |
+| profile           | object  | Object containing properties of the users, the object schema should be well-known type of Appendix B. This can be the public profile of the selected account.                                                       |
+| apps_meta         | object  | Information about user data of different apps. Property names are the domain name of the app. Each property value is an object of containing properties for `storage` and `publicKey`. See "Application Meta Data". |
+| username          | string  | BNS username, owned by the first public key of `public_keys` claim. Can be the empty string                                                                                                                         |
```
janniks, line 283: We're currently in the process of removing the `username` from the authentication flow completely. https://github.com/hirosystems/stacks.js/pull/1230 Keeping the username adds a bunch of complications and should be verified by the application anyway, so it makes sense for the application to fetch the username in the first place.

I bumped the token version to 1.4.0 for this, in case anybody is validating very strictly.

_some discussions/comments here:_
- https://github.com/hirosystems/stacks.js/pull/1230
- https://github.com/hirosystems/stacks-wallet-web/pull/2333

zone117x: +1 this seems like a good separation of concerns

aulneau:
I'm in the process of building a new library that implements this SIP in micro-stacks and I have a few questions:

In the auth request area:

`manifest_uri`: is this required anymore? I don't think we need it. Can you describe use cases for this?
`redirect_uri`: this is also not used anymore in the context of extension/native based wallets. this seems to be a hold over from early blockstack days.

In the auth response section:

`core_token`: is this needed? I don't think anything uses this, nor generates a token for it.
`email` wouldn't this be better in some profile? nothing uses this to date

friedger:
Replaced by https://github.com/stacksgov/sips/pull/166

PR #53, opened 2021-12-02
SIP-015: Upgrading Proof-of-Transfer Consensus

kantai:
This SIP draft proposes a set of upgrades to Stacking and PoX -- most of the current draft contents cover a proposed `pox-2` contract, and how it interacts with Stacking in the `stacks-node` and Clarity VM. However, this SIP also includes a proposal to address blockstack/stacks-blockchain#1805 -- some of the PR description text was copied into the draft as placeholder text; clearly that portion of the SIP needs to be more fleshed out. Because this SIP is incorporating that proposal, the draft also lists @jcnelson as an author (who opened the draft PR for addressing that issue).

### DIFF
--- a/sips/sip-013/sip-013-pox-upgrade.md
+++ b/sips/sip-013/sip-013-pox-upgrade.md
@@ -1,0 +1,171 @@
+## New methods in PoX-2
+
+### 1. `stack-extend`
+
+This method allows direct stackers to re-lock their funds for an additional
+12 cycles *before* their current lock has expired. The target use case for
+this is to allow users to repeatedly stack without cooldowns.
+
+This method checks that the stacker is still allowed to stack, that
+the stacker has not delegated to an operator, and that their funds are
+currently stacked. The caller may supply a new reward address for the
+extension.
+
+The special case handler for the PoX contract in the Clarity VM will
+check this method's return value and set the stacker's STX account to
+"auto-unlock" at the end of the last extended-to reward cycle.
+
+### 2. `delegate-stack-extend`
+
+This method allows operators to re-lock the funds of one of their
+delegation participants for additional cycles *before* that participants
+lock has expired. The target use case for this is to allow delegation
+operators to repeatedly stack on behalf of their users without cooldowns.
+
+This method checks that the delegator is still authorized on behalf of
+the given stacker (and remains authorized until the end of the
+locked-for period) and that the funds are currently stacked.
+
+Note that, just as with the existing `delegate-stack-stx` function,
+this method locks *but does not commit* the user's STX. The delegation
+operator must invoke `stack-aggregation-commit` to set a reward address
+for the locked funds.
+
+The special case handler for the PoX contract in the Clarity VM will
+check this method's return value and set the stacker's STX account to
+"auto-unlock" at the end of the last extended-to reward cycle.
+
+### 3. `stack-increase`
+
+This method allows direct stackers to lock additional funds for
+the remaining cycles on their current lock.
+
+This method checks that the stacker is still allowed to stack, that
+the stacker has not delegated to an operator, that their funds are
+currently stacked, and that they have enough unlocked funds to cover
+the increase. The caller may *not* supply a new reward address for the
+increase.
+
+The special case handler for the PoX contract in the Clarity VM will
+check this method's return value and set the locked amount in the
+stacker's STX account to correspond to the increased amount.
+
+### 4. `delegate-stack-increase`
+
+This method allows operators to lock additional funds for one of their
+delegation participants for the remaining cycles on that user's
+current lock.
+
+This method checks that the delegator is still authorized on behalf of
+the given stacker (and remains authorized until the end of the
+locked-for period), that the increased amount remains less than the
+delegation's `amount-ustx` field, that the user is currently locked,
+and that the user has enough unlocked funds to cover the increase.
+
+Note that, just as with the existing `delegate-stack-stx` function,
+this method locks *but does not commit* the user's STX. The delegation
+operator must invoke `stack-aggregation-commit` to set a reward address
+for the newly locked funds.
+
+The special case handler for the PoX contract in the Clarity VM will
+check this method's return value and set the locked amount in the
+stacker's STX account to correspond to the increased amount.
+
+### 5. `stack-unlock`
+
+This method allows direct stackers (not delegation participants) to
+unlock their funds *after* the active reward cycle: removing the
+account from all future reward sets.
+
+The special case handler for the PoX contract in the Clarity VM will
+check this method's return value and set the stacker's STX account to
+"auto-unlock" at the start of the next reward cycle (i.e., the user
+account will behave as if the current cycle is their last active
+cycle).
+
+### 6. `stack-unlock-early`
+
+This method allows direct stackers (not delegation participants) to unlock
+their funds during the *current* reward cycle if, and only if, the user's
+stacked amount was not sufficient to obtain a rewards slot.
+
+This method will behave similarly to `stack-unlock`, removing the account
+from all future reward sets. The special case handler for the PoX contract
+in the Clarity VM will check this method's return value and set the stacker's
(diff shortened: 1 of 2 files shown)


sips/sip-013/sip-013-pox-upgrade.md
```diff
@@ -0,0 +1,373 @@
...
+
+1. After unlocking, Stackers **must** wait one cycle before Stacking
+   again. This is an implementation consequence, not a requirement of
+   the algorithm itself.
+2. If users fail to qualify during a cycle, their funds remain locked
+   for the cycle regardless.
+3. PoX cannot support advanced behaviors that could be used for
+   different kinds of smart contracts: changing the PoX reward
+   address, increasing or decreasing lock amounts, etc. These features
+   would enable new kinds of applications that build on top of PoX.
```
jcnelson, line 50: I think there's a fourth shortcoming: nodes cannot determine which Stacks block is the PoX anchor block in reward cycle N without processing all Stacks blocks in all reward cycles up to N - 1.  This not only makes the initial chain sync tedious and error-prone, but also makes the chain vulnerable to deep reorgs by hiding PoX anchor blocks and precludes us from building light clients.  This fix is addressed by my PR here: https://github.com/blockstack/stacks-blockchain/pull/2707.

jcnelson: Ah, nm -- I see you discuss that later on.

sips/sip-013/sip-013-pox-upgrade.md
```diff
@@ -0,0 +1,373 @@
...
+Note that, just as with the existing `delegate-stack-stx` function,
+this method locks *but does not commit* the user's STX. The delegation
+operator must invoke `stack-aggregation-commit` to set a reward address
+for the newly locked funds.
+
+The special case handler for the PoX contract in the Clarity VM will
+check this method's return value and set the locked amount in the
+stacker's STX account to correspond to the increased amount.
+
+### 5. `stack-unlock`
```
jcnelson, line 193: Does this method exist yet?  Also, I'm not sure this is a desirable feature (at least, not without a steep early-withdrawal penalty).  Part of the appeal of PoX's current behavior is that it compels people to hodl.  Also, the number of reward cycles the stacker locks for is a strong signal to miners and the ecosystem that this stacker has full confidence in PoX for the lock-up duration.  If we take that away, then it means that everyone is potentially just locking for the next reward cycle (since they can just exit afterwards), and there's not really a point to having a lock-up duration.

jcnelson: I think this would be fine if you unlocked your tokens in-between when you locked them and when your first payout reward cycle begins, however (is that what this is for?).

sips/sip-013/sip-013-pox-upgrade.md
```diff
@@ -0,0 +1,373 @@
...
+
+This method allows a user who delegated their PoX functionality
+to an operator to unlock a locked _but not committed_ amount from their
+account. The reason that the amount cannot have been committed yet is that
+the operator and the PoX contract cannot distinguish between which accounts
+have been used during a given cycle's commitment-- and allowing users to
+"uncommit" from the operator's pool would introduce denial-of-service vectors
+to the pool.
+
+### 8. `stack-aggregation-rollback`
```
jcnelson, line 226: Under what circumstances would an operator use this method?

sips/sip-013/sip-013-pox-upgrade.md
jcnelson: Looking at the example below, if the purpose of this method is to facilitate changing the PoX reward address, then why not just have a method to do that with the already-locked tokens?

sips/sip-013/sip-013-pox-upgrade.md
```diff
@@ -0,0 +1,373 @@
...
+blocks that passes through the canonical history of anchor blocks
+which resides on the canonical Bitcoin chain.
+
+# The PoX-2 Contract
+
+This section of the SIP specifies the new and changed behaviors of the
+PoX-2 contract and also provides new use cases that demonstrate how
+the new PoX-2 contract could be used.
+
+## New methods in PoX-2
```
jcnelson, line 120: Can this section include the signatures and return values for each new method?

314159265359879:
the SIP-013 name is also being used here: https://github.com/stacksgov/sips/pull/42 how do we prevent these conflicts?

jcnelson:
Yeah -- let's call this SIP 015.

friedger:
@jcnelson I thought draft sips never have a number until the draft is accepted according to sip-000

sips/sip-013/sip-013-pox-upgrade.md
kantai: If they wanted to increase the amount that they committed, but the increase by itself wouldn't be enough to get another entry in the rewards table.

sips/sip-013/sip-013-pox-upgrade.md
kantai: > Does this method exist yet?

Some methods of this SIP have been implemented in `next`, most haven't yet. But, I think as far as getting feedback on the SIP goes, we should ignore what has or hasn't been implemented: I think feedback on the SIP should inform what and how things get implemented, rather than the inverse.

> Part of the appeal of PoX's current behavior is that it compels people to hodl. Also, the number of reward cycles the stacker locks for is a strong signal to miners and the ecosystem that this stacker has full confidence in PoX for the lock-up duration. If we take that away, then it means that everyone is potentially just locking for the next reward cycle (since they can just exit afterwards), and there's not really a point to having a lock-up duration.

This is a very useful feature in and of itself -- obviously, it would break incentives to allow users to unlock during a reward cycle, but before one that hasn't occurred yet shouldn't alter incentives. Rather, _without_ this feature, PoX would encourage users that wish to have the freedom to unlock to only stack for one cycle at a time, continuously invoking `stack-extend` to get the same support, but with much more overhead (and stress about timing transactions before prepare phases).

jcnelson:
@friedger You are correct; I'm working under the assumption that SIP 011, SIP 012, SIP 013, and SIP 014 will eventually be accepted, but none of them officially have SIP numbers yet.  Also, SIP 000 does not say anything about the order in which numbers are assigned; it only requires that they are unique.

sips/sip-013/sip-013-pox-upgrade.md
jcnelson: > Rather, without this feature, PoX would encourage users that wish to have the freedom to unlock to only stack for one cycle at a time, continuously invoking stack-extend to get the same support, but with much more overhead (and stress about timing transactions before prepare phases).

I think that's desirable, actually.  If you want the option to stack for a shorter amount of time, and want the freedom to decide whether or not to extend in each cycle, then the system *should* require you to send a transaction in each cycle to extend your lock.  Put another way, the system should offer you a better deal for locking up for longer, since you locking up your tokens provides a service to the ecosystem (i.e. you signal confidence to miners that the network will be stable during your lockup).  In general, we *want* people locking up their STX for as long as possible, since it makes liquid STX more valuable, which in turn makes it easier for miners to sell their newly-minted STX at a profit.

FriendsFerdinand:
Is this the SIP for the Stacks 2.1 update? I noticed that one of the functions in the proposal (`assemble-principal`), is already in the stacks-blockchain repo as a PR (https://github.com/stacks-network/stacks-blockchain/pull/2756).

kantai:
> Is this the SIP for the Stacks 2.1 update? I noticed that one of the functions in the proposal (`assemble-principal`), is already in the stacks-blockchain repo as a PR ([stacks-network/stacks-blockchain#2756](https://github.com/stacks-network/stacks-blockchain/pull/2756)).

Yep -- this SIP is for 2.1 updates. The PRs in the stacks-blockchain repo should all be open against the `next` branch, which is where any 2.1 changes will be staged.

jcnelson:
Closing in favor of #95.  Please continue discussion there.

PR #57, opened 2021-12-28
SIP-018 Signed Structured Data

MarvinJanssen:
The Signed Structured Data specification proposed in this SIP describes a standard way to deterministically encode and sign Structured Data. Structured Data is data that can be represented in human-readable format and used in applications and smart contracts. The aim of the standard is to produce signatures that are straightforward and inexpensive to verify by smart contracts.

Reference implementation:
- https://github.com/MarvinJanssen/stx-signed-structured-data

Requires:
- `to-consensus-buff` / CV serialisation in Clarity as described in https://github.com/blockstack/stacks-blockchain/issues/2693. I am currently using a custom REPL / Clarinet that includes this function. I will make an effort to publish the branch as soon as possible.

See also:
- https://github.com/hirosystems/stacks-wallet-web/issues/1051
- https://github.com/blockstack/stacks-blockchain/issues/2693
- https://github.com/blockstack/stacks-blockchain/issues/2982


sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,204 @@
...
+
+Title: Signed Structured Data
+
+Author: Marvin Janssen <https://github.com/MarvinJanssen>
+
+Consideration: Technical
+
+Type: Standard
+
+Status: Accepted
```
jcnelson, line 13: Point of order -- the status of this PR is "draft".

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,204 @@
...
+
+Created: 28 December 2021
+
+License: CC0-1.0
+
+Sign-off: Jude Nelson <jude@stacks.org>, Technical Steering Committee Chair
```
jcnelson, line 19: Point of order -- I have not signed off on this yet.

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,204 @@
...
+
+Layer: Applications
+
+Discussions-To: https://github.com/stacksgov/sips
+
+# Abstract
+
+The Signed Structured Data specification describes a standard way to deterministically encode and sign Structured Data. Structured Data is data that can be represented in human-readable format and used in applications and smart contracts. The aim of the standard is to produce signatures that are straightforward and inexpensive to verify by smart contracts.
```
jcnelson, line 27: Formatting: can you make it so that each line is no more than ~100 characters long?  Otherwise, patching this document will generate a lot of merge conflicts.

jcnelson: Can you be more specific about "Structured Data"?  From this document, it looks like "Structured Data" is just any Clarity data that has a canonical encoding according to SIP 005 (so, Clarity values).  The "human readable" descriptor does not appear to be relevant or informative for what you're trying to do here.

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,204 @@
...
+
+# License and Copyright
+
+This SIP is made available under the terms of the Creative Commons CC0 1.0 Universal license, available at https://creativecommons.org/publicdomain/zero/1.0/
+This SIP's copyright is held by the Stacks Open Internet Foundation.
+
+
+# Introduction
+
+Digital signatures are at the heart of blockchains. They allow users to trustlessly transfer assets, invoke smart contracts and more. To perform these actions, a user signs a transaction and broadcasts it to the network. However, there are situations in which signed proofs without having to broadcast transactions are desirable. A few common use-cases include: (1) prove to an external application or entity that a user is in control of an address; (2) authorise an action to be performed by a smart contract at a later stage (like a meta transaction); (3) participate in an off-chain mechanism that is later settled on-chain (like a subnet).
```
jcnelson, line 37: A few things here:

* "Trustlessly" isn't correct -- there is definitely some trust involved in a blockchain's operation.  Can we say "they allow users to transfer assets, invoke smart contracts, and more, without a designated trusted third party," which is more correct?

* "Signed proofs" isn't a term I'm familiar with.  Maybe it would be better to say "there are situations in which it is desirable to submit signed data to a blockchain without encoding it as a transaction"?

* Formatting: Can you separate these list items with newlines?

* What's a "meta transaction?" This is the first time it's been used in this document.

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,204 @@
...
+
+It is important that signed messages are understandable for humans. For transactions, this is obvious: wallet applications display whom is receiving how many tokens of what kind. Likewise, the input parameters, function name, and target contract of contract calls are properly listed. Signed messages that are not transactions should be no different.
```
jcnelson, line 39: I'm confused by the meaning of this paragraph.  Why is it undesirable that it's possible to post signed encrypted data (i.e. not understandable to humans) to a blockchain?

I think you might be trying to advocate for representing structured data as a Clarity value (even when alternative structured data representations exist, such as JSON), but this isn't making that case.

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,204 @@
...
+
+The language properties of Clarity make producing and verifying such signed messages intuitive. The specification therefore leverages existing standards and encoding schemes. Messages are represented as native Clarity values and encoded in wire format in preparation for signing. It also makes sure that these signed messages are chain and application specific. There is a focus on ease of verification on the smart contract level.
```
jcnelson, line 41: What is it about the language that makes it intuitive?  Are you referring to the SIP 005 encoding of Clarity values?  If so, then can you just say that specifically here?

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,204 @@
...
+
+Below is an example of what a wallet application could show when signing a SIP018 message.
+
+![Wallet SIP018 concept](SIP-018-001.png)
+
+# Specification
+
+The challenge lies in producing messages that are both meaningful to humans as well as easy to be processed on-chain. Luckily, Clarity is a strongly-typed interpreted language. Structured Data is therefore a Clarity Value expression as detailed in *SIP002: Smart Contract Language*. These value expressions are encoded in Stacks wire format, as detailed in *SIP005: Blocks and Transactions*, and then hashed with SHA256. The resulting hash is used as the input for signing.
```
jcnelson, line 49: Reading ahead from here, I don't think that what this spec is calling for is a "structured data" spec that is "human meaningful."  The real proposal of this SIP is a new Clarity function `to-consensus-buff` that takes a Clarity value as input and returns a `(buff X)` type as output, where `X` is the number of bytes required to encode the given Clarity value according to SIP 005.  This proposal is not obvious at all until halfway through the spec.

Instead of trying to argue for this SIP in terms of needing a structured data specification, can it instead simply propose `to-consensus-buff` outright, and then use some motivating examples to justify the breaking change?  For example, you'd need `to-consensus-buff` if you want to implement something like Ethereum's meta-transactions.  You'd probably want to add other examples as well.

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Definitions
+
+*Clarity Value*: a native Clarity value as expressed in *SIP002: Smart Contract Language*.
+
+*SHA256*: a cryptographic hash function producing hashes with a length of 32 bytes.
```
jcnelson, line 55: Why not just link the NIST spec directly?

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,204 @@
...
+
+*Signature*: a proof produced by a signing algorithm, in this case ECDSA with the `secp256k1` curve, encoded in RSV order.
```
jcnelson, line 57: Signatures are usually not referred to as "proofs".  Why not just say that you're referring to a `secp256k1` ECDSA signature?  Also, `RSV order` is not standardized anywhere -- what does this mean?  Do you mean a recoverable signature?

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Formal specification
+
+To obtain a signature *S* of structured data *D* using private key *K*:
+
+- `S = signStructuredData(D, K)`; where,
+
+- `signStructuredData(D, K) = sign(messageHash(D), K)`; where `sign` produces a `secp256k1` signature of a message hash using private key *K*; and,
+
+- `messageHash(D) = sha256(structuredDataPrefix || domainHash || structuredDataHash(D))`; where `structuredDataPrefix` is a static value of `"\xC0"`, and `domainHash` and `structuredDataHash` are as described below.
```
jcnelson, line 71: I'm not convinced that this SIP needs to talk about structured data prefixes or domain hashes as a requirement.  A SIP that's proposing a breaking change (in this case, `to-consensus-buff`) should just focus on the breaking change.  A downstream consumer of `to-consensus-buff` would have its own orthogonal security and namespacing considerations that motivate these two fields.  I'd just drop all discussion of this completely, and put it into a separate SIP.

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,204 @@
...
+The length `len` of the string fields are not important as long as the resulting `domain` Clarity Value does not exceed the permitted maximum length. It is encouraged that wallet applications display this information to the user. Other fields may be added per the discretion of the application developer. For example, another differentiating field may be a `verifying-contract` that contains the principal of the contract performing the signature validation.
+
+**Definition of Structured Data**
+
+Structured Data can be any valid Clarity Value except `buff`. It is assumed that it will usually take the form of a `tuple` type, as it allows for multiple fields with readable names to be encoded. The standard does permit it to be another type to remain flexible. Still, developers should consider the informational content of requesting a user to sign a simple type like a `principal`, `int`, `uint`, or `bool`. A tuple can of course contain any type, including `buff`.
+
+A future SIP may define standard types of Structured Data.
+
+## Collisions and replay attacks
+
```
jcnelson, line 113: So, notwithstanding the recommendation that discussions pertaining to how to safely encode Clarity data belongs in a separate SIP, it's not the case that the transaction version is limited to `0x00` or `0x80` in the future.  In fact, the transaction codec only requires that the leading bit be set in the transaction in order for it to be considered a testnet transaction.  So, `0x01` is a valid mainnet transaction version, as are `0x02..0x7f`.  Bytes `0x80...0xff` are for testnet.

jcnelson (commented):
Thank you for submitting this SIP!  Based on my initial read, my recommendation is to split this SIP into two or more SIPs with the following proposals:

* A SIP that proposes `to-consensus-buff`, which is a breaking change and thus needs a `consensus` consideration.  It could activate with, but no sooner than Stacks 2.1.

* A SIP that proposes a data signing and verifying scheme built around `to-consensus-buff` that recommends signed data begin with a canonical preamble that encodes a message prefix constant, a domain, a chain ID, etc.  This could activate anytime after `to-consensus-buff` goes live, and it does *not* need `consensus` consideration.

sips/sip-018/sip-018-signed-structured-data.md
MarvinJanssen: The language in SIP-000 seemed unclear to me.
>- _Draft_. The SIP is still being prepared for formal submission. It does not yet
  have a SIP number.

Pending feedback, the SIP is considered sufficiently complete and thus submitted.

Will change to "Draft".

sips/sip-018/sip-018-signed-structured-data.md
MarvinJanssen: Misread—thought I had to select the person to sign off.

sips/sip-018/sip-018-signed-structured-data.md
MarvinJanssen: 1. Sure, that is what I meant.
2. Sure. But this data is not necessarily meant to be submitted to a blockchain. For example, many dapps in the EVM-space and beyond authenticate users on centralised components using signed messages.
3. Ok.
4. Added under "Definitions".

sips/sip-018/sip-018-signed-structured-data.md
MarvinJanssen: I never stated it is undesirable that it is possible to post signed encrypted data to a blockchain. I stated that it is desirable that the user knows what he or she is signing when an app request a signature. Example: there is a lot of precedent in the Ethereum world where users were prompted to sign hashes of 0xProtocol order data and messing it up / getting scammed and losing their tokens. No way to verify that the order struct hash you are signing in your wallet app matches what you see on the app.

Here is what that looked like:
bad 

We don't want to go down that route.

jcnelson: > I never stated it is undesirable that it is possible to post signed encrypted data to a blockchain.  I stated that it is desirable that the user knows what he or she is signing when an app request a signature

I don't know how to reconcile "It is important that signed messages are **understandable for humans**" (emphasis mine) with the above, in the context of describing a data representation that lives in the Clarity smart contract.  As written, the text is calling for a structured data schema that pertains to how _Clarity values_ are represented within the execution context of a _smart contract_.  Making sure that users know what they're signing is all well and good, but that's not what the text says.  In general, this is not something the Clarity VM can enforce, nor is it something that a structured data schema can ever hope to mandate (since "understandable to humans" is in the eye of the beholder, not the specification).

Can you change this text to clarify this?

jcnelson: I think I understand what you're trying to say.  You're saying that in the context of users signing data with a key that potentially owns data on a blockchain, there are two important usability considerations:

* The user must be able to understand what it is they are signing, as well as the implications of publishing the signed message in a public space (such as a blockchain).

* Any software that supports signing data besides transactions should structure this data such that it is all but impossible for the resulting signature to be a signature for a valid transaction (this is usually achieved by signing both the data and a preamble, like what you propose here).

Part of this SIP is to propose a schema for point (2).  I think this is a worthy goal, but I also think it should be a separate SIP from the SIP that describes `to-consensus-buff`.  In general, we'd like to keep all the Stacks 2.1 SIPs as "small" as possible, so that they only describe the breaking changes they create.

sips/sip-018/sip-018-signed-structured-data.md
jcnelson: In light of other things you've said on Discord, I think I now see what you mean here.  You want it to be the case that the data being encoded has a representation that's amenable to human interpretation.  It's not limited to Clarity values, nor does it exclude e.g. a bare hash with a preimage known to the signer.  Is that understanding correct?

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,210 @@
...
+This SIP's copyright is held by the Stacks Open Internet Foundation.
+
+
+# Introduction
+
+Digital signatures are at the heart of blockchains. They allow users to transfer assets, invoke smart contracts and more, without a designated trusted third party. To perform these actions, a user signs a transaction and broadcasts it to the network. However, there are situations in which it is desirable to produce signed messages that are not transactions. A few common use-cases include:
+
+1. prove to an external application or entity that a user is in control of an address;
+2. authorise an action to be performed by a smart contract at a later stage (like a meta transaction, see below); 
+3. participate in an off-chain mechanism that is later settled on-chain (like a subnet).
```
jcnelson, line 41: These are called "hyperchains" now :)

sips/sip-018/sip-018-signed-structured-data.md
jcnelson: Alternatively, discussion of `to-consensus-buff` could get folded into SIP 015 and dropped here.

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,210 @@
...
+
+`chain-id` contains the ID of the chain the Signed Structured Data is targeting.
+
+The length `len` of the string fields are not important as long as the resulting `domain` Clarity Value does not exceed the permitted maximum length. It is encouraged that wallet applications display this information to the user. Other fields may be added per the discretion of the application developer. For example, another differentiating field may be a `verifying-contract` that contains the principal of the contract performing the signature validation.
+
+**Definition of Structured Data**
+
+Structured Data can be any valid Clarity Value except `buff`. It is assumed that it will usually take the form of a `tuple` type, as it allows for multiple fields with readable names to be encoded. The standard does permit it to be another type to remain flexible. Still, developers should consider the informational content of requesting a user to sign a simple type like a `principal`, `int`, `uint`, or `bool`. A tuple can of course contain any type, including `buff`.
```
jcnelson, line 114: Why exclude `buff`?  If the requirement is that the data being signed is comprehensible by the signer, then does the wire format matter?  Also, a nefarious contract would find a different way to encode opaque data, so limiting `buff` would only hinder non-nefarious developers.

sips/sip-018/sip-018-signed-structured-data.md
jcnelson: If you want a prefix that is guaranteed to never be a valid transaction, you could use the byte sequence `0x00 0x00 0x00 0x00 0x00 0x00`.  This encodes a mainnet version (first byte), a chain ID of 0 (next 4 bytes), and a currently-invalid transaction authorization type (`0x00` -- only `0x04` and `0x05` are supported, per SIP 005).

jcnelson:
I took another pass at this, in light of the recent OpenSea phishing attack.  While I stand by my belief that discussion of `to-consensus-buff` should be part of the Stacks 2.1 SIP, I think the rest of this is good.  Most of my comments are just asking for minor clarifications.

friedger:
How would a Verifiable Credential be signed with this structure?

sips/sip-018/sip-018-signed-structured-data.md
MarvinJanssen: Good point.

radicleart:
Marvin - a question on making the domain info stronger by eg including domain names and user agent info?

cf. EIP 712 says this.. 

```The way to solve this is by introducing a domain separator, a 256-bit number. This is a value unique to each domain that is ‘mixed in’ the signature. It makes signatures from different domains incompatible. The domain separator is designed to include bits of DApp unique information such as the name of the DApp, the intended validator contract address, the expected DApp domain name, etc. The user and user-agent can use this information to mitigate phishing attacks, where a malicious DApp tries to trick the user into signing a message for another DApp.```

**Comment by aulneau** [comment]
Hi all -- just wanted to let folks know that I've implemented support for interacting with wallets that support this standard within micro-stacks:

Relevant versions:
micro-stacks: >= 0.5.0
@micro-stacks/react: >= 0.2.1

Some docs:
https://docs.micro-stacks.dev/modules/core/connect/message-signing
https://docs.micro-stacks.dev/modules/integrations/react/message-signing

File: sips/sip-018/sip-018-signed-structured-data.md [hunk 14]
```diff
@@ -0,0 +1,210 @@
...
+# Activation
+
+This SIP will be considered activated if the following two conditions are met:
+
+1. `to-consensus-buff` or equivalent functionality is added to Stacks 2.1.
+2. At least one popular wallet application implements the standard before or within three months of the launch of Stacks 2.1.
+
+# Reference Implementations
+
+https://github.com/MarvinJanssen/stx-signed-structured-data
```
janniks, line 209: Would it make sense to include examples/test-vectors similar to [BIP322 Test vectors →](https://github.com/bitcoin/bips/blob/master/bip-0322.mediawiki#test-vectors)

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,210 @@
...
+# Backwards Compatibility
+
+Not applicable
+
+# Activation
+
+This SIP will be considered activated if the following two conditions are met:
+
+1. `to-consensus-buff` or equivalent functionality is added to Stacks 2.1.
+2. At least one popular wallet application implements the standard before or within three months of the launch of Stacks 2.1.
```
jcnelson, line 205: You can easily give yourself a year for this if you wanted.

jcnelson:
Did another pass.  Let's go ahead and advance this to Accepted :tada: 

Before we can advance to Recommended, I have two outstanding comments that need to be addressed.  Also, @kantai (who is now on the technical CAB) needs to make a pass as well.

sips/sip-018/sip-018-signed-structured-data.md
MarvinJanssen: Great idea.

sips/sip-018/sip-018-signed-structured-data.md
MarvinJanssen: Fair. And I guess it might become a moot point since Hiro has since begun implementation. (Although it is not yet according to the standard.)
- https://github.com/hirosystems/connect/issues/225

sips/sip-018/sip-018-signed-structured-data.md
MarvinJanssen: Would you recommend such a prefix instead of a single byte? Is there a non-zero prefix we could use and specifically reserve it for signed messages?

MarvinJanssen:
I added test vectors at the end of the SIP per @janniks in 395ea83 and moved it to Accepted status. Still an open question regarding the prefix. (See comment above.)

MarvinJanssen:
@radicleart I'm wary adding a domain name to the domain object. I can foresee future applications that might not use a domain name or do not want to be strongly coupled to one. For example, an open DeFi protocol built on top of SIP018 might be used across different applications. The protocol then simply defines a unique name or other identifier in the domain object. That is not to say a future standard based on SIP018 cannot introduce a domain name in the domain object for specific purposes (like app sign in).

jiga:
I propose adding `signature timestamp`  and `signature expiration timestamp` fields as part of domain.
this may help mitigate potential misuse of signed messages e.g. prevent replay of meta transactions by compromised dapps
ideally `signature timestamp` to be autogenerated while signing messages

sips/sip-018/sip-018-signed-structured-data.md
jcnelson: Sure, this SIP could propose reserving a 1-byte prefix, and declare that this value can never be a transaction version.  How about `0xff`, or `0x7f`?

jcnelson: To be clear, we'll need to add this to Stacks 2.1 so it will never accept a transaction with this version (but now is the time to do that).  If you go this route, please also update the Activation section to require that Stacks 2.1 ships with this change.

MarvinJanssen:
@jiga I don't think those fields should be part of SIP018 as I don't think they are always desirable. Furthermore, those fields should probably go in the payload and not the domain. (An `order` tuple may have an `expiry` field, for example.)

Meta transactions are a valid use case and I would suggest a future meta transaction SIP to be formulated on top of SIP018.

sips/sip-018/sip-018-signed-structured-data.md
MarvinJanssen: I'm partial to `0x7f` if there is a strong reason we shouldn't the already used `0xc0`. Also, I feel this SIP could activate without requiring Stacks 2.1 block the prefix byte as it is currently unused. As a comparison, EIP712's prefix of `[0x19, 0x01]` was also not explicitly blocked in the node as far as I know. It was chosen because no valid RLP encoding produces a byte string starting with `0x19`.

jcnelson: > Also, I feel this SIP could activate without requiring Stacks 2.1 block the prefix byte as it is currently unused. 

We don't have a choice.  The network will accept a transaction with version byte `0x7f` today.  If we're going to reserve a byte to indicate a SIP-018 payload, then we must also update Stacks so it will _never_ accept a transaction that starts with that byte.  Otherwise you run the risk of malicious SIP-018 apps tricking users into signing well-formed transactions.

MarvinJanssen: Ah, I didn't know that! I thought it would reject the transaction if the first/version byte were `0xc0` (or `0x7f` for that matter). I should've checked that instead of making the assumption. So the safest prefix to pick right now would be something like `[0xc0, 0x00, 0x00, 0x00, 0x00, 0xc0]` or `[0x7f, 0x00, 0x00, 0x00, 0x00, 0x7f]`.

Given that full support cannot happen before Stacks 2.1 any (due to the inability to hash principals), we could try for the solution you suggested before; that is, block that specific version byte in the node. What do you think? I'm still leaning towards a single prefix byte over six. Seems more optimal. But whatever we go for, in the end it is just a matter of updating the SIP and reference implementations.

jcnelson: I think one of these options would work:

* Stipulate a 6-byte prefix with the above constraints 
* Stipulate a 6-byte prefix, and use the first 5 bytes to encode something useful a versioning scheme, a payload length, a vanity string, or what-not.
* Stipulate a 1-byte prefix that activates with 2.1, *and* add a very explicit warning to this SIP that implementors _should not_ eagerly implement this SIP until Stacks 2.1 (given the security risk)

MarvinJanssen: The last one is not great because SIP018 can definitely be used today. To add, Hiro is actively implementing the standard right now so  we are going to see these messages very soon. Once that happens it might be painful to change the prefix.
- https://github.com/hirosystems/connect/issues/225
- https://github.com/hirosystems/stacks-wallet-web/issues/2460

>- Stipulate a 6-byte prefix, and use the first 5 bytes to encode something useful a versioning scheme, a payload length, a vanity string, or what-not.

I considered this as well but the domain already contains that information. Domain objects are easier to assemble for end devs over remembering the right encoding for the prefix. A static prefix is probably a lot easier.
```clarity
(to-consensus-buff {domain: "My App", chain-id: chain-id, version: "1.0.0"})
```

How about this prefix: `[0x53, 0x49, 0x50, 0x30, 0x31, 0x38]`? It spells `SIP018`.

MarvinJanssen: To add: the readable prefix should make the reference implementation and those of end devs straightforward, e.g.:

```clarity
(define-constant sip018-prefix (to-buff "SIP018"))
(define-constant domain-hash (sha256
 {
  name: "My App",
  chain-id: chain-id,
  version: "1.0.0"
 }
))
(define-constant message-hash-prefix (concat sip018-prefix domain-hash))

(define-read-only (sip018-message-hash (payload {some-type}))
  (sha256 (concat message-hash-prefix (sha256 (to-consensus-buff payload))))
)
```

jcnelson: > The last one is not great because SIP018 can definitely be used today. To add, Hiro is actively implementing the standard right now so we are going to see these messages very soon.

Yeah, that's what I thought.  Looks like it needs to be option 1 or 2.

> How about this prefix: [0x53, 0x49, 0x50, 0x30, 0x31, 0x38]? It spells SIP018

Sounds good :+1:

MarvinJanssen:
SIP document updated with the new "SIP018" prefix in 318bad3!

To everyone that was/is implementing SIP018: this is a **BREAKING CHANGE**.

The prefix has been changed from a single-byte `0xC0` to six bytes: `[0x53, 0x49, 0x50, 0x30, 0x31, 0x38]`. (Which spells "SIP018" in ASCII.)

If you are curious why, see this discussion: https://github.com/stacksgov/sips/pull/57#discussion_r776508478

The reference implementation has also been updated to reflect the change:
- https://github.com/MarvinJanssen/stx-signed-structured-data

aulneau:
`micro-stacks@0.5.1` has been updated to support the latest implementation of this SIP, matching the test vectors supplied in https://github.com/stacksgov/sips/commit/318bad31f09fb60b02181f1d1c81b26ba8254408

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,386 @@
...
+
+Discussions-To: https://github.com/stacksgov/sips
+
+# Abstract
+
+The Signed Structured Data specification describes a standard way to
+deterministically encode and sign Structured Data. Structured Data is data that
+can be represented in human-readable format and used in applications and smart
+contracts. The aim of the standard is to produce signatures that are
+straightforward and inexpensive to verify by smart contracts.
```
kantai, line 31: Can you add a sentence to the abstract that states that this SIP also proposes two new Clarity methods?

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,386 @@
...
+interpreted language. Structured Data is therefore a Clarity Value expression as
+detailed in _SIP002: Smart Contract Language_. These value expressions are
+encoded in Stacks wire format, as detailed in _SIP005: Blocks and Transactions_,
+and then hashed with SHA256. The resulting hash is used as the input for
+signing.
+
+## Definitions
+
+_Clarity Value_: a native Clarity value as expressed in _SIP002: Smart Contract
+Language_.
```
kantai, line 85: Can you add links to these three SIP references?

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,386 @@
...
+The `domain` is a `tuple` type Clarity Value that represents the domain in which
+messages are signed. Its Clarity type definition is as follows:
+
+```clojure
+{
+	name: (string-ascii len)
+	version: (string-ascii len)
+	chain-id: uint
+}
+```
```
kantai, line 148: Using `clar` as the programming language would be better:

```clar
{
	name: (string-ascii len)
	version: (string-ascii len)
	chain-id: uint
}
```

sips/sip-018/sip-018-signed-structured-data.md
kantai: Never mind, I see that this SIP doesn't itself propose new functions.

sips/sip-018/sip-018-signed-structured-data.md
```diff
@@ -0,0 +1,386 @@
...
+
+**Definition of** `domain`
+
+The `domain` is a `tuple` type Clarity Value that represents the domain in which
+messages are signed. Its Clarity type definition is as follows:
+
+```clojure
+{
+	name: (string-ascii len)
+	version: (string-ascii len)
```
janniks, line 145: Could there be disadvantages of excluding utf strings as a data type for `name`? i.e. not allowing Japanese/etc. characters. Internet hostnames get around this by using [Punycode](https://en.wikipedia.org/wiki/Punycode) which is an ascii subset — might be good enough but probably depends on how the signed data is intended to be used (more specifically, how the `domain` is additionally matched).

Maybe not an issue, but I wanted to bring it up...

MarvinJanssen: ASCII was chosen on purpose because I assumed that the name in the domain object would be used for "app identity" purposes by the wallet. I wanted to prevent people mimicking different names using UTF8 tricks. I think a future SIP could define a proper "display name" or similar, as you say.

However, we would not want limit the SIP arbitrarily in favour of "Western language". What do you think considering the above?

MarvinJanssen:
Implemented suggestions and moved to Recommended.

jcnelson:
Cool!  Do the authors feel that this is ready for activation?

MarvinJanssen:
Yep. But will give @janniks time to respond to the last open conversation.

sips/sip-018/sip-018-signed-structured-data.md
janniks: Okay, thanks — seems reasonable and important for user safety.

jcnelson:
Wanna ping me when it's ready? @janniks @MarvinJanssen

MarvinJanssen:
Ready to go @jcnelson

jcnelson:
Cool, it's now listed as "activating" in the README.  Can you set the status to `Activation-in-Progress`?  Thanks!

MarvinJanssen:
Done!

MarvinJanssen:
Looking at the activation criteria:

>1. `to-consensus-buff` or equivalent functionality is added to Stacks 2.1.
>2. At least one popular wallet application implements the standard before or within one year of the launch of Stacks 2.1.

Do you think both have been met @jcnelson? Hiro Wallet as well as the Stacks Ledger applet added support for SIP018 messages and `to-consensus-buff` has been added to Stacks 2.1. Can we thus consider this SIP activated or do we wait until the launch of Stacks 2.1?

jcnelson:
Just an update on this -- we will merge this when 2.1 activates on mainnet.

PR #59, opened 2022-01-20
SIP for Wallet Protocol

friedger:
This SIP specifies how applications and wallet should exchange messages and how applications and wallets agree on the protocol.

While writing "SIP for Authentication Protocol", it became clear that the communication should be handled in a separate SIP.

### DIFF
--- a/sips/sip-x/sip-wallet-protocol.md
+++ b/sips/sip-x/sip-wallet-protocol.md
@@ -42,7 +42,7 @@ The communication between application and authenticator can happen in various wa
 
 ### Stacks Provider
 
-Stacks Provider is a common interface used for web applications to communicate with the authenticator via a browser extension.
+The _Stacks Provider_ is a common interface exposed to dApps as a JavaScript object, often injected by a Web Extension.
 
 It provides functions to handle
 
--- a/sips/sip-x/sip-wallet-protocol.md
+++ b/sips/sip-x/sip-wallet-protocol.md
@@ -1,0 +1,84 @@
+
+# Preamble
+
+SIP Number: X
+
+Title: Specification of the Wallet Protcol
+
+Author: Friedger Müffke (mail@friedger.de)
+
+Consideration: Technical
+
+Type: Standard
+
+Status: Draft
+
+Created: 20 January 2022
+
+License: CC0-1.0
+
+Sign-off:
+
+# Abstract
+
+Decentralized application do not handle private keys of users. Instead users have wallets that manage private keys, authentication, message signing etc. This SIP specifies a ways to exchange messages between applications and wallets and how to agree on one.
+
+# License and Copyright
+
+This SIP is made available under the terms of the Creative Commons CC0 1.0 Universal license, available at https://creativecommons.org/publicdomain/zero/1.0/
+This SIP’s copyright is held by the Stacks Open Internet Foundation.
+
+# Introduction
+
+Decentralized application do not want to store credentials of their users. Instead users use a trusted wallet that manages the user's private keys savely. Therefore, applications needs a way to communicate with the user's wallet e.g. for requesting signatured of a transaction, authentication or message signing.
+
+Different transport protocols for exchanging messages between applications and wallets have been established in the web3 ecosystem. In the following sections the transport protocols are described and a standardized way how to agree on a transport protocol is defined.
+
+# Specification
+
+## Transport Protocols
+
+The communication between application and authenticator can happen in various ways. The subsections below define common transport protocols.
+
+### Stacks Provider
+
+Stacks Provider is a common interface used for web applications to communicate with the authenticator via a browser extension.
+
+It provides functions to handle
+
+1. authentication
+2. transaction signing
+
+https://github.com/hirosystems/connect/blob/main/packages/connect/src/types/provider.ts
+
+The function `authenticationRequest` expects the JWT of the authentication request as parameter and must return the authentication response as encoded JWT.
+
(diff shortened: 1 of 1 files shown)


kyranjamie:
Thanks for starting this SIP, @friedger. 

What are your thoughts on following a scheme similar to [EIP-1102](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1102.md), whereby we use a JSON RPC style API, rather than a set of methods. I'd be in favour of this.

kyranjamie:
> A rpc style api makes sense because it would work also for a https provider

Another benefit is that, with Web Extensions, it'll be easier to forward events when they go through the same method. For each new method, a new handler is needed in the forwarding script contexts.

_In page script context_ → _Content script context_ → _Background script context_

sips/sip-x/sip-wallet-protocol.md
```diff
@@ -0,0 +1,84 @@
...
+Decentralized application do not handle private keys of users. Instead users have wallets that manage private keys, authentication, message signing etc. This SIP specifies a ways to exchange messages between applications and wallets and how to agree on one.
+
+# License and Copyright
+
+This SIP is made available under the terms of the Creative Commons CC0 1.0 Universal license, available at https://creativecommons.org/publicdomain/zero/1.0/
+This SIP’s copyright is held by the Stacks Open Internet Foundation.
+
+# Introduction
+
+Decentralized application do not want to store credentials of their users. Instead users use a trusted wallet that manages the user's private keys savely. Therefore, applications needs a way to communicate with the user's wallet e.g. for requesting signatured of a transaction, authentication or message signing.
```
kyranjamie, line 33:
```suggestion
Decentralized applications, _dApps_, do not want to store credentials of their users. Instead users use a trusted wallet that manages the user's private keys savely. Therefore, applications needs a way to communicate with the user's wallet e.g. for requesting signatured of a transaction, authentication or message signing.
```

sips/sip-x/sip-wallet-protocol.md
```diff
@@ -0,0 +1,84 @@
...
+This transport protocol is implemented by the Blockstack Browser and the Stacks cli.
+
+### App Links
+
+On mobile devices, applications can use app links/deep links to send authentication requests and receive the response. They must use the same query parameters for the authentication requests and responses as the HTTPS protocol, i.e. `authRequest` and `authResponse`.
+
+This transport protocol is implemented by the following authenticator apps:
+
+- [Wise app](https://github.com/PravicaInc/wise-js) and client library[wise-js](https://github.com/PravicaInc/wise-js)
+- [Circles app](https://github.com/blocoio/stacks-circles-app) for Android.
```
kyranjamie, line 73: Aren't these details better suited to [your other SIP](https://github.com/stacksgov/sips/pull/50)?

kyranjamie (commented):
Drawing parallels from Ethereum, currently this SIP encompasses:
- [EIP-1102](https://eips.ethereum.org/EIPS/eip-1102) Opt-in account exposure
- [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193) Ethereum Provider JavaScript API
- [EIP-1474](https://eips.ethereum.org/EIPS/eip-1474) Remote procedure call specification

@friedger I wonder if we should scope this to be more specific about the format/API of the provider protocol, and less about about _what_ methods/functionality it may support.

sips/sip-x/sip-wallet-protocol.md
```diff
@@ -0,0 +1,84 @@
...
+
+Created: 20 January 2022
+
+License: CC0-1.0
+
+Sign-off:
+
+# Abstract
+
+Decentralized application do not handle private keys of users. Instead users have wallets that manage private keys, authentication, message signing etc. This SIP specifies a ways to exchange messages between applications and wallets and how to agree on one.
```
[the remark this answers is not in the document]
jcnelson: Either text is fine with me.  RPC is a form of message exchange, and there will almost certainly be apps that use the message-passing paradigm instead of the RPC paradigm for communication.

sips/sip-x/sip-wallet-protocol.md
```diff
@@ -0,0 +1,84 @@
...
+
+Decentralized applications, _dApps_, do not want to store credentials of their users. Instead users use a trusted wallet that manages the user's private keys savely. Therefore, applications needs a way to communicate with the user's wallet e.g. for requesting signatured of a transaction, authentication or message signing.
+
+Different transport protocols for exchanging messages between applications and wallets have been established in the web3 ecosystem. In the following sections the transport protocols are described and a standardized way how to agree on a transport protocol is defined.
+
+# Specification
+
+## Transport Protocols
+
+The communication between application and authenticator can happen in various ways. The subsections below define common transport protocols.
```
jcnelson, line 41: Is an _authenticator_ different from a _wallet_ here?  If so, can you clarify the difference?

sips/sip-x/sip-wallet-protocol.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Stacks Provider
+
+The _Stacks Provider_ is a common interface exposed to dApps as a JavaScript object, often injected by a Web Extension.
```
jcnelson, line 45: Does the SIP require that the Stacks provider be a Javascript object, and does it require that there exists a Web extension to inject it?  If not, can you clarify that this is an implementation example, as opposed to a design requirement?  Thanks.

sips/sip-x/sip-wallet-protocol.md
```diff
@@ -0,0 +1,84 @@
...
+
+It provides functions to handle
+
+1. authentication
+2. transaction signing
+
+https://github.com/hirosystems/connect/blob/main/packages/connect/src/types/provider.ts
```
jcnelson, line 52: I think this is describing a particular implementation.  The SIP should describe the public interfaces and API contracts, as well as the transport protocol used to send and receive messages to interact with them.

friedger:
Replaced by SIP-030

PR #66, opened 2022-05-02
SIP-019: Use PoX for fuelling Bitcoin development

lgalabru:
Assuming the totality of the 4,000 slots can never be fully reserved, can we use this quantity of forever vacant slots for doing something good, that would be beneficial for the Stacks Blockchain, while staying neutral, from a Stacker / Miner / PoX point of view?
I'd like to use this PR to generate discussions around the idea of using what seems to be a "waste" in the PoX implementation. I acknowledge that the document in its current form is not conforming to what would be expected for an actual SIP, but this should be enough to kickstart conversations.

### DIFF
--- a/sips/sip-019/sip-019-bitcoin-development-funding.md
+++ b/sips/sip-019/sip-019-bitcoin-development-funding.md
@@ -1,0 +1,77 @@
+# Preamble
+
+SIP Number: 019
+
+Title: Fund Bitcoin Development
+
+Author: Ludo Galabru <ludo@hiro.so>
+
+Consideration: Technical
+
+Type: Consensus
+
+Status: Draft
+
+Created: 2 May 2022
+
+License: CC0-1.0
+
+Sign-off: 
+
+Discussions-To: https://github.com/stacksgov/sips
+
+# Abstract
+
+This document evaluates the consequences of turning the tokens being 
+burnt for minting Stacks into a source of funding for its base chain. 
+
+# Introduction
+
+The Stacks chain exists thanks to the base chain it anchors its blocks
+to. The PoX does reward the nodes building this chain, but does not help
+consolidating the development of this base chain.
+
+The Stacks chain launched in January 2021. Since then, the chain saw a 
+total of 32 PoX cycles.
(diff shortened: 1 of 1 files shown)


sips/sip-019/sip-019-bitcoin-development-funding.md
```diff
@@ -0,0 +1,77 @@
...
+| 24 | 3,755 | 245 | 72 | 4.41 |
+| 25 | 3,654 | 346 | 71 | 6.1415 |
+| 26 | 3,787 | 213 | 60 | 3.195 |
+| 27 | 3,701 | 299 | 56 | 4.186 |
+| 28 | 3,638 | 362 | 46 | 4.163 |
+| 29 | 3,685 | 315 | 55 | 4.33125 |
+| 30 | 3,899 | 101 | 54 | 1.3635 |
+| 31 | 3,673 | 327 | 46 | 3.7605 |
+
+Assuming that the data from stacking.club is correctly collected, over the last 31 cycles, looking exclusively at "vacant PoX slots", an approximate total of 146 BTC were burnt.
```
jcnelson, line 73: Wow, that's huge!  Arguably, burning BTC is in and of itself useful because it makes all BTC holders slightly richer ;)

MarvinJanssen:
>but in its most naive version, the Bitcoin core development team address could be used (instead of the 0x00 burn address) when Stacks miners are processing vacant PoX slots

That would be cool even.

At the Lab we have an idea on how we could potentially solve the "write back to Bitcoin problem". That might open the door to having a DAO on Stacks that manages the Bitcoin sent to vacant slots. A decentralised dev fund where all STX holders can vote, how much fun would that be?

GinaAbrams:
Great idea, Ludo!

jcnelson:
Is there a way to structure the mechanism so that a cartel of Stacks miners can't just take control of it to re-distribute BTC back to themselves?  That would permit the cartel to discount-mine during this time.

jcnelson:
Edited the title because this SIP does not yet have an assigned number.

Issue #69, opened 2022-05-18
Structured message for Sign-In with Stacks


friedger:
SIP 018 defines how any structured messages should be signed.
There should be a SIP that defines the message structure for sign-in with stacks that users can sign and services, web apps, off-chain entities accept as authentication.

PR #70, opened 2022-05-18
Sign-in with stacks

friedger:
This PR defines a standard for message used for sign-in with stacks (#69)

It is mainly inspired by EIP-4361.

The standard requires the following limits that can be discussed:
* URLs are limited to 80 ascii characters so that they can be shown on a normal screen.
* 10 resources can be added to the message.

### DIFF
--- a/sips/sip-x sign-in-with-stacks/sip-x-sign-in-with-stacks.md
+++ b/sips/sip-x sign-in-with-stacks/sip-x-sign-in-with-stacks.md
@@ -1,0 +1,87 @@
+# Preamble
+
+SIP Number: X
+
+Title: Sign In With Stacks
+
+Author: Friedger Müffke (mail@friedger.de), Leo Pradel
+
+Consideration: Technical
+
+Type: Standard
+
+Status: Draft
+
+Created: 18 May 2022
+
+License: CC0-1.0
+
+Sign-off:
+
+Layer: Applications
+
+Discussions-To: https://github.com/stacksgov/sips
+
+# Abstract
+
+Web application often provide their services only to authenticated users. In
+Web2, this was done through username and password or federated logins. In Web3,
+users can proof their digital identity by cryptographically signing that the
+user ownes the private key associated with that digital identity.
+
+SIP-018 defines the structure of signatures in general. This SIP defines the
+message format that web applications and similar off-chain services should use for their users to sign.
+
+# License and Copyright
+
+This SIP is made available under the terms of the Creative Commons CC0 1.0 Universal license, available at https://creativecommons.org/publicdomain/zero/1.0/
+This SIP’s copyright is held by the Stacks Open Internet Foundation.
+
+# Introduction
+
+# Specification
+
+Sign-In with Stacks works as follows:
+
+1. The wallet presents the user with a structured message as defined in SIP-018. The message is a clarity value of type `tuple` with the properties described below.
+2. The signature is then presented to the server, which checks the signature’s validity and message content.
+3. Additional fields, including expiration-time, not-before, request-id, chain-id, and resources may be incorporated as part of authentication for the session.
+4. The server may further fetch data associated with the public key, the stacks address, or other data sources that may or may not be permissioned.
+
+## Properties
+
+| name            | type                        | description                                                                                                                                                                        |
+| --------------- | --------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| title           | (string-ascii 305)          | Must be the application's domain name (max 255) followed by ` wants you to sign in with your Stacks account:`                                                                      |
+| address         | principal                   | The address of the signer                                                                                                                                                          |
+| statement       | (string-ascii 80)           | (optional) Describes the terms and conditions the user agrees to by using the application.                                                                                         |
+| URI             | (string-ascii 80)           | An RFC 3986 URI referring to the resource that is the subject of the signing (as in the subject of a claim).                                                                       |
+| version         | uint                        | is the current version of the message, which MUST be X for this specification.                                                                                                     |
+| chain-id        | uint                        | (optional) the chain ID to which the session is bound. This must correspond to the version of the address.                                                                         |
+| nonce           | (string-ascii 64)           | randomized token used to prevent replay attacks, at least 8 alphanumeric characters.                                                                                               |
+| issued-at       | (string-ascii 27)           | The ISO 8601 datetime string of the current time.                                                                                                                                  |
+| expiration-time | (string-ascii 27)           | (optional) The ISO 8601 datetime string that, if present, indicates when the signed authentication message is no longer valid.                                                     |
+| not-before      | (string-ascii 27)           | (optional) The ISO 8601 datetime string that, if present, indicates when the signed authentication message will become valid.                                                      |
+| request-id      | (string-ascii 64)           | an system-specific identifier that may be used to uniquely refer to the sign-in request.                                                                                           |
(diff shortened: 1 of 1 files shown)


sips/sip-x sign-in-with-stacks/sip-x-sign-in-with-stacks.md
```diff
@@ -0,0 +1,91 @@
...
+| nonce           | (string-ascii 64)           | randomized token used to prevent replay attacks, at least 8 alphanumeric characters.                                                                                               |
+| issued-at       | (string-ascii 27)           | The ISO 8601 datetime string of the current time.                                                                                                                                  |
+| expiration-time | (string-ascii 27)           | (optional) The ISO 8601 datetime string that, if present, indicates when the signed authentication message is no longer valid.                                                     |
+| not-before      | (string-ascii 27)           | (optional) The ISO 8601 datetime string that, if present, indicates when the signed authentication message will become valid.                                                      |
+| request-id      | (string-ascii 64)           | an system-specific identifier that may be used to uniquely refer to the sign-in request.                                                                                           |
```
pradel, line 65: request-id should be optional too @friedger

sips/sip-x sign-in-with-stacks/sip-x-sign-in-with-stacks.md
```diff
@@ -0,0 +1,91 @@
...
+
+Layer: Applications
+
+Discussions-To: https://github.com/stacksgov/sips
+
+# Abstract
+
+Web application often provide their services only to authenticated users. In
+Web2, this was done through username and password or federated logins. In Web3,
+users can proof their digital identity by cryptographically signing that the
```
jcnelson, line 29: typo: "prove"

sips/sip-x sign-in-with-stacks/sip-x-sign-in-with-stacks.md
```diff
@@ -0,0 +1,91 @@
...
+| version         | uint                        | is the current version of the message, which MUST be X for this specification.                                                                                                     |
+| chain-id        | uint                        | (optional) the chain ID to which the session is bound. This must correspond to the version of the address.                                                                         |
+| nonce           | (string-ascii 64)           | randomized token used to prevent replay attacks, at least 8 alphanumeric characters.                                                                                               |
+| issued-at       | (string-ascii 27)           | The ISO 8601 datetime string of the current time.                                                                                                                                  |
+| expiration-time | (string-ascii 27)           | (optional) The ISO 8601 datetime string that, if present, indicates when the signed authentication message is no longer valid.                                                     |
```
jcnelson, line 63: One of the tricky things with authentication tokens like this is making sure they can't be replayed by someone who captures them.  Is there a way we can mandate `expiration-time` or perhaps use a `timeout`?  Maybe we could use the Stacks blockchain height to determine how long a login token is good for (e.g. "This token expires at Stacks block height 12345").

obycode:
I'd love to see this revived. Has anyone built any implementations of this?

pradel:
@obycode I have a working implementation for Sigle, if you are interested in using it I can publish it on npm. Can extract it to a smaller repo and publish it there https://github.com/sigle/sigle/tree/main/packages/sign-in-with-stacks

pradel:
Published it as a separate package on npm `npm i sign-in-with-stacks` and github at https://github.com/pradel/sign-in-with-stacks for anyone to use!

314159265359879 (commented):
Good ideas here: the EIP-4361 message grammar, the upstream replay/domain-binding security analysis, and near-zero developer re-education cost are all good reasons to base this on SIWE.

**Align normatively with CAIP-122.** CAIP-122 generalizes SIWE into a chain-agnostic data model, making EIP-4361 one namespace implementation of it rather than the root standard (Haardik & Ukustov, 2022; Chang, Rocco, Millegan, Johnson, & Terbu, 2021). A Stacks namespace profile already exists and specifies `stacks:secp256k1` with SHA256 pre-hashing (Pradel, 2023). This SIP currently restates that profile instead of referencing it. Referencing it is leverage: Stacks inherits SIWX-capable adapters rather than needing bespoke per-app integration — `siwx` already lists Stacks as a supported namespace (LearnWeb3DAO, n.d.), and x402 ships CAIP-122 auth today (Coinbase Developer Platform, n.d.). Chain-agnostic auth plugins are also being requested upstream in Better Auth (better-auth, 2025).

**App side** stays as expected: generate and store a nonce, construct the message, call the wallet's sign-message RPC, verify server-side, bind to a session.

**Wallet side** is small but non-optional. No new RPC method is needed. The wallet must detect a SIWS-shaped payload inside the existing sign-message call, render a structured consent screen, and match the message's `domain` field against the actual request origin. That origin check is the entire security value over a raw message prompt and can only live in the wallet.

**Suggested basis: SIP-030, not legacy Connect.** Concretely: specify transport as `stx_signMessage` with `messageType: "utf8"`; state that `stx_getAddresses` supplies the address before message construction; drop any dependence on the legacy `@stacks/auth` JWT flow (janniks, 2024).

**Remaining gaps**
1. No SIP-030 method binding is described.
2. Verification does not specify deriving the Stacks address from the returned `publicKey` and comparing it to the `address` field — without this a wallet can return a signature from a different key than the one claimed.
3. The 80-character URI and 10-resource limits are wallet display constraints in a wire format; move them to a SHOULD in wallet UX guidance.
4. Presenting a Bitcoin address alongside the Stacks address in a single flow is unspecified, which is the practical blocker for dual-chain wallets.

I believe the Leather tracking issue (leather-io/mono#1832) was closed unintentionally; I will check with the team about reopening it.

---

**References**
<small>
better-auth. (2025). *SIWX* (Issue No. 6738). GitHub. https://github.com/better-auth/better-auth/issues/6738
Chang, W., Rocco, G., Millegan, B., Johnson, N., & Terbu, O. (2021). *ERC-4361: Sign-In with Ethereum* (Ethereum Improvement Proposals No. 4361). https://eips.ethereum.org/EIPS/eip-4361
Coinbase Developer Platform. (n.d.). *Sign-In-With-X (SIWX)*. x402 documentation. https://docs.x402.org/extensions/sign-in-with-x
Haardik, & Ukustov, S. (2022). *CAIP-122: Sign in with X (SIWx)* (Chain Agnostic Improvement Proposals No. 122). https://standards.chainagnostic.org/CAIPs/caip-122
Aryzing, Janniks, Jamie, K. & Aboelenein, M. (2024). *SIP-030: Wallet interface*. https://github.com/stacksgov/sips/pull/166
LearnWeb3DAO. (n.d.). *siwx: Sign in with X* [Computer software]. GitHub. https://github.com/LearnWeb3DAO/siwx
Pradel, L. (2023). *Stacks namespace — SIWx* [Draft]. Chain Agnostic Namespaces. https://namespaces.chainagnostic.org/stacks/caip122
</small>

314159265359879:
**What about implementing this in `@stacks/connect`?** @kyranjamie suggested as much in the Leather issue.

Wallets do not strictly need library support, `stx_signMessage` already carries the string. But without a canonical serializer, each wallet writes its own parser against the grammar, and the failure mode is silent divergence: wallet A recognizes the payload and shows the consent screen, wallet B does not and shows a raw prompt. Apps then feature-detect per wallet, which is the fragmentation the standard exists to prevent.

A shared parser and verifier in `@stacks/connect`, consumable by wallets as well as apps, as SIP-030 already anticipates for validation libraries, gives every implementation the same byte-for-byte message. Note the domain-origin comparison itself still has to run inside the wallet against the browser-provided origin; the library can only supply the parsed `domain` term, not a trustworthy origin.

Suggest implementing the CAIP-122 Stacks profile rather than SIWS-specific code. Same effort, and the serializer stays compatible with SIWX adapters.

Issue #77, opened 2022-06-28
First step in submitting a SIP


njordhov:
**The [README](https://github.com/stacksgov/sips/blob/main/README.md) should provide a reasonable first step in the process of submitting a SIP, such as suggesting starting by posting a new issue here with an initial proposal for the SIP.**

[SIP 000](https://github.com/stacksgov/sips/blob/main/sips/sip-000/sip-000-stacks-improvement-proposal-process.md) provides an high-level overview of the SIP process but not a practical first step. The [SIPS README](https://github.com/stacksgov/sips/blob/main/README.md) in the *sips* directory suggests that _New SIPs may be submitted via pull request_. Having to submit a pull request as a first step sets the bar high for engaging with the SIP process, which @GinaAbrams argues in issue #73 already has *a high barrier to entry for folks that don't use GitHub day in and day out.*

Posting a short proposal in the issue tracker will lower the barrier to entry and open for further onboarding and hand-holding (if required).

njordhov:
There should be a process in place to handle incoming SIP proposals on the issue tracker, such as having a [SIP editor](https://github.com/stacksgov/sips/blob/main/sips/sip-000/sip-000-stacks-improvement-proposal-process.md#sip-editor-duties) respond to the proposal within reasonable time.

radicleart:
Do we have a way to consistently tag SIPs with the consideration boards they pertain to or that the author maybe provides as an opener - "technical," "economic," "governance" etc - is it possible e.g. to build a filter for this maybe via an external system?

njordhov:
For brief SIP proposals posted in the issue tracker, the responding editor could use github's tagging system to suggest a classification, which could make it listed as a card for the matching [project board](https://docs.github.com/en/issues/organizing-your-work-with-project-boards/managing-project-boards/about-project-boards).

jcnelson:
> @radicleart Do we have a way to consistently tag SIPs with the consideration boards they pertain to or that the author maybe provides as an opener - "technical," "economic," "governance" etc - is it possible e.g. to build a filter for this maybe via an external system?

Why not just tag them with github tags?  Then, each CAB member can just filter PRs by their respective tag.

njordhov:
Per today's weekly SIP call, a very first step could be announcing the proposed SIP in a new *sip* channel on the Stacks discord messaging system.

PR #78, opened 2022-06-30
Explain first step in the SIP process

njordhov:
Resolves stacksgov/sips#77

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,11 @@
 The SIPs describe the design, implementation, and governance
 of the Stacks 2.0 blockchain.  The SIP process
 ([SIP-000](./sips/sip-000/sip-000-stacks-improvement-proposal-process.md))
-describes how to make a SIP and get it ratified.  Anyone in the Stacks community
-may submit a SIP.
+describes how to make a SIP and get it ratified.  
+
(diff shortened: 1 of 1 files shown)


MarvinJanssen:
@Hero-Gamer seems like a nice little addition. Can we merge?

Hero-Gamer:
I am working with the SIP Editors committee on detailing that 1st step process more extensively of any proposal from the community. 
I do wanna point one thing out, the first step may / may not apply to the bigger SIPs like Stacks 2.1 upgrade or sBTC two way peg. sBTC SIP-021 went to Pull Request first, so it would not follow the "first step in the SIP process" so to speak.

Leaves the question of: Should ALL SIPs include the bigger SIP follow the same 1st step, or is it better to have 2 categories of 1st step? Would love to hear your thoughts. Happy to jump on calls with anyone on this or talk about it on SIP call.

njordhov:
> Should ALL SIPs include the bigger SIP follow the same 1st step

It benefits transparency and consistency to have all new SIPs on the issue tracker rather than buried as pull requests.

BhinbahadurUK:
Yes


To all requests yes go ahead

UK Bhinbahadur

PR #84, opened 2022-09-02
Safe Transfer Standard for SIP-009/SIP-010 Assets

radicleart:
A suggestion for two traits that can be implemented by SIP-009 and SIP-010 contracts going forward which will make them more safely interoperable with Marketplace contracts.


MarvinJanssen:
@radicleart would love to breathe some life into this draft. What do you need to move it forward?

PR #96, opened 2022-10-25
chore: fix typo in CAB bylaws

obycode:


whoabuddy (approved):
Good catch on the bad copy-pasta!

PR #97, opened 2022-10-28
Technical CAB meeting minutes 20221027

obycode:


considerations/minutes/technical-cab/2022-10-27.md
```diff
@@ -0,0 +1,88 @@
...
+      availability for the group
+    - Additional meetings outside of this regularly scheduled meeting may be
+      required for specific SIPs and will be scheduled ad-hoc
+- SIP-015 discussions
+  - Concerns have been expressed that this SIP is too large and contains
+    unrelated items
+    - Members feel that the CAB should be able to vote on individual, related
+      items, not one kitchen sink SIP
+    - With the timing and current state of this SIP, it would be difficult to
+      make a change like this at this time
```
njordhov, line 74: Vague to say "timing and current state of this SIP" when the reason for the reluctance is that the SIP is already implemented. Change to _As the SIP is already implemented, it would be difficult to make a substantial change like this_.

obycode: Sounds good.

whoabuddy (commented):
LGTM - thanks for following a similar format to the original, as more of these come in I think we can standardize a template for the required fields and suspect others will follow suit with what's here.

Regarding the action items - is this something we want in the minutes? Will it be updated as tasks are completed, or is it just an artifact from the actions discussed in the meeting?

obycode:
> Regarding the action items - is this something we want in the minutes? Will it be updated as tasks are completed, or is it just an artifact from the actions discussed in the meeting?

Good question. I like having action items at the end of the minutes. Would it make sense if we don't update the minutes once it's merged, so these action items act, like you said, as an artifact from the discussion, and a point to review at the next meeting?

whoabuddy:
If it's just an artifact then I think it makes sense to keep in there, my initial thought is we wouldn't want to have a buried to-do list that doesn't get updated.

Now that there are more CABs, more CAB participants, and a planned faster cadence for changes through the SIP process it will be interesting to figure out the best procedures here. 

I'll get this merged in and we'll keep iterating!

PR #99, opened 2022-11-02
Governance CAB minutes for SIP-015

whoabuddy:
This PR adds the minutes from the Governance CAB meeting re: SIP-015 on October 31, 2021. :ghost: :jack_o_lantern: :ghost:

It also makes some slight edits to the Technical CAB notes to bring the formatting in sync. There will likely be another round of this once all meeting minutes are submitted from each CAB. Tagging @obycode for awareness but no action needed.

Based on our action items, we'd like to tag the stacking pool operators @friedger and @yknl to gather their sentiment regarding:

- the transition to continuous stacking and addition of new functions
- the implementation and being ready for users to transition

A quick answer as a comment on this PR would be sufficient. We were also looking to contact PlanBetter but didn't have a good GitHub handle, and will try to relay the message via DIscord. :pray: 

Also, there were two areas around the voting process that were not clear from the SIP:

-  can Stackers utilizing an Exchange product, e.g. Okcoin Earn, participate in the vote?
-  can Exchanges vote as a solo stacker given they stack on behalf of their participants?

The first is likely no, but it is not communicated anywhere in the SIP and may be confusing for users.

The second is likely yes, but if that did happen would the votes be excluded? Should that be communicated in the SIP?

Outside of the questions raised above, the Governance CAB unanimously approves SIP-015.

### DIFF
--- a/considerations/minutes/governance-cab/2021-11-05-sip-012.md
+++ b/considerations/minutes/governance-cab/2021-11-05-sip-012.md
@@ -2,20 +2,25 @@
 
 ## Meeting Information
 
-Location: Online
-Recorded: No
-Date: November 5, 2021
-Time: 4:00pm - 5:00pm UTC
+**Location:** Online
 
-Attendees:
+**Recorded:** No
+
+**Date:** November 5, 2021
+
+**Time:** 4:00pm - 5:00pm UTC
+
+**Attendees:**
 
 - Jason Schrader
 - Harold Davis
 - Juliet Oberding
 
-Topic(s):
+**Topic(s):**
 
 - [SIP-012: Burn Height Selection for a Network Upgrade to Introduce New Cost-Limits](https://github.com/stacksgov/sips/pull/41)
+
+**Materials:** None
 
 ## Meeting Notes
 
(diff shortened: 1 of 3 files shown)


friedger:
* the transition to continuous stacking and addition of new functions
All but one question could be answered by the SIP authors. The open question is tracked here: https://github.com/stacks-network/stacks-blockchain/issues/3360

* the implementation and being ready for users to transition
Not sure what to answer here. We would want to test the new flow on testnet

planbetterHQ:
Thank you @whoabuddy.

Same comments as @friedger here, with one addition:

How will pools be able to test this? Will it be available in testnet first? If so, when and for how long? We need time to prepare our frontends and to be able to test this thoroughly before it goes live.

Thanks

yknl:
As mentioned already above, it is critical that pools are given ample time to implement and test their changes before 2.1 goes live. The testnet should be available for pools at least 30 days in advance.

whoabuddy:
@jcnelson given the feedback above is there a testing plan or something the pool operators can participate in to ensure they have sufficient time to implement and test changes as soon as code is complete? I recall a doc similar to that but couldn't find the link.

whoabuddy:
@Hero-Gamer that was the one, thanks! I've added a [comment to follow up on the discussion](https://github.com/stacks-network/stacks-blockchain/discussions/3242#discussioncomment-4058639) so we can get this PR merged in and finalized.

Tagging @yknl @friedger @planbetterHQ for visibility!

PR #102, opened 2022-11-03
feat: simpler voting rules that work better with respect to the snapshot

jcnelson:
This PR simplifies the voting rules regarding non-Stackers in a way that makes it much easier for EcosystemDAO and web frontends to track vote tallies.

I'm sending this as a PR to make it easier for the governance CAB to review before merging.  Does this text work for y'all?

### DIFF
--- a/sips/sip-015/sip-015-network-upgrade.md
+++ b/sips/sip-015/sip-015-network-upgrade.md
@@ -1629,13 +1629,10 @@ If a user is Stacking, then their STX can be used to vote in one of two ways,
 depending on whether or not they are solo-stacking or stacking through a
 delegate.
 
-The reason voting is streched across two reward cycles is to accommodate users
-whose STX will have unlocked during the voting period.  If a user's STX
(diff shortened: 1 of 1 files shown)


radicleart (commented):
This looks good to me - thanks for the revisions!

whoabuddy:
I've asked the Governance CAB to take a look and we plan to get an answer back to you asap.

My first impression: simpler is better! Once confirmed we'll need to update these examples as well:
https://github.com/stacksgov/sips/blob/fix/simpler-voting-rules/sips/sip-015/sip-015-network-upgrade.md#examples-1

whoabuddy:
The Governance CAB reviewed the new voting methods async, votes below were recorded from our Discord group and everyone is in favor of the new rules!

- Juliet: yes
- Harold : yes
- Zero.btc: yes
- Jason: yes
- Orlando: yes

whoabuddy (approved):
LGTM!

PR #106, opened 2022-11-12
SIP-020 bitwise operators to Clarity 2

obycode:
This SIP proposes new bitwise operators to be added to Clarity2. It would need to activate with 2.1, as it is a consensus breaking change.

Reference implementation: https://github.com/stacks-network/stacks-blockchain/pull/3389 (in progress)

This change should not hold up code complete for 2.1. If the PR is not approved in time, then it should be saved for the next hard-fork.

In addition to sign off from the Technical CAB for this change, the SIP will also require approval from the Governance CAB to approve the activation strategy.

### DIFF
--- a/sips/sip-xyz/sip-xyz-bitwise-ops.md
+++ b/sips/sip-xyz/sip-xyz-bitwise-ops.md
@@ -119,8 +119,11 @@ inputs.
 - **Inputs:** int | uint
 - **Output:** int | uint
 
-Returns the result of bitwise not, effectively reversing the bits of `i1` (1's
-complement).
+Returns the one's compliment (sometimes also called the bitwise compliment or
+not operator) of `i1`, effectively reversing the bits in `i1`.
+
+In other words, every bit that is `1` in `ì1` will be `0` in the result.
+Conversely, every bit that is `0` in `i1` will be `1` in the result.
 
 ### Examples
 
@@ -195,3 +198,5 @@ This SIP will be a rider on SIP-015. It will be considered activated if SIP-015
 # Reference Implementations
 
 - https://github.com/stacks-network/stacks-blockchain/pull/3389
+- See also discussions in
+  https://github.com/stacks-network/stacks-blockchain/pull/3382
--- a/sips/sip-xyz/sip-xyz-bitwise-ops.md
+++ b/sips/sip-xyz/sip-xyz-bitwise-ops.md
@@ -51,7 +51,8 @@ bit field for example, would be much more difficult to implement without the use
 of these operations. When executing a contract using these operations, the
 common hardware on which miners and nodes are likely to be running can all
 perform these operations very efficiently -- these are typically single cycle
-operations.
+operations. Note that the addition of these bitwise operations commits the
+Clarity VM to using 2's complement to represent integers.
 
 # Specification
 
@@ -144,6 +145,8 @@ Conversely, every bit that is `0` in `i1` will be `1` in the result.
 
 Shifts all bits in `i1` to the left by the number of places specified in `shamt`
 modulo 128 (the bit width of Clarity integers). New bits are filled with zeros.
+When `i1` is a `uint` (unsigned), this is a logical shift. When `i1` is an `int`
+(signed), this is an arithmetic shift, preserving the sign.
 
 ### Examples
 
@@ -193,8 +196,8 @@ this SIP is activated, and valid after it is activated.
 
 # Activation
 
-This SIP will be a rider on SIP-015. It will be considered activated if SIP-015
-(and Stacks 2.1) is activated.
+This SIP will be a rider on SIP-015. It will be considered activated if and only
+if SIP-015 (and Stacks 2.1) is activated.
 
 # Reference Implementations
 
--- a/sips/sip-xyz/sip-xyz-bitwise-ops.md
+++ b/sips/sip-xyz/sip-xyz-bitwise-ops.md
@@ -46,11 +46,12 @@ by the Stacks Open Internet Foundation.
 # Introduction
 
 Bitwise operations are common in other programming languages. Common algorithms,
-including many used in encryption for example, would be much more difficult to
-implement without the use of these operations. When executing a contract using
-these operations, the common hardware on which miners and nodes are likely to be
-running can all perform these operations very efficiently -- these are typically
-single cycle operations.
+including many used in encryption, or the ability to set and check flags in a
+bit field for example, would be much more difficult to implement without the use
+of these operations. When executing a contract using these operations, the
+common hardware on which miners and nodes are likely to be running can all
+perform these operations very efficiently -- these are typically single cycle
+operations.
 
 # Specification
 
--- a/sips/sip-xyz/sip-xyz-bitwise-ops.md
+++ b/sips/sip-xyz/sip-xyz-bitwise-ops.md
@@ -145,18 +145,21 @@ Conversely, every bit that is `0` in `i1` will be `1` in the result.
 
 Shifts all bits in `i1` to the left by the number of places specified in `shamt`
 modulo 128 (the bit width of Clarity integers). New bits are filled with zeros.
-When `i1` is a `uint` (unsigned), this is a logical shift. When `i1` is an `int`
-(signed), this is an arithmetic shift, preserving the sign.
+
+Note that there is a deliberate choice made to ignore arithmetic overflow for
+this operation. In use cases where overflow should be detected, developers
+should use `*`, `/`, and `pow` instead of the shift operators.
 
 ### Examples
 
 ```
-(<< 2 u1) ;; Returns 4
 (<< 16 u2) ;; Returns 64
 (<< -64 u1) ;; Returns -128
 (<< u4 u2) ;; Returns u16
-(<< u240282366920938463463374607431768211327 u3) ;; Returns 30756142965880123323311949751266331049856
-(<< u123 u24028236699) ;; Returns 16508780544 (123 << 27)
+(<< 123 u9999999999) ;; Returns -170141183460469231731687303715884105728 (== 123 << 127)
+(<< u123 u9999999999) ;; Returns u170141183460469231731687303715884105728 (== u123 << 127)
+(<< -1 u7) ;; Returns -128
+(<< -1 u128) ;; Returns -1
 ```
 
 ## Bitwise Right Shift (`>>`)
@@ -172,6 +175,10 @@ Shifts all the bits in `i1` to the right by the number of places specified in
 sign is preserved, meaning that new bits are filled with the value of the
 previous sign-bit.
 
+Note that there is a deliberate choice made to ignore arithmetic overflow for
+this operation. In use cases where overflow should be detected, developers
+should use `*`, `/`, and `pow` instead of the shift operators.
+
 ### Examples
 
 ```
@@ -179,9 +186,12 @@ previous sign-bit.
 (>> 128 u2) ;; Returns 32
 (>> -64 u1) ;; Returns -32
 (>> u128 u2) ;; Returns u32
-(>> u240282366920938463463374607431768211327 u127) ;; Returns 0
-(>> u123 u2402820) ;; Returns 7 (u123 >> 4)
-(>> -3 u12) ;; Returns -1
+(>> 123 u9999999999) ;; Returns 0
+(>> u123 u9999999999) ;; Returns u0
+(>> -128 u7) ;; Returns -1
+(>> -256 u1) ;; Returns -128
+(>> 5 u2) ;; Returns 1
+(>> -5 u2) ;; Returns -2
 ```
 
 # Related work
--- a/sips/sip-xyz/sip-xyz-bitwise-ops.md
+++ b/sips/sip-xyz/sip-xyz-bitwise-ops.md
@@ -29,12 +29,12 @@ This SIP adds bitwise operations to the Clarity language which could simplify
 the implementation of smart contracts that require manipulation of bits. The
 changes include the addition of the following operations:
 
-- Bitwise Xor (`^`)
-- Bitwise And (`&`)
-- Bitwise Or (`|`)
-- Bitwise Not (`~`)
-- Binary Left Shift (`<<`)
-- Binary Right Shift (`>>`)
+- Bitwise Xor (`bit-xor`)
+- Bitwise And (`bit-and`)
+- Bitwise Or (`bit-or`)
+- Bitwise Not (`bit-not`)
+- Binary Left Shift (`bit-shift-left`)
+- Binary Right Shift (`bit-shift-right`)
 
 # License and Copyright
 
@@ -56,9 +56,9 @@ Clarity VM to using 2's complement to represent integers.
 
 # Specification
 
-## Bitwise Xor (`^`)
+## Bitwise Xor (`bit-xor`)
 
-`(^ i1 i2...)`
+`(bit-xor i1 i2...)`
 
 - **Inputs:** int, ... | uint, ...
 - **Output:** int | uint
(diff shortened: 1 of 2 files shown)


sips/sip-xyz/sip-xyz-bitwise-ops.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+## Bitwise Not (`~`)
+
+`(~ i1)`
+
+- **Inputs:** int | uint
+- **Output:** int | uint
+
+Returns the result of bitwise not, effectively reversing the bits of `i1` (1's
```
cylewitruk, line 122: I updated the description of this operator:

```Returns the one's compliment (sometimes also called the bitwise compliment or not operator) of `i1`, effectively reversing the bits in `i1`.
In other words, every bit that is `1` in ì1` will be `0` in the result.  Conversely, every bit that is `0` in `i1` will be `1` in the result.```

File: sips/sip-xyz/sip-xyz-bitwise-ops.md [hunk 2]
```diff
@@ -0,0 +1,195 @@
...
+### Examples
+
+```
+(~ 3) ;; Returns -4
+(~ u128) ;; Returns u340282366920938463463374607431768211327
+(~ 128) ;; Returns -129
+(~ -128) ;; Returns 127
+```
+
+## Bitwise Left Shift (`<<`)
```
cylewitruk, line 134: There is a pending discussion on the behavior of this function.

sips/sip-xyz/sip-xyz-bitwise-ops.md
```diff
@@ -0,0 +1,195 @@
...
+```
+(<< 2 u1) ;; Returns 4
+(<< 16 u2) ;; Returns 64
+(<< -64 u1) ;; Returns -128
+(<< u4 u2) ;; Returns u16
+(<< u240282366920938463463374607431768211327 u3) ;; Returns 30756142965880123323311949751266331049856
+(<< u123 u24028236699) ;; Returns 0
+```
+
+## Bitwise Right Shift (`>>`)
```
cylewitruk, line 155: There is a pending discussion on the behavior of this function.

sips/sip-xyz/sip-xyz-bitwise-ops.md
```diff
@@ -0,0 +1,195 @@
...
+this SIP is activated, and valid after it is activated.
+
+# Activation
+
+This SIP will be a rider on SIP-015. It will be considered activated if SIP-015
+(and Stacks 2.1) is activated.
+
+# Reference Implementations
+
+- https://github.com/stacks-network/stacks-blockchain/pull/3389
```
cylewitruk, line 195: Maybe also worth referencing PR #3382 since that's where most of the discussions have taken place?

sips/sip-xyz/sip-xyz-bitwise-ops.md
```diff
@@ -0,0 +1,195 @@
...
+# License and Copyright
+
+This SIP is made available under the terms of the Creative Commons CC0 1.0
+Universal license, available at
+https://creativecommons.org/publicdomain/zero/1.0/ This SIP's copyright is held
+by the Stacks Open Internet Foundation.
+
+# Introduction
+
+Bitwise operations are common in other programming languages. Common algorithms,
```
cylewitruk, line 48: My primary use case was simply being able to use `flags`, and easily setting/determining which flag was set.  Maybe also worth mentioning this here.

obycode: Added in 1a19399b649aa81f81ada6345a2d659c0e3d0dcf

sips/sip-xyz/sip-xyz-bitwise-ops.md
obycode: Added in 1bdfb06c0f38beacdcadf8c1221027f379b4adb4

sips/sip-xyz/sip-xyz-bitwise-ops.md
obycode: Added in 1bdfb06c0f38beacdcadf8c1221027f379b4adb4

sips/sip-xyz/sip-xyz-bitwise-ops.md
obycode: Updated in 78edc78a32bab0722d32c899e90208b1c86ae1bd based on our discussion.

sips/sip-xyz/sip-xyz-bitwise-ops.md
obycode: Updated in 78edc78a32bab0722d32c899e90208b1c86ae1bd based on our discussion.

sips/sip-xyz/sip-xyz-bitwise-ops.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Bitwise Xor (`^`)
+
+`(^ i1 i2...)`
+
+- **Inputs:** int, ... | uint, ...
+- **Output:** int | uint
+
+Returns the result of bitwise exclusive or'ing a variable number of integer
+inputs.
```
jcnelson, line 66: It might be worthwhile to explain somewhere that the VM now commits to using 2's complement to represent negative integers.  This has a direct impact on how bitwise operations act on negative integers.

sips/sip-xyz/sip-xyz-bitwise-ops.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Bitwise Left Shift (`<<`)
+
+`(<< i1 shamt)`
+
+- **Inputs:** int, uint | uint, uint
+- **Outputs:** int | uint
+
+Shifts all bits in `i1` to the left by the number of places specified in `shamt`
+modulo 128 (the bit width of Clarity integers). New bits are filled with zeros.
```
jcnelson, line 146: > `shamt` modulo 128

Do you mean `min(shamt, 128)`?  `(<< i1 shamt)` for any `shamt >= 128` ought to evaluate to 0.

jcnelson: Same question below for `>>`

jcnelson: Also, what is the behavior of `<<` for negative integers?  Is the sign bit preserved? I think it should be -- a shift on a negative integer in either direction ought to be an arithmetic shift, and a shift on an unsigned integer ought to be a logical shift.

sips/sip-xyz/sip-xyz-bitwise-ops.md
```diff
@@ -0,0 +1,203 @@
...
+
+# Backwards Compatibility
+
+Because this SIP introduces new Clarity operators, it is a consensus-breaking
+change. A contract that uses one of these new operators would be invalid before
+this SIP is activated, and valid after it is activated.
+
+# Activation
+
+This SIP will be a rider on SIP-015. It will be considered activated if SIP-015
```
jcnelson, line 196: Minor nit -- "It will be considered activated if **and only if** SIP-015 is activated".  This SIP should not active if SIP 015 fails to activate.

jcnelson (commented):
Mostly looks good.  Just had a couple minor points.  Once they're addressed, please feel free to transition this to `Accepted` status and call it SIP 020.

sips/sip-xyz/sip-xyz-bitwise-ops.md
obycode: > > `shamt` modulo 128
> 
> 
> 
> Do you mean `min(shamt, 128)`?  `(<< i1 shamt)` for any `shamt >= 128` ought to evaluate to 0.

Cyle and I discussed this and realized that the common behavior from other languages, which should be the least surprising behavior, is to mask the shift amount, or modulo by the bit width. Cyle is going to make the corresponding change in the implementation.

obycode: > Also, what is the behavior of `<<` for negative integers?  Is the sign bit preserved? I think it should be -- a shift on a negative integer in either direction ought to be an arithmetic shift, and a shift on an unsigned integer ought to be a logical shift.

Good point. I agree and will make that change tomorrow.

cylewitruk: > Also, what is the behavior of `<<` for negative integers?  Is the sign bit preserved? I think it should be -- a shift on a negative integer in either direction ought to be an arithmetic shift, and a shift on an unsigned integer ought to be a logical shift.

The current implemented behavior preserves the sign bit for negative integers in both shift operations:
```        
Ok(Value::Int(-128)),   // (<< -1 u7)
Ok(Value::Int(-1)),       // (<< -1 u128)
Ok(Value::Int(-1)),       // (>> -128 u7)
Ok(Value::Int(-128)),   // (>> -256 64)
Ok(Value::Int(1)),        // (>> 5 u2)
Ok(Value::Int(-2)),      // (>> -5 u2)
```

sips/sip-xyz/sip-xyz-bitwise-ops.md
obycode: Added a note in 166ae07

sips/sip-xyz/sip-xyz-bitwise-ops.md
obycode: Added in 166ae07.

obycode:
@lnow Members of the Technical CAB have requested your input on this SIP.

LNow:
@obycode I'm sorry, but I don't have time for this.

wileyj:
@lgalabru would you have some time to add some input here?

lgalabru:
LGTM, but I'm a bit surprised / ambivalent on the shift behavior approach, and the "wrapping" strategy (`<< 129` == `<< 1`).
Could it become a foot gun for developers?
It does not look like there's a consensus between languages on how to address that, but many behavior can be observed in the nature (credit to @obycode for the quick research): 
- JS: wrapping
- Swift: erroring
- Rust: panicking

If we wanted to be consistent with the overflow/underflow runtime errors that we took for the `math` operators, we would throw a runtime error, no?

jcnelson:
> If we wanted to be consistent with the overflow/underflow runtime errors that we took for the math operators, we would throw a runtime error, no?

I strongly believe that we do _NOT_ want this.  If the developer wants runtime errors for bit-shifting, then they can use `*`, `/`, and/or `pow`.  Therefore, shifting ought _NOT_ to throw errors so developers can choose which semantics they want.

> LGTM, but I'm a bit surprised / ambivalent on the shift behavior approach, and the "wrapping" strategy (<< 129 == << 1).
Could it become a foot gun for developers?

Different microarchitectures handle it differently.  I think it's undefined behavior in C, and it's definitely undefined in C++11.  I'm fine with wrapping because it doesn't throw errors.

obycode:
I'm comfortable with the masking behavior currently described in this SIP, or the behavior I originally had, where a shift amount > 128 will just clear all the bits and result in a `0` or `-1`, but I don't like the error behavior, because in typical uses of these bitwise operations, you'll be expecting to drop bits. As Jude said, if you want the error, you can use `*`, `/`, and `pow`.

wileyj:
> > If we wanted to be consistent with the overflow/underflow runtime errors that we took for the math operators, we would throw a runtime error, no?
> 
> I strongly believe that we do _NOT_ want this. If the developer wants runtime errors for bit-shifting, then they can use `*`, `/`, and/or `pow`. Therefore, shifting ought _NOT_ to throw errors so developers can choose which semantics they want.
> 
> > LGTM, but I'm a bit surprised / ambivalent on the shift behavior approach, and the "wrapping" strategy (<< 129 == << 1).
> > Could it become a foot gun for developers?
> 
> Different microarchitectures handle it differently. I think it's undefined behavior in C, and it's definitely undefined in C++11. I'm fine with wrapping because it doesn't throw errors.

If this is something we *don't* want to include in this SIP...could it open the door to some unintended consequences (like the ones @lgalabru mentioned)?

jcnelson:
> If this is something we don't want to include in this SIP...could it open the door to some unintended consequences (like the ones @lgalabru mentioned)?

Wrapping the shift operand is supported on all architectures, because architectures only start to exhibit undefined behavior when the shift amount exceeds the bit width of the integer being shifted.  By explicitly masking the lower 128 bits of the shift argument, we avoid this altogether.

sips/sip-xyz/sip-xyz-bitwise-ops.md
```diff
@@ -0,0 +1,206 @@
...
+## Bitwise Left Shift (`<<`)
+
+`(<< i1 shamt)`
+
+- **Inputs:** int, uint | uint, uint
+- **Outputs:** int | uint
+
+Shifts all bits in `i1` to the left by the number of places specified in `shamt`
+modulo 128 (the bit width of Clarity integers). New bits are filled with zeros.
+When `i1` is a `uint` (unsigned), this is a logical shift. When `i1` is an `int`
```
jcnelson, line 148: On further testing, this text is wrong.  The sign bit is _not_ preserved for `int`, and that's also consistent with an arithmetic left shift.

Given that we aren't doing any arithmetic overflows here, it might be prudent to mention up front that this is a deliberate choice.  If the developer wants arithmetic errors, they can use `*`, `/`, and `pow`.

obycode: > On further testing, this text is wrong. The sign bit is not preserved for int, and that's also consistent with an arithmetic left shift.
Yes, I noticed this too. Will update.

obycode: Updated in [a2cf4f6](https://github.com/stacksgov/sips/pull/106/commits/a2cf4f6bdb86da855039a7fc87f4c56b08c0a43f)

jcnelson:
I see the symbols have changed.  Is that the feedback from the technical CAB?

obycode:
> I see the symbols have changed. Is that the feedback from the technical CAB?

Exactly.

LNow:
I'm finally off work, so here is my feedback:

Current description of each function might be difficult to understand by new developers who never had a chance to code in other language.
In my experience definitions provided by Microsoft are very clean and easy to digest. I think you could re-use them. 
Here is example for `<<` https://learn.microsoft.com/en-us/cpp/cpp/left-shift-and-right-shift-operators-input-and-output?view=msvc-170#left-shifts

Bitwise logic (and, or, xor) are easier to understand when you present table that shows two inputs and output, and right after that simple example with binary representation of each value used in such example.

#### AND `&`
|Input A|Input B|Output|
|-|-|-|
|1|1|1|
|0|1|0|
|0|1|0|
|0|0|0| 

```clarity
(& 17 30) ;; Returns 16
;; Input A (17):  0000000000010001
;; Input B (30):  0000000000011110
;; Output  (16):  0000000000010000 
```

Bit shifting examples should also have binary representation that shows exactly what is going on.
```clarity
(<< 17 u2)  ;; Returns 68
;; Input (17):  0000000000010001
;; Output (68): 0000000001000100
```
And also it might be worth to mention (and show) that order in which we specify arguments in `|, & and ^` doesn't matter.
```clarity
(| 64 -32 -16) ;; Returns -16
(| 64 -16 -32) ;; Returns -16
(| -32 64 -16) ;; Returns -16
(| -32 -16 64) ;; Returns -16
(| -16 64 -32) ;; Returns -16
(| -16 -32 64) ;; Returns -16
```

It might be difficult to show in a readable way 128bit numbers, therefore examples in which you decide to show binary representation should have information that it has been simplified and shortened to X bits.

LNow:
If you want to activate this together with SIP-015, then I would advice to add a proper note to voting website (https://sip015.xyz)

cc: @whoabuddy @radicleart

radicleart:
@LNow i'll ask @jrmith to update the blog with the info that SIP-020 will activate with on SIP-015.

LNow:
@radicleart this SIP was meant to be a "rider" on SIP-015, so I think its authors and core team didn't want it to be subject of any vote. However I believe they should be transparent and give users full picture of what they want to release with 2.1.

If they decide to put it under a separate vote - great. If not, then adding a note to description of 2.1 voting is the only reasonable way to handle it in a fair way.

obycode:
The situation was that it was a last minute contribution from a community member and it seems like a non-controversial add on, and therefore not worth the effort of a separate vote, but we wanted the Governance CAB and Steering Committee to confirm that. 

It definitely wouldn't hurt to add a note about it in the voting page.

wileyj:
With the additional changes made over the past few days as recommended by the Technical CAB and in the comments here, this comment indicates approval of this SIP after a vote approving this proposal. 

One comment of note regarding this proposal and the urgency to have it voted on for 2.1:
```
The added functionality is neither commonly requested by the community nor a blocker for contract developers. The original proposal has several red flags indicating the addition shouldn't be a priority, admitting that the lack of functionality is not a problem and that the solution just would be nice to have. In the discussion, Jude shares that when it comes to the new functions, you can implement them today with the existing operations. We should expect submitted SIPs to make the case that there is a substantial problem, with use cases demonstrating an actual need for a solution.
```

jcnelson:
@whoabuddy @obycode Has this been signed off on by both your CABs?

whoabuddy:
@jcnelson yes! Catching up the official minutes today but the Governance CAB approves :+1:

obycode:
Yes, see @wileyj's message above. I sat this one out since I was the author and @wileyj took the lead.

jcnelson:
Just an update on this -- we will merge this once 2.1 activates on mainnet.

PR #107, opened 2022-11-23
Add gov cab minutes for sip-020

whoabuddy:
This PR adds the Governance CAB minutes for the decision on SIP-020, with all members in favor :+1:


jcnelson:
Thanks!

PR #109, opened 2022-11-30
fix: updated `?` suffixes

obycode:
This was discussed in the comments, and updated in the implementation, but never made it into the actual SIP.


jcnelson (approved):

jcnelson:
Thanks @obycode!

PR #58, opened 2022-01-05
Update README.md

moodmosaic:
Fix small typos

PR #127, opened 2023-04-20
SIP-022: Emergency Fix to PoX Stacking Increase

jcnelson:
This is a SIP for addressing the bug in `stacks-increase` that causes the PoX contract to become out-of-sync with the true number of locked STX (details here: https://forum.stacks.org/t/a-bug-in-stacks-increase-call-is-impacting-stacking-rewards-this-cycle/14867).


sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,235 @@
...
+```
+
+The `RawRewardSetEntry` data are pulled directly from the
+`reward-cycle-pox-address-list` data map, and the `.amount_stacked` field is
+equal to the `total-ustx` field that was erroneously set in `stack-increase`.
+
+Considering these consequences with respect to the [blockchain catastrophic
+failure and recovery guidelines](https://github.com/stacksgov/sips/pull/10),
+this bug would be categorized as requiring an _immediate_ hard fork to rectify.
+THe network has not yet crashed, but it is in imminent danger of crashing and
```
rafaelcr, line 136:
```suggestion
The network has not yet crashed, but it is in imminent danger of crashing and
```

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,235 @@
...
+response](https://bitcointalk.org/index.php?topic=823.0) was to quickly release
+a patched version of Bitcoin and rally enough miners and users to upgrade.  In a
+matter of hours, the canonical Bitcoin chain ceased to include any transactions
+that minted too much BTC.
+
+# Backwards Compatibility
+
+The activation of this SIP does _not_ mean that the ability to increase one's locked STX is 
+disabled.  The `delegate-stack-increase` function is not affected by this bug.
+Solo stackers who want to contineu increasing their locked STX will need to
```
rafaelcr, line 203:
```suggestion
Solo stackers who want to continue increasing their locked STX will need to
```

rafaelcr (commented):
SIP Editor review LGTM, just a couple typos.
Let's move this to Accepted for CAB review with the proposed number SIP-022.

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,235 @@
...
+auto-unlocks (which mutates state in `reward-cycle-pox-address-list`).
+
+  The routine for resetting `reward-cycle-pox-address-list` shall be invoked while
+  processing every Stacks block whose parent Stacks block was mined in a Bitcoin
+  block before the activation height.  Given that the set of solo stackers in
+  cycle #56 is small and fixed, the additional overhead of fixing all records in
+  `reward-cycle-pox-address-list` will be minimal.  Nevertheless, as a safety
+  precaution, each Stacks block whose parent was mined prior to the Bitcoin
+  block activation height must contain zero transactions and confirm zero
+  microblocks.
```
friedger, line 172: Any transaction? No activity until the block activation?

314159265359879: I read this as _no pox-2 transactions_, good question. Perhaps it can be made more explicit.

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,235 @@
...
+after the Bitcoin block activation height passes.  In addition, Stacks 2.2 nodes
+will ignore block-commits from Stacks 2.1 nodes.  Similar changes were made for
+Stacks 2.05 and Stacks 2.1 to ensure that the new network cleanly separates from
+stragglers still following the old rules.
+
+# Activation
+
+The Bitcoin block activation height will need to pass prior to the selection of
+the PoX anchor block for reward cycle #57 (Bitcoin block height 785751).  This
+SIP proposes Bitcoin block height 785551, which is 200 Bitcoin blocks prior.  In
```
obycode, line 221: I think this should be reward cycle 58 and bitcoin blocks 787751 and 787551, respectively.

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,235 @@
...
+start of the next reward cycle at the time of this writing.
+There is less than one reward cycle remaining to fix this
+problem, and yet a Stacker vote would require at least one complete reward cycle
+to carry out a STX coin vote (not accounting for any additional time required for
+sending out adequate public communications and tabulating the vote afterwards).
+
+# Specification
+
+Given the lack of time to conduct a fair, public vote to activate this SIP, the
+proposed fix in this SIP is as parsimonious and discreet as possible.  It will:
```
diwakergupta, line 148: There's nothing discreet about the fix :)

sips/sip-022/sip-022-emergency-pox-fix.md
diwakergupta: +1 good to clarify -- I believe this is only relevant to the block in which the epoch activates.

sips/sip-022/sip-022-emergency-pox-fix.md
diwakergupta: Earlier there's a reference to 785,550 too.

diwakergupta (commented):
Thanks for getting the SIP out so quickly Jude, much appreciated 🙏🏽 Mostly minor comments

Hero-Gamer (commented):
minor editing

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,235 @@
...
+  precaution, each Stacks block whose parent was mined prior to the Bitcoin
+  block activation height must contain zero transactions and confirm zero
+  microblocks.
+
+* Set the minimum required block-commit memo bits to `0x07`.  All block-commits
+  after the Bitcoin block activation height must have a memo value of at least
+`0x07`.  This ensures that miners that do not upgrade from Stacks 2.1 will not
+be able to mine in Stacks 2.2.
+
+* Set the peer network version bits to `0x18000007`.  This ensures that follower
```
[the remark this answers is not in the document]
wileyj: Good catch - I think it would be a good idea to update each network's version bits for uniformity (mocknet included).

AcrossfireX:
SIP Editor chiming in. This looks good from my perspective and feel we should push forward ASAP.

sips/sip-022/sip-022-emergency-pox-fix.md
MarvinJanssen: I read it as all blocks until activation. Perhaps to prevent potentially hitting the liquid Stacks assertion mentioned in the doc?

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,235 @@
...
+
+* Disable `stack-increase` after the Bitcoin block activation height.  This
+  shall be achieved by altering the Clarity VM to declare that this function
+cannot be called, thereby preventing the buggy code body from being
+reached.  This shall not constitute a new version of the Clarity language;
+instead, the Stacks _chainstate specification_ shall be altered to require that
+this particular function in this particular contract after the Bitcoin block
+activation height is treated as absent from all Stacks forks.
+
+* Reset all `total-ustx` values in the `reward-cycle-pox-address-list` that
```
friedger, line 158: Is this considered a change to the _chainstate specification_?
Something like `total-ustx` of `reward-cycle-pox-address-list` shall not be used to calculated the locked amount.

sips/sip-022/sip-022-emergency-pox-fix.md
jcnelson: The first Stacks block mined after the repairs to `reward-cycle-pox-address-list` are executed must contain zero transactions and confirm zero microblocks.  So, activity is allowed until the SIP activation, but the first Stacks block mined afterwards would be empty.

sips/sip-022/sip-022-emergency-pox-fix.md
friedger: Can you clarify in what way this change to the on-chain data is different to auto unlock?

Can we say that this SIP requires that the solo stacker lock amounts are auto unlocked earlier than usually and immediately relocked using a new calculation of the locking amount?

I would like to better understand why this SIP is different to arbitrary on-chain data manipulation (like set the balance of hacker A to 0). 

The current specification is too detailed/technical to understand it.

friedger: Proposal for better understanding:

* Audit the locked amount for solo stackers and the corresponding amounts in the table `reward-cycle-pox-address-list` that is used to determine the number of reward slots. Correct the table entry if the locked amount of the users account (derived from using stack-stx, stack-increase and stack-extend transaction) do not match the stored total amount in the table `reward-cycle-pox-address-list`. The locked amount of the users' accounts shall be the correct value.  
  
  This correction can be seen as an auto-unlock at the time of activation followed by an immediate imaginary stack-stx transaction with the corrected amount. 
  
  This reset shall happen prior to any reward cycle processing in
cycle # 58 -- namely, prior to the call to
`get_reward_threshold_and_participation()` and prior to calculating any STX
auto-unlocks (which mutates state in `reward-cycle-pox-address-list`).
  
   The auditing and correction routine shall be invoked while
  processing every Stacks block whose parent Stacks block was mined in a Bitcoin
  block before the activation height.  Given that the set of solo stackers in
  cycle # 56 and cycle # 57 is small and fixed, the additional overhead of fixing all records in
  `reward-cycle-pox-address-list` will be minimal.  Nevertheless, as a safety
  precaution, each Stacks block whose parent was mined prior to the Bitcoin
  block activation height must contain zero transactions and confirm zero
  microblocks.

friedger: This makes me think whether it is enough to unlock all solo stackers early and make them stack again for cycle 58

friedger:
Referencing for readers of this SIP. Highly likely solution: Disabling PoX for 58 and launching pox-3. See https://forum.stacks.org/t/a-bug-in-stacks-increase-call-is-impacting-stacking-rewards-this-cycle/14867/8

sips/sip-022/sip-022-emergency-pox-fix.md
jcnelson: Yup

jcnelson:
Hey everyone, the text of this SIP has changed substantially.  In line with the public calls we've had about this, this SIP now proposes **two hard forks** -- one to disable PoX in cycle #58, and one to re-enable it via instantiating a fixed `pox-3` contract in cycle #59.  @whoabuddy @obycode I think your respective CABs will need to re-review this before we can advance it to Recommended.

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -130,56 +130,96 @@ The `RawRewardSetEntry` data are pulled directly from the
...
+`reward-cycle-pox-address-list` data map was considered.  This would have only
+required a single hard fork, and no disruption of PoX payouts (however,
+`stack-extend` would have been disabled).  This approach was ultimately dropped
+because of the interactions between delegated stacking and solo stacking -- it
+would not be possible to retroactively compute the correct values for the
+`reward-cycle-pox-address-list` data map.  This, in turn, meant that any repair
+tactic would require the node to calculate the correct values for
+`reward-cycle-pox-address-list` as the blocks in reward cycle #57 were
+processed, which in turn meant that nodes would need to boot from genesis to
+derive the correct data with which to repare the data map at the start of reward
```
wileyj, line 219: `with which to repare` -> `with which to repair`

wileyj (approved):
lgtm

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,279 @@
...
+On 19 April 2023, it was discovered that there was a bug in the PoX smart
+contract which would allow a user to claim that they have stacked far, far more
+STX than they had locked.  Exploiting this bug both allows the user to increase
+their PoX reward slot allotment, while also driving up the stacking minimum.
+Furthermore, it creates the conditions for a network-wide crash:  if the total
+amount of STX stacked as reported by the PoX smart contract were to ever exceed
+the total amount of liquid STX, then the node would crash into an irrecoverable
+state.  This bug has already been triggered in the wild.
+
+This SIP proposes **two immediate consensus-breaking changes** that both prevents this bug
```
Hero-Gamer, line 36: both prevent

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,279 @@
...
+  nodes that do not upgrade to Stacks 2.3 will not be able to talk to Stacks
+2.3 nodes.
+
+## Alternative Approaches Considered
+
+It is unusual that this SIP proposes two hard forks in rapid succession without 
+a Stacker-based vote for activating the new rules.  The
+reason for this is because PoX must first be disabled prior to the start of
+reward cycle #58 in order to avert a network-wide
+crash.  This cannot be delayed, but there is insufficent time to prepare `pox-3`
```
Hero-Gamer, line 196: is insufficient

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,279 @@
...
+
+* The total STX locked to the PoX address owned by caller of `stack-increase`
+  will be potentially higher than the amount of STX that the caller possesses.
+The caller will receive PoX payouts _as if_ they had locked that much STX.  So,
+the caller receives undue BTC.
+
+* The stacking threshold is raised to account for the PoX contract's reported
+  increase in STX locked.
+
+Furthermore, if this bug is triggered enough times, the Stacks network will crash.  This is
```
friedger, line 107: Before this, add a short paragraph about stackers contributing to two reward addresses:

Similar consequences are expected when a stacker contributes to two different PoX addresses and at least one PoX address would benefit from auto-unlocking. This behaviour has been not seen in the wild.

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,279 @@
...
+* Set the mainnet peer network version bits to `0x18000007`.  This ensures that follower
+  nodes that do not upgrade to Stacks 2.2 will not be able to talk to Stacks
+2.2 nodes.
+
+The second hard fork will deactivate `pox-2` and instantiate a new PoX
+implementation `pox-3`.  The `pox-3` contract code will:
+
+* Fix the aforementioned `stacks-increase` bug.
+
+* Add a `delegated-to` field in the `stacking-state` data map, so that the
```
friedger, line 174: These two `delegated-to` points can be summarized into one for better understanding.

* Prevent stackers from contributing to two or more PoX addresses. In particular, 
    * add a `delegated-to` field in the `stacking-state` data map, so that the `stacking-state` entry for a PoX user will indicate the principal to which
their STX are delegated (if they are using delegated stacking). 
    * add checks to the delegated stacking public functions that validate the
  `delegated-to` field in the Stacker's `stacking-state` map entry.

sips/sip-022/sip-022-emergency-pox-fix.md
wileyj: this is grammatically correct, no change is needed here

friedger (commented):
I have added a motivation why we want to add `delegate-to` field to the map and group the two requirements together.

sips/sip-022/sip-022-emergency-pox-fix.md
wileyj:
```
add a delegated-to field (with checks to validate the data) in the stacking-state data map, so that the stacking-state entry for a PoX user will indicate the principal to which their STX are delegated (if they are using delegated stacking).
```
?

sips/sip-022/sip-022-emergency-pox-fix.md
wileyj: > Similar consequences are expected when a stacker contributes to two different PoX addresses and at least one PoX address would benefit from auto-unlocking. This behaviour has been not seen in the wild. 
Furthermore, if this bug is triggered enough times, the Stacks network will crash.  This is
because of a runtime assertion in the PoX reward set calculation logic that
verifies that the total locked STX does not exceed the total liquid STX.  This
would no longer be true due to this bug.  The offending assertion is detailed
below:


like this? i don't see an issue adding this but i'd prefer it @pavitthrap gives the OK to this change

sips/sip-022/sip-022-emergency-pox-fix.md
friedger: What is not clear?

wileyj: maybe i wasn't clear  in what i wrote - what i replied with was a proposed change to address your comment.

MarvinJanssen (approved):
LGTM

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,284 @@
...
+The node software for Stacks 2.2 and 2.3 shall be merged to the `master` branch of the
+reference implementation no later than three days prior to the activation
+height.  This means that everyone shall have at least three days to upgrade
+their Stacks 2.1 nodes to Stacks 2.2, and at least three days to upgrade from
+2.2 to 2.3.
+
+# Reference Implementation
+
+The reference implementation of this SIP can be found in the
+`feat/address-pox-increase` branch of
```
friedger, line 282: branch for first hard fork is `feat/sip-022-pox-disable` (
https://github.com/stacks-network/stacks-blockchain/pull/3677)

No branch for second hard fork yet.

friedger (commented):
From the reference implementation I understand that the first hard fork disables pox-2 already

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,284 @@
...
+equal to the `total-ustx` field that was erroneously set in `stack-increase`.
+
+Considering these consequences with respect to the [draft blockchain catastrophic
+failure and recovery guidelines](https://github.com/stacksgov/sips/pull/10),
+this bug would be categorized as requiring an _immediate_ hard fork to rectify.
+The network has not yet crashed, but it is in imminent danger of crashing and
+there is insufficient time to execute a Stacker-based SIP vote as has been customary for
+past hard forks.  This SIP instead proposes that this hard fork activate at the
+start of the next reward cycle (#58) at the time of this writing.
+
```
njordhov, line 144: In lieu of a community vote, perhaps it would be appropriate to summon the Stack Foundation board as a stand-in.

wileyj: that's an interesting idea - would the various CAB's voting on this SIP also suffice here vs a board vote?

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,284 @@
...
+* Set the minimum required block-commit memo bits to `0x07`.  All block-commits
+  after the Bitcoin block activation height must have a memo value of at least
+`0x07`.  This ensures that miners that do not upgrade from Stacks 2.1 will not
+be able to mine in Stacks 2.2.
+
+* Set the mainnet peer network version bits to `0x18000007`.  This ensures that follower
+  nodes that do not upgrade to Stacks 2.2 will not be able to talk to Stacks
+2.2 nodes.
+
+The second hard fork will deactivate `pox-2` and instantiate a new PoX
```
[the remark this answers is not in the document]
wileyj: I think the wording is correct. 
The first hard fork will *disable* pox-2, but the second hard fork will deactivate it.

sips/sip-022/sip-022-emergency-pox-fix.md
Hero-Gamer: * Stacks Foundation boards is only for the Foundation entity, not boards for the rest of ecosystem is how I read it.
* If we want something like that would be more appropriate to have ecosystem "decentralized boards", which is kind of what Steering Committee is acting as too? Would cross-over responsibility if that's the idea. Option is to have more members on this SC committee. 
* Currently: Requires A. SIP Editor approval to "Accept", CAB votes approval to "Recommend", Steering Committee approval to become "Activation-in-Progress" 
as per SIP-000: https://github.com/stacksgov/sips/blob/main/sips/sip-000/sip-000-stacks-improvement-proposal-process.md

sips/sip-022/.sip-022-emergency-pox-fix.md.swp
obycode, line 1: Now we know Jude uses `vi` 😄.

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,284 @@
...
+A subsequent hard fork to re-enable PoX would have the effect of restoring
+Stacker voting.
+
+# Specification
+
+Given the lack of time to conduct a Stacker vote to activate this SIP, the
+proposed fix in this SIP is as parsimonious and discreet as possible.  It will
+execute as a set of **two hard forks**.
+
+The first hard fork, which activates at the start of reward cycle #58 will disable PoX.
```
obycode, line 158: Should this say "before the start of reward cycle #58"?

wileyj: I think the wording is correct, since it will only activate when the cycle starts, but node operators will need to upgrade before the start of cycle 58

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,284 @@
...
+This hard fork would do the following:
+
+* Prevent all stacking functions from being called in `pox-2`.  The `pox-2`
```
obycode, line 161: Is this wording correct? I think the functions can still be called, however they will not take any effect. Maybe better wording would be:

> Make all stacking functions in the `pox-2` contract ineffective in affecting the locked/unlocked account balances, like the `pox` contract is now.

sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -0,0 +1,284 @@
...
+before `pox-3` would activate, there would not exist a means of carrying out a
+Stacker-based vote to upgrade the system (as had been the case in all prior
+hard forks).  Beyond receiving a BTC yield, **the lack of a functioning PoX contract
+prevents consensus-level SIPs from being voted upon**, which necessitates an
+expedient restoration.
+
+A less-disruptive alternative approach of repairing the
+`reward-cycle-pox-address-list` data map was considered.  This would have only
+required a single hard fork, and no disruption of PoX payouts (however,
+`stack-extend` would have been disabled).  This approach was ultimately dropped
```
obycode, line 217:
```suggestion
`stack-increase` would have been disabled).  This approach was ultimately dropped
```

sips/sip-022/sip-022-emergency-pox-fix.md
obycode: In the activation section it says:

> The Bitcoin block activation height for the first hard fork will need to pass prior to the selection of the PoX anchor block for reward cycle #58 (Bitcoin block height 787851). This SIP proposes Bitcoin block height 787651, which is 200 Bitcoin blocks prior.

sips/sip-022/sip-022-emergency-pox-fix.md
wileyj: i think it's correct as written , since pox will be disabled at the start of cycle 58. 
changing it to the proposed may introduce confusion

obycode: I think the current wording will cause confusion, because the functions can still be called, they just won't do anything useful. The wording here makes me think that I will get an error if I try to call them, which is not correct.

sips/sip-022/sip-022-emergency-pox-fix.md
wileyj: I see your point, but i still feel the current wording is correct. 
the only alternative i can see working here is instead of 
>The first hard fork, which activates at the start of reward cycle #58 will disable PoX.
the bitcoin block where this fork is activated is mentioned

sips/sip-022/sip-022-emergency-pox-fix.md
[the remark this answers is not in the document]
jcnelson: @friedger is correct.  The first hard-fork will declare `pox-2` to be defunct, just like `pox-1` is.  The second hard-fork instantiates `pox-3`.  We're doing two hard forks because we don't have time to do one.

sips/sip-022/sip-022-emergency-pox-fix.md
obycode: How about:

> The first hard fork, which activates before the start of reward cycle #58, will disable PoX beginning with cycle #58.

obycode (approved):
The Technical CAB approves this SIP. Meeting minutes available [here](https://github.com/stacksgov/sips/blob/53b61924f39bb78eae23116b7463fcb27c388bcd/considerations/minutes/technical-cab/2023-04-26-sip-022.md).

whoabuddy:
Posting this for reference - our vote and approval was in before merge!

The Governance CAB approves this SIP. Meeting minutes available in #128.

Sign off PR - #130

jiga:
SIP Editor Update 6/30/2023:
RICE Details: Reach - 500,000, Impact - 3x, Confidence - 95%, Effort - 3 person-months.
Final RICE Score: 475,000

Status: Approved

PR #128, opened 2023-04-20
Add gov cab minutes for SIP-022

whoabuddy:
This PR adds the Governance CAB minutes for the decision on SIP-022, with all members in favor :+1:


whoabuddy:
Noting that as SIP-022 is still evolving in implementation, the Governance CAB is planning to do a secondary review in follow-up to this one once the details are finalized. This is not ready to merge yet.

PR #130, opened 2023-04-26
Add SIP-022 sign-offs and update status

obycode:
Does it make sense for the CAB chairs to open a PR like this upon approval?

### DIFF
--- a/sips/sip-022/sip-022-emergency-pox-fix.md
+++ b/sips/sip-022/sip-022-emergency-pox-fix.md
@@ -21,7 +21,7 @@ Created: 19 April 2023
 
 License: BSD 2-Clause
 
-Sign-off: 
+Sign-off: Brice Dobry <brice@hiro.so> (Technical CAB Chair)
 
 Discussions-To: https://github.com/stacksgov/sips
 
--- a/sips/sip-022/sip-022-emergency-pox-fix.md
+++ b/sips/sip-022/sip-022-emergency-pox-fix.md
@@ -21,7 +21,7 @@ Created: 19 April 2023
 
 License: BSD 2-Clause
 
-Sign-off: Brice Dobry <brice@hiro.so> (Technical CAB Chair)
+Sign-off: Brice Dobry <brice@hiro.so> (Technical CAB Chair), Jason Schrader <jason@joinfreehold.com> (Governance CAB Chair)
 
 Discussions-To: https://github.com/stacksgov/sips
 


sips/sip-022/sip-022-emergency-pox-fix.md
```diff
@@ -21,7 +21,7 @@ Created: 19 April 2023
 
 License: BSD 2-Clause
 
-Sign-off: 
+Sign-off: Brice Dobry <brice@hiro.so> (Technical CAB Chair)
```
whoabuddy, line 24: Might be easier to consolidate our sign-offs in one PR, and this removes the title after the email like we formatted the sign-offs in SIP-015. Looking at other SIPs though, some include a title and some don't, so it'd be good to standardize a format here moving forward.

```suggestion
Sign-off: Brice Dobry <brice@hiro.so>, Jason Schrader <jason@joinfreehold.com>
```

Links for reference:

- #129 
- #128

obycode: It seems useful to include some indication of the CAB that each signer is representing, so that it's there as an artifact.

whoabuddy:
Also would this PR be the best place to change the status to Recommended given the sign-offs?

sips/sip-022/sip-022-emergency-pox-fix.md
obycode: And yes, it does make sense to do it all in one PR so we don't have to conflict with each other. Ideally, I'd like each person to add their own sign off in a commit, so we'd all need permission to open PRs here (instead of in a fork like this is).

whoabuddy: > It seems useful to include some indication of the CAB that each signer is representing, so that it's there as an artifact.

I'm fine with this, generally I look to the last SIP to keep formatting consistent, and if we find a format we like it might be worth amending the old SIPs. We can open a separate issue for that.

> And yes, it does make sense to do it all in one PR so we don't have to conflict with each other. 

So on the CAB side (documenting for later):

- each CAB will create a PR to add meeting notes / vote outcomes
- if approved, each CAB chair will add their sign-off to the SIP in the same PR
- if all CABs sign-off, a CAB chair can update the status to Recommended

If we end up liking this structure we can do something similar for SIP editors to go from Draft -> Accepted.

> Ideally, I'd like each person to add their own sign off in a commit, so we'd all need permission to open PRs here (instead of in a fork like this is).

This makes sense, each CAB should be able to designate at least one person to have access to the org / repo.

Sent you an invite! For now I think I can commit to this PR even though it's on a fork.

whoabuddy:
Linking my earlier reply, in case it gets lost in the resolved convo:
https://github.com/stacksgov/sips/pull/130#discussion_r1179243124

I was able to commit my suggestion and updated it to include the CAB titles per the convo.

Updated the title of the issue, and think we found a good flow for these moving forward.

sips/sip-022/sip-022-emergency-pox-fix.md
obycode: Sounds like a good plan to me 👍

jcnelson (approved):

Issue #60, opened 2022-02-23
Trait for minting NFT's with any FT and STX using unified interface


LNow:
As the ecosystem grows we see more and more NFT contracts that can be minted not only with plain STX but also with different FT (ie. CityCoins). Each have a slightly different function name and signature. I think we can do better.

There should be a trait that defines unified interface which can be used to mint new NFT with STX and any SIP-010 compliant FT. Such trait could look like this:
```clarity
(use-trait sip-010-trait 'SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.sip-010-trait-ft-standard.sip-010-trait)

(define-trait (
    ;; mints 1 NFT using supplied FT as tender
    (mint-with (<sip-010-trait>) (response bool uint))

    ;; mints X NFTs using supplied FT as tender
    (mint-many-with (uint <sip-010-trait>) (response bool uint))

    ;; returns minting price in supplied FT
    (get-mint-price-in (<sip-010-trait>) (response uint uint))
))
```

And to make it work also for STX we could define global wrapped STX SIP-010 compliant FT that is not issuing new token, but uses plain STX:
```clarity
(impl-trait 'SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.sip-010-trait-ft-standard.sip-010-trait)

(define-read-only (get-balance (owner principal))
    (ok (stx-get-balance owner))
)

(define-read-only (get-decimals)
    (ok u6)
)

(define-read-only (get-name)
    (ok "Wrapped STX")
)

(define-read-only (get-symbol)
    (ok "WSTX")
)

(define-read-only (get-token-uri)
    (ok (some u"https://www.stacks.co"))
)

(define-read-only (get-total-supply)
    (ok stx-liquid-supply)
)

(define-public (transfer (amount uint) (sender principal) (recipient principal) (memo (optional (buff 34))))
    (begin
        (try! (stx-transfer? amount sender recipient))
        (match memo to-print (print to-print) 0x)
        (ok true)
    )
)
```
Very similar approach can be used by NFT marketplaces to define unified interface for listing/buying NFTs for STX and any SIP-010 compliant FT.

@Jamil @friedger @dantrevino @MarvinJanssen @radicleart

friedger:
@LNow Yes, please! That is a good proposal. We should also add standard error codes for 
* unsupported token: u500
* insufficient balance: u103
* not authorized to mint: u403

[...]

PR #140, opened 2023-05-12
SIP-022: update block activation for cycle 60

wileyj:
Adds a second addendum addressing the block activation heights for release 2.4. 
Accepted text is not changed; this PR adds a another section specifying that the activation will happen prior to cycle 60 and specifies the block height at which activation will occur. 

> The second hard fork will take effect 400 Bitcoin blocks prior to the start of the prepare phase for cycle #60, which is Bitcoin block height 791551.


obycode (approved):
Looks good!

PR #149, opened 2023-07-19
(Not ready for review) Better Blocks

jcnelson:
Supersedes #147

### DIFF
--- a/sips/sip-0xx/sip-0xx-nakamoto.md
+++ b/sips/sip-0xx/sip-0xx-nakamoto.md
@@ -1,0 +1,48 @@
+## Producer Set Selection
+
+The producer set selection for term _N_ occurs during term _N-2_. Similar to the leader block commitments used in the current miner selection process, as defined in [SIP-001](https://github.com/stacksgov/sips/blob/main/sips/sip-001/sip-001-burn-election.md) and amended in [SIP-007](https://github.com/stacksgov/sips/blob/main/sips/sip-007/sip-007-stacking-consensus.md), would-be producers issue a Bitcoin transaction known as a producer set transfer.
+
+### Producer Set Enrollments
+
+Producer set enrollments have the same constraints on the Bitcoin transaction's inputs as PoX leader block commitments. Specifically, the first input of this Bitcoin operation must originate from the same address as the second output of the [VRF key registration](https://github.com/stacksgov/sips/blob/main/sips/sip-001/sip-001-burn-election.md#leader-vrf-key-registrations). The first output of a producer set enrollment must be an `OP_RETURN` with the following data:
+
+```
+            0      2  3     7    11    15    19                         80
+            |------|--|-----|-----|-----|-----|-------------------------|
+             magic  op set   set   key   key       padding
+                       block txoff block txoff
+```
+
+Where `op = @` and:
+
+- `set_block` is the burn block height of the final block announced in the previous term, N-3. This ensures that the enrollment is only accepted if it is processed during the correct term.
+- `set_txoff` is the vtxindex for the final block announced in the previous term, N-3.
+- `key_block` is the burn block height of this producer's VRF key registration
+- `key_txoff` is the vtxindex for this producer's VRF key registration
+
+The subsequent output(s) in this transaction are the PoX outputs:
+
+1. If the producer set enrollment is in a reward cycle, then outputs 1 through 20 must go to the chosen PoX recipients.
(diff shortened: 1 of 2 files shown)


jcnelson:
@kantai @obycode Can we work on this PR instead of #147?  Are you able to commit to this branch directly?  Because, I can't commit to `hirosystems/sips`.

obycode:
> @kantai @obycode Can we work on this PR instead of #147?  Are you able to commit to this branch directly?  Because, I can't commit to `hirosystems/sips`.

I can but Aaron mentioned that he couldn't commit to the branch I had originally created.

sips/sip-0xx/sip-0xx-nakamoto.md
```diff
@@ -0,0 +1,658 @@
...
+
+Would-be producers with valid producer set enrollments in term _N-2_ are
+eligible to be included in the producer set for term _N_. The total number of
+producers in a set needs to be limited to prevent coinbase payouts from
+including too many accounts, as this could slow down event processing and even
+open a new DoS vector. This cap will also prevent the block-signing process from
+becoming too expensive. To this end, the total amount of BTC spent in the
+outputs described in "Producer Set Enrollments" is used to select the producer
+set. Would-be producers are ranked by these BTC expenditures, and the top 100
+will be selected for the producer set.
```
obycode, line 333: Previously, I brought up rolling over credits for the BTC expenditures of would-be producers that do not make it into the top 100. The point behind this would be to ensure that small miners wouldn't be disadvantaged by a large miner that could push them out of the top 100, thus wasting their BTC for no rewards. 

Open questions:
- Do you think this is necessary to avoid disadvantaging small miners? 
- Does this open a DoS attack? 
  - Maybe it should have some minimum amount, some maximum number of consecutive terms that credits can rollover, and/or some maximum number of producers.

sips/sip-0xx/sip-0xx-nakamoto.md
```diff
@@ -0,0 +1,658 @@
...
+
+## Producer Rewards
+
+During a term, producers in the set are eligible to receive a portion of the
+coinbase rewards and transaction fees for blocks they produce. Since a term is
+defined as 10 Bitcoin blocks, the coinbase reward is equal to 10 times the
+coinbase as defined in **_<which SIP defines this?>_**. This amount is
```
obycode, line 340: Where is the coinbase amount defined? I couldn't find it.

sips/sip-0xx/sip-0xx-nakamoto.md
```diff
@@ -0,0 +1,658 @@
...
+distributed to all producers in the set proportionally, based on the percentage
+of the total BTC spent in the producer set enrollments. All producers receive
+their portion of the coinbase, regardless of whether or not they signed the
+blocks produced by the set. The coinbase transaction should be the first
+transaction in a term.
+
+The producer set is then incentivized to continue producing blocks throughout
+the term by the transaction fees. Transaction fees are paid only to producer set
+participants who signed the blocks produced. For each block, _B_, the total BTC
```
obycode, line 349: Does this incentivize some kind of race to submit a block to the stackers to sign before all of the producers have had a chance to sign it (after 67% have signed)?

kantai:
> Are you able to commit to this branch directly? Because, I can't commit to hirosystems/sips.

No -- I don't have any commit permissions on this repo. You should have write perms to `hirosystems/sips` now -- can you push to my branch there?

sips/sip-0xx/sip-0xx-nakamoto.md
kantai: I think it does incentivize trying to minimize the set of signers, yes. I'm not sure if this is a time-race, though.

sips/sip-0xx/sip-0xx-nakamoto.md
kantai: > Do you think this is necessary to avoid disadvantaging small miners?

I don't think so. Looking at the history of miner contributions in Stacks 2.0, the highest number of miners is ~7-8, so I think it's probably not worth the mechanics in the protocol to support roll overs.

> Does this open a DoS attack?

Depending on implementation, yes. It would need a maximum rollover lifetime and a lower limit.

However, on reflection, I wonder if setting a "top 100" rather than a threshold like 1% of the total is a bad design. A top 100 incentivizes big miners to split up their commitments to crowd out smaller miners. For example:

jcnelson:
Yup, we're good -- I'll push to https://github.com/hirosystems/sips/commits/draft/better-blocks

PR #139, opened 2023-05-05
Multisignature transaction signing order independence

fess-v:
Hey everyone!

We are developing a multisig solution named Asigna (Asigna.io), and we have encountered an important issue. Following the weekly SIP call discussion, we created this SIP to expedite the process. We have attached two GitHub issues for stacks.js (https://github.com/hirosystems/stacks.js/issues/1487) and stacks blockchain (https://github.com/stacks-network/stacks-blockchain/issues/2622). This issue has been labeled as "in the icebox," and we would like to bring it to the attention of the Stacks development team.

Please note that the first commit only contains a draft proposal to initiate discussions and will be updated with more technical information later.

The initial discussion was initiated by the following post on the Stacks forum (https://forum.stacks.org/t/implementing-a-fully-functioning-multisig-wallet-on-stacks/14889).

The issue is in the next_verification method from the MultisigSpendingCondition struct and the verify method for message signatures. Verify goes subsequently over the initial public key array, therefore checks their signatures subsequently, while the next_verification method checks the next signature by wrapping the previous signature into a new hash, which makes it tied to the previous signer.
In my opinion, the solution is to untie this relation and make each sign only the original transaction by each signer and check their signatures accordingly.
The same thing should be implemented in the javascript Stacks.js library for signature construction.
It is also possible to change the order in which signatures are being checked and still wrap the previous signature with a new hash, but I just want to ask if it's really required.

Thanks everyone!

PR #152, opened 2023-08-31
SIP-027: non sequential multisig transactions

jbencin:
While implementing Stacks multisig in the Ledger app (Zondax/ledger-stacks#152), I found the current multisig format confusing and hard to work with.  Other developers that have tried to work with multisig seem feel the same way (example: PR #139), and as far as I know there is currently no Stacks wallet with full multisig support. So wrote up this SIP which makes slight modifications to SIP-005 to add a new multisig transaction type which is a bit simpler and allows participants to sign in any order.


AcrossfireX:
@jbencin - can you comment on how this relates to https://github.com/stacks-network/stacks-blockchain/pull/3710 and its associated SIP draft? Is it possible to fold these into one shared SIP effort? Let me know if it makes more sense for them to remain seperate.

jbencin:
@AcrossfireX That PR is now implementing this draft SIP. I know there is another draft SIP mentioning order-independent multisig (#139), but that one lacks implementation details

fess-v:
I will close the previous PR so we can move all discussions to this one instead @jbencin @AcrossfireX

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,129 @@
...
+   ECDSA recoverable signature) pair to recover the public key that generated it.
+   byte,
+6. Repeat step 4 for each signature, so that all of the public keys are
+   recovered.
+7. Verify that the sequence of public keys hash to the address, using
+   the address's indicated public key hashing algorithm.
+
+# Related Work
+
+This section will be expanded upon after this SIP is ratified.
```
MarvinJanssen, line 119: SIPs generally cannot be changed after ratification. The exceptions are to fix a mistake/typo or perhaps add a clarification.

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,129 @@
...
+Status: Draft
+
+Created: 2023-08-30
+
+License: BSD 2-Clause
+
+Sign-off:
+
+Discussions-To: https://github.com/stacksgov/sips
+
```
MarvinJanssen, line 22: `Layer` SIP field needs to be added: https://github.com/stacksgov/sips/blob/main/sips/sip-000/sip-000-stacks-improvement-proposal-process.md.

MarvinJanssen (commented):
I welcome this SIP as it will make multisig much easier to use on Stacks. Excited to see more work in this direction. Some initial comments.

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: I just copied this from another SIP. I updated it with a link to #139

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: This is `Consensus (hard fork)` right?

MarvinJanssen: Yes.

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,137 @@
...
+- This does not reduce the size of a transaction, as each intermediate signature must still be included
+- The algorithm for doing this is complex, and several developers have a hard time understanding and implementing it correctly
+
+This document proposes having each signer sign the transaction directly:
+
+```
+signature_1(tx), signature_2(tx), ..., signature_n(tx)
+```
+
+This would address all of the concerns listed above, and would not increase transaction size or make it easier to forge a signature
```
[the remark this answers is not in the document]
fess-v: I would add something like:
Imagine a DAO that has 5 management team members. They decide to make a 3 out of 5 multisig on Stacks.
With the current version, signing a transaction in the creation order is required, i.e., after the 3rd owner only the 4th and 5th can sign, after the 5th owner signed - no one can add another signature.
It makes usage of the multisig painful for a lot of scenarios, for example:
- It is impossible to start signing a transaction from the 4th and 5th owners since they won't have enough owners after them to finalize the transaction
- If the 1st owner initiated a transaction, the 4th owner signed, the 2nd and 3rd owners can't sign this transaction anymore, so only the 5th owner can finalize it, the case of 2nd or 3rd owners being online and 5th not - they need to wait until the last owner makes this signature.
- If the 3rd owner initiated and signed a transaction, he needs the 4th and 5th owners' signatures in order to finalize it. 1st and 2nd owners can't participate until they make a new signing flow for that transaction.
In short, for small multisig wallets it seems to be not a blocking problem, but for multisig wallets with 4 or more owners, it becomes really hard to initiate transactions and get signatures in place.
@jo-tm @jbencin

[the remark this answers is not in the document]
jbencin: @fess-v This is a good example, it shows that when you establish signer order by creating a multisig address, not all signers are equal. Those near the top of the list have much more flexibility than those near the bottom

Would you please add this to the draft SIP? While you're at it, add your name and email to the authors list, you've done a lot of work on this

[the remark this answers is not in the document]
fess-v: Sure! will do that soon, thank you @jbencin

AcrossfireX:
This SIP should be considered Accepted from the SIP Editors and should progress to technical CAB review. (Likely already underway)

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,152 @@
...
+This section should be interpreted as a patch to the existing "[Specification](https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md#specification)" section of SIP-005.
+For anything not mentioned here, the rules from SIP-005 still apply.
+
+## Transactions
+
+### Transaction Encoding
+
+#### Transaction Authorization
+
+Add new hash modes `0x05` and `0x07`. These numbers were chosen in order to used in order to maintain the following relationships:
```
obycode, line 84:
```suggestion
Add new hash modes `0x05` and `0x07`. These numbers were chosen in order to maintain the following relationships:
```

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,152 @@
...
+
+0. Extract the public key(s) and signature(s) from the spending condition.
+1. Clear the spending condition.
+2. Serialize the transaction into a byte sequence, and hash it to form an
+   initial `sighash`.
+3. Calculate the `presign-sighash` from the `sighash`, authorization type byte,
+   fee, and nonce.
+4. Use the `presign-sighash` and the next (public key encoding byte,
+   ECDSA recoverable signature) pair to recover the public key that generated it.
+   byte,
```
obycode, line 126:
```suggestion
```

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,152 @@
...
+
+Discussions-To: https://github.com/stacksgov/sips
+
+# Abstract
+
+This SIP proposes a new multisig transaction format which is intended to be easier to use than the current format described in SIP-005.
+It does not remove support for the current format, rather it is intended to co-exist with the old format and give users a choice of which format to use.
+
+The issue with the current format is that it establishes a signer order when funds are sent to multisig account address, and requires signers to sign in the same order to spend the funds.
+In practice, the current format has proven proven difficult for developers to understand and implement, as evidenced by the lack of Stacks multisig implementations today.
```
wileyj, line 29:
```suggestion
In practice, the current format has proven difficult to understand and implement, as evidenced by the lack of Stacks multisig implementations today.
```

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,152 @@
...
+
+This new format intends to simplify the signing algorithm and remove the requirement for in-order signing, without comprimising on security or increasing transaction size.
+It is expected that this will lead to better wallet support for Stacks multisig transactions.
+
+# Introduction
+
+Currently, a multisig transaction requires the first signer to sign the transaction itself, and following signers to sign the signature of the previous signer.
```
wileyj, line 36:
```suggestion
Currently, a multisig transaction requires the first signer to sign the transaction itself, and subsequent signers to sign the signature of the previous signer.
```

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,152 @@
...
+For a transaction with *n* signers, the final signature is generated in the following way:
+
+```
+signature_n(...(signature_2(signature_1(tx))))
+```
+
+There are a few drawbacks to doing it this way:
+
+- The order in which the signers must sign is fixed as soon as funds are send to a multisig account, which limits flexibility when creating a spending transaction from a multisig account
```
wileyj, line 45:
```suggestion
- The order in which the signers must sign is fixed as soon as funds are sent to a multisig account, which limits flexibility when creating a spending transaction from a multisig account
```

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,152 @@
...
+```
+signature_1(tx), signature_2(tx), ..., signature_n(tx)
+```
+
+This would address all of the concerns listed above, and would not increase transaction size or make it easier to forge a signature
+
+## Examples
+
+Imagine a DAO that has a management team comprised of five members.
+They create a 3 out of 5 multisig account on Stacks.
```
wileyj, line 61: Combine lines 60-61:
```
Imagine a DAO that has a management team comprised of five members, and tey create a 3 out of 5 multisig account on Stacks.
```

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,152 @@
...
+The existing multisig standard mandates that all transactions from this account be signed in an order which is established upon account creation.
+The ordering requirement creates a hierarchy where keys near the start of the sequence have more flexibility than those near the end. 
+To illustrate some of the limitations this creates:
+
+- In a scenario where the 1st member initiates a transaction and the 4th signs it, it prohibits the 2nd and 3rd members from signing. The responsibility then falls solely on the 5th member to finalize the transaction.
+- Once the 5th member has signed a transaction, no further signatures are possible.
+- If the 3rd member initiates a transaction, only the 4th and 5th members are eligible to provide subsequent signatures.
+- Initiating a transaction by the 4th or 5th member is impossible, as there are insufficient subsequent members to complete the signing process.
+
+To summarize, while such a multisig setup might suffice for smaller teams, as the number of required signers increases, it becomes increasingly difficult to create a transaction. This SIP aims to remove these limitations.
```
wileyj, line 71:
```suggestion
While such a multisig setup might suffice for smaller teams, as the number of required signers increases, it becomes increasingly difficult to create a transaction. This SIP aims to remove these limitations.
```

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,152 @@
...
+| `0x02` | Single-signature | 20 | Bitcoin P2WPK-P2SH |
+| `0x03` | Multi-signature | 20 | Bitcoin P2WSH-P2SH |
+| `0x05` | Non-sequential multi-signature | 20 | Bitcoin redeem script P2SH |
+| `0x07` | Non-sequential multi-signature | 20 | Bitcoin P2WSH-P2SH |
+
+#### Transaction Signing and Verifying
+
+The steps for signing a non-sequential multisig transaction (hash modes `0x05` and `0x07`) shall be as follows:
+
+0. Set the spending condition address, and optionally, its signature count.
```
wileyj, line 102: I think for SIPS/documents in general, we should start numbered lists with `1` rather than the *technically correct* `0`. 
Cosmetic change - I don't feel strongly if it remains starting at `0` though.

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,152 @@
...
+   `sighash` with the authorization type byte (0x04 or 0x05), the fee (as an 8-byte big-endian value),
+   and the nonce (as an 8-byte big-endian value).
+4. Calculate the ECDSA signature over the `presign-sighash` by treating this
+   hash as the message digest.  Note that the signature must be a `libsecp256k1`
+   recoverable signature. Store the message signature and public key encoding byte as a signature auth field.
+5. Repeat step 4 until the signer threshold is reached.
+
+The steps for verifying a non-sequential multisig transaction (hash modes `0x05` and `0x07`) shall be as follows:
+
+0. Extract the public key(s) and signature(s) from the spending condition.
```
wileyj, line 118: I think for SIPS/documents in general, we should start numbered lists with `1` rather than the *technically correct* `0`. 
Cosmetic change - I don't feel strongly if it remains starting at `0` though.

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,152 @@
...
+| `0x07` | Non-sequential multi-signature | 20 | Bitcoin P2WSH-P2SH |
+
+#### Transaction Signing and Verifying
+
+The steps for signing a non-sequential multisig transaction (hash modes `0x05` and `0x07`) shall be as follows:
+
+0. Set the spending condition address, and optionally, its signature count.
+1. Clear the other spending condition fields, using the appropriate algorithm below.
+   If this is a sponsored transaction, and the signer is the origin, then set the sponsor spending condition
+   to the "signing sentinel" value (see below).
```
wileyj, line 105: Can this `(see below)` point more directly? I don't see any other reference to a "signing sentinel". 
example:
```
..(see below in point 4)
```

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: I'd prefer to keep it at `0` *only* because this is a modification of SIP-05 and that uses 0 indexing for these steps

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,152 @@
...
+
+# Backwards Compatibility
+
+The Stacks Blockchain will continue to treat multisig transactions using the current format as valid.
+
+# Activation
+
+# Reference Implementations
+
+To be implemented in Rust. See https://github.com/blockstack/stacks-blockchain.
```
obycode, line 152: This points to the old repo. Is https://github.com/stacks-network/stacks-blockchain/pull/3710 an implementation of this? Should you link to that here?

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: This should be defined in SIP-05 where I copied it from

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: This is copy/pasted from an old SIP, I'll update it

jbencin:
One more small thing I want to add here...

Even though the signing order is independent, the order of the public keys in the auth fields still determines how the address is generated. For backwards compatibility and to enable previously generated multisig accounts to use order-independent signing, I don't think we should mandate public key order, but I think we should **recommend** a public key ordering of least to greatest (which is equivalent to lexicographical sorting of hex-encoded values) to remove the guesswork when creating a transaction

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: In general it's considered a best practice in Markdown to write one sentence per line. It displays the same, but makes diffs and merges easier

AcrossfireX:
@jcnelson - would we be able to get this in front of the Steering Committee for final approval and merge as the activation points to it being ready for epoch 3.0? It has been approved by the SIP editors and Technical CAB and seems ready for final discussions by the Steering Comittee.

Also would you prefer that the editors assign a formal SIP number to this as it has not yet been chosen. It appears that SIP-026 would be available.

obycode (approved):
This SIP was approved by the Technical CAB on [27 Oct 2023](https://github.com/stacksgov/sips/blob/main/considerations/minutes/technical-cab/2023-10-27-pr-152-153.md)

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,160 @@
...
+For a transaction with *n* signers, the final signature is generated in the following way:
+
+```
+signature_n(...(signature_2(signature_1(tx))))
+```
+
+There are a few drawbacks to doing it this way:
+
+- The order in which the signers must sign is fixed as soon as funds are sent to a multisig account, which limits flexibility when creating a spending transaction from a multisig account
+- The process of signing a transaction requires each signer to validate the entire signature chain before signing, in order to make sure it matches the transaction, leading to `O(n^2)` signing times
```
jcnelson, line 46: Slight correction -- it requires each signer to perform `O(n)` signature verifications.  The `O(n^2)` total verifications is distributed across `n` signers.

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,160 @@
...
+
+The steps for signing a non-sequential multisig transaction (hash modes `0x05` and `0x07`) shall be as follows:
+
+0. Set the spending condition address, and optionally, its signature count.
+1. Clear the other spending condition fields, using the appropriate algorithm below.
+   If this is a sponsored transaction, and the signer is the origin, then set the sponsor spending condition
+   to the "signing sentinel" value (see below).
+2. Serialize the transaction into a byte sequence, and hash it to form an
+   initial `sighash`.
+3. Calculate the `presign-sighash` over the `sighash` by hashing the
```
jcnelson, line 108: Can you be more specific in what you mean by "hashing the `sighash` with ...` here?  Do you mean to say that you concatenate the `sighash` with these fields?

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,160 @@
...
+   `sighash` with the authorization type byte (0x04 or 0x05), the fee (as an 8-byte big-endian value),
+   and the nonce (as an 8-byte big-endian value).
+4. Calculate the ECDSA signature over the `presign-sighash` by treating this
+   hash as the message digest.  Note that the signature must be a `libsecp256k1`
+   recoverable signature. Store the message signature and public key encoding byte as a signature auth field.
```
jcnelson, line 113: Please further specify whether or not the recoverable signature is in `vrs` or `rsv` order

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jcnelson: I don't think the signature count should be optional.  Either this algorithm needs to work without it, or it should be mandatory.

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,160 @@
...
+5. Repeat step 4 until the signer threshold is reached.
```
jcnelson, line 114: You're missing some steps related to how to serialize the resulting public keys and signatures into the transaction.

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jcnelson: How are the public keys and signatures encoded?

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,160 @@
...
+
+The steps for verifying a non-sequential multisig transaction (hash modes `0x05` and `0x07`) shall be as follows:
+
+0. Extract the public key(s) and signature(s) from the spending condition.
+1. Clear the spending condition.
```
jcnelson, line 119: Please explain what it means to "clear" the spending condition with respect to how it affects the binary representation. Someone implementing this SIP needs to know this in order to calculate the `sighash` in the next step.

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,160 @@
...
+2. Serialize the transaction into a byte sequence, and hash it to form an
+   initial `sighash`.
+3. Calculate the `presign-sighash` from the `sighash`, authorization type byte,
+   fee, and nonce.
+4. Use the `presign-sighash` and the next (public key encoding byte,
+   ECDSA recoverable signature) pair to recover the public key that generated it.
+6. Repeat step 4 for each signature, so that all of the public keys are
+   recovered.
+7. Verify that the sequence of public keys hash to the address, using
+   the address's indicated public key hashing algorithm.
```
jcnelson, line 129: So, SIP-005 includes a description of how each `TransactionAuth` struct is laid out in memory, as well as a description of what each of the bytes mean.  This is missing from this SIP.  Please add it.

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,160 @@
...
+
```
jcnelson, line 130: I think a missing requirement in this section is to mandate a particular public key representation. SIP-005 mandates a compressed representation, with an extra byte to indicate whether or not the _uncompressed_ representation should be used when calculating the address.

Does this SIP require that all public keys be compressed, or does it allow the same flexibility as SIP-005?  I think it needs to to the latter, since the design of the `TransactionAuth` structs is meant to support representing the same public key formats as those supported in Bitcoin.  The reason for this is that it's possible to determine the Stacks address of users who hold their STX with Bitcoin keys -- because the 20-byte hash component of Stacks and Bitcoin addresses is calculated the same way, these users (and counterparties that interact with them) can determine their funds' Stacks address without knowing the public keys.  I think we need to preserve that property here.

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: Yes, I'll clarify

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: Agreed, but I don't the code actually treats it as optional. This line was copied directly from SIP-005, so we may want to revise that also

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: Again, this was copied straight from SIP-005, so if we change one we should change the other

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: This is specified in SIP-005. I was trying to keep this document minimal and not reproduce everything here. I mention on lines 75-76 that for anything not explicitly mentioned here, SIP-005 still applies

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: Zero it. Again, same language as SIP-005

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: Same as SIP-005. Either format is allowed except in P2WSH where uncompressed keys are not

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: I've tried to say that everything in SIP-005 still applies. Can you think of a way to make this clearer without copying the entire text of SIP-005's "Transaction Encoding" section?

jbencin:
@jcnelson I copied all the text from SIP-005's "Transaction Encoding" section and merged it with this SIP. This can now be viewed as a complete replacement for that section

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,643 @@
...
+#### Chain ID
+
+The chain ID identifies which instance of the Stacks blockchain this transaction
+is destined for.  Because the main Stacks blockchain and Stacks app chains
+(described in a future SIP) share the same transaction wire format, this field
+is used to distinguish between each chain's transactions.  Transactions for the
+main Stacks blockchain MUST have a chain ID of `0x00000000`.
+
+#### Transaction Authorization
```
jcnelson, line 128: Hi @jbencin, you only need to include the description of the _new_ "Transaction Authorization" data this SIP presents.  You don't need anything else.  Specifically, the reader needs to be able to read this section, and know exactly how to construct byte-for-byte a valid, well-formed `TransactionAuth` that supports non-sequential signatures.

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jcnelson: The problem I'm pointing out here is that this step is not well-specified when taken out of context.  How is the reader supposed to know what hash algorithm to use, for example?  Can't the text just say it here?

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,643 @@
...
+
+Discussions-To: https://github.com/stacksgov/sips
+
+# Abstract
+
+This SIP proposes a new multisig transaction format which is intended to be easier to use than the current format described in SIP-005.
+It does not remove support for the current format, rather it is intended to co-exist with the old format and give users a choice of which format to use.
+
+The issue with the current format is that it establishes a signer order when funds are sent to multisig account address, and requires signers to sign in the same order to spend the funds.
+In practice, the current format has proven difficult to understand and implement, as evidenced by the lack of Stacks multisig implementations today.
```
kyranjamie, line 29:
```suggestion
In practice, the current format has proven difficult to understand and implement, as evident by the lack of Stacks multisig implementations today.
```

sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md
jbencin: @jcnelson I guess I misunderstood what you were asking for and included too much this time. I've updated it again so that it this SIP only contains the "Transaction Authorization" and "Transaction Signing and Verifying" sections. Because it was easy to do, these sections also contain details for the old multisig formats, so that it's no longer necessary for the user to read the corresponding sections of SIP-005

obycode (approved):
The latest updates look good to me.

radicleart:
Hi - i think the activation criteria for this SIP requires a community vote ?

If so the activation criteria sections needs to be updated to reflect this? For example, see https://github.com/stacksgov/sips/blob/a1c478e059ac1e1581adcbf15de94c6f66f3e83b/sips/sip-02x/sip-02x-non-sequential-multisig-transactions.md

radicleart:
Hi Please update the activation criteria to include the community vote. Something like the following will be great.


### Activation
Since this SIP requires a change to the stacks consensus rules a community vote is additionally required.

#### Process of Activation
Users can vote to approve this SIP with either their locked/stacked STX or with unlocked/liquid STX, or both. The criteria for the stacker and non-stacker voting is as follows.

**For Stackers:**

In order for this SIP to activate, the following criteria must be met by the set of Stacked STX:
- At least double the amount of Stacked STX locked by the largest Stacker in the cycle preceding the vote must vote at all to activate this SIP.
- Of the Stacked STX that vote, at least 70% of them must vote "yes."

The voting addresses will be;

- yBtcAddress: "399iMhKN9fjpPJLYHzieZA1PfHsFxijyVY",
- nBtcAddress: "31ssu69FmpxS6bAxjNrX1DfApD8RekK7kp",
- yStxAddress: "SPA17ZSXKXS4D8FC51H1KWQDFS31NM29SKZRTCF8",
- nStxAddress: "SP39DK8BWFM2SA0E3F6NA72104EYG9XB8NXZ91NBE"

which encode the hashes of the following phrases into bitcoin / stacks addresses: 

**Yes to Non-sequential Multisig Transactions**
**No to Non-sequential Multisig Transactions**

Stackers (pool and solo) vote by sending a dust stacks to the corresponding stacks address **from the account where their stacks are locked**.

Solo stackers only, can also vote by sending a bitcoin dust transaction (6000 sats) to the corresponding bitcoin address. **Note: bitcoin transactions must be sent from the same address (not wallet) that the user receives rewards to - solo stackers using wallets which rotate sending addresses need to take note that votes sent from other addresses in the same wallet will not be included in the final tally**.

**For Non-Stackers:**

Users with liquid STX can vote on proposals using the Ecosystem DAO. Liquid STX is the users balance, less any STX they have locked in PoX stacking protocol, at the block height at which the voting started (preventing the same STX from being transferred between accounts and used to effectively double vote). This is referred to generally as "snapshot" voting.

For this SIP to pass, 70% of all liquid STX committed by voting must be in favour of the proposal.

The act of not voting is the act of siding with the outcome, whatever it may be. We believe that these thresholds are sufficient to demonstrate interest from Stackers -- Stacks users who have a long-term interest in the Stacks blockchain's successful operation -- in performing this upgrade.

diwakergupta:
> Hi - i think the activation criteria for this SIP requires a community vote ?
> 
> If so the activation criteria sections needs to be updated to reflect this? For example, see [Nakamoto activation criteria](https://github.com/stacksgov/sips/blob/c297381235b2fd0ab9c5c360ff471def34052f2c/sips/sip-021/sip-021-nakamoto.md#process-of-activation)
> 
> @AcrossfireX can you please confirm this is the case ?

Why does this SIP require a community vote? Community votes is a very heavy mechanism (in terms of time, effort, communications etc) and should be used sparingly IMO. Is there clear guidance on which SIPs warrant a full-blown community vote?

obycode:
> Why does this SIP require a community vote?

This has been a discussion amongst the CABs, SIP editors, etc. and the current _consensus_ is that anything that is consensus changing should require a community vote. One goal will be to make the voting process simpler so that it's not such a big deal.

obycode:
Also, noting that the SIP doesn't have to be consensus breaking to require a community vote -- some SIPs might want a community vote even if not consensus changing -- but if it is consensus changing, then that triggers a vote requirement.

Hero-Gamer:
> **For Stackers:**
> 
> In order for this SIP to activate, the following criteria must be met by the set of Stacked STX:
> 
>     * At least double the amount of Stacked STX locked by the largest Stacker in the cycle preceding the vote must vote at all to activate this SIP.
> 
>     * Of the Stacked STX that vote, at least 70% of them must vote "yes."

My suggestion would be to change to: 
* **At least 80 million Stacked STX must vote at all to activate this SIP.** 

* **Of the Stacked STX that vote, at least 80% of them must vote "yes."**

Historically that's what we've done. E.g. in SIP-015. If you want to be more conservative, make it 50 million given it's a less advertised SIP vs Nakamoto/sBTC.


> 
> **For Non-Stackers:**
> 
> Users with liquid STX can vote on proposals using the Ecosystem DAO. Liquid STX is the users balance, less any STX they have locked in PoX stacking protocol, at the block height at which the voting started (preventing the same STX from being transferred between accounts and used to effectively double vote). This is referred to generally as "snapshot" voting.
> 
> For this SIP to pass, 70% of all liquid STX committed by voting must be in favour of the proposal.
> 

FYI Historically it has been: 
* **For this to pass 66% of all liquid STX committed by voting must in favour of the proposal.**

If wanna be safe, lower to 66%. I think it's fine to higher the bar to 70% based on looking at the typical % rate in favor over the past few votes. If you want to be more ambitious you can adjust it to 80%.

**I think 70% actually is a good experiment to trial.**

Hero-Gamer:
Hi @fess-v this SIP will be assigned to: **SIP-027**
Please can you update the text and content accordingly?

obycode:
The vote results need to be posted to this SIP, then it should be merged.

Hero-Gamer:
> The vote results need to be posted to this SIP, then it should be merged.

Brice I got all the stats for you!

Activation Status
At the end of cycle 90, the following vote was calculated. A total of 118,632,231 STX participated.
* For solo stacking, 100% voted 'Yes.' Total voting power is 3,449,000 STX balance with votes cast from 1 account. 
* For pool stacking, 100% voted 'Yes.' Total voting power is 114,914,556 STX balance with votes cast from 75 accounts. 
* For non-stackers, 99.9933% voted 'Yes.' Total voting power of ‘Yes’ is 268,674 STX balance with votes cast from 157 accounts.  For non-stackers, 0.0067% voted ’No.’ Total voting power of ‘No’ is 17 STX balance from 3 account.
268,691.89

All voting criteria from STX holders have been met. A breakdown of the transactions can be found [here](https://stx.eco/dao/proposals/SP3JP0N1ZXGASRJ0F7QAHWFPGTVK9T2XNXDB908Z.sip-027-multisig-transactions/results).
A copy of the scripts used to tabulate the solo and pool stacking can be found here.

**Script** will need the link for the script from **Albu** I've asked if he can request uploading to SIP-027 folder
so there last sentence "here" can have the URL.

Hero-Gamer:
cc: @jbencin

obycode (approved):
LGTM

jcnelson (approved):

jcnelson:
It is truly a pleasure to merge this in <3

jcnelson:
Hi reviewers,

Can you please approve and merge https://github.com/stacksgov/sips/pull/191 instead?  It contains some fixes to this SIP, which I can't push directly since this PR comes from @jbencin's account.

Thanks!

jcnelson (changes requested):
Only requesting changes to that the superseding PR https://github.com/stacksgov/sips/pull/191 can be reviewed and merged.  It contains the same commit history as this; I merely took the liberty of resolving my own comments on top of this.

Please review the above PR instead so we can get this merged in :tada:

sips/sip-027/sip-027-non-sequential-multisig-transactions.md
```diff
@@ -0,0 +1,413 @@
...
+
+## Process of Activation
+Users can vote to approve this SIP with either their locked/stacked STX or with unlocked/liquid STX, or both. The criteria for the stacker and non-stacker voting is as follows.
+
+## For Stackers:
+
+In order for this SIP to activate, the following criteria must be met by the set of Stacked STX:
+
+- At least double the amount of Stacked STX locked by the largest Stacker in the cycle preceding the vote must vote at all to activate this SIP.
+- Of the Stacked STX that vote, at least 70% of them must vote "yes."
```
Hero-Gamer, line 383: Just looking at the SIP-27 while reviewing SIP-28, noticed slight discrepancy 
between the [public blog announcement](https://stacks.org/sip-027-voting) of the Stackers' voting threshold vs the SIP text. 

Recommend update the SIP text to match the blog. 

i.e. 

* At least 80 million Stacked STX must vote at all to activate this SIP.

* Of the Stacked STX that vote, at least 80% of them must vote "yes."

jbencin:
Applied @Hero-Gamer's latest suggestion and merged @jcnelson's changes from #191

jcnelson (approved):

jcnelson:
just need one more approval cc @obycode

obycode (approved):

kelven7224 (approved):

kelven7224:
> While implementing Stacks multisig in the Ledger app (Zondax/ledger-stacks#152), I found the current multisig format confusing and hard to work with.  Other developers that have tried to work with multisig seem feel the same way (example: PR #139), and as far as I know there is currently no Stacks wallet with full multisig support. So wrote up this SIP which makes slight modifications to SIP-005 to add a new multisig transaction type which is a bit simpler and allows participants to sign in any order.

sips/sip-027/sip-027-non-sequential-multisig-transactions.md
kelven7224: > Just looking at the SIP-27 while reviewing SIP-28, noticed slight discrepancy 
> between the [public blog announcement](https://stacks.org/sip-027-voting) of the Stackers' voting threshold vs the SIP text. 
> 
> Recommend update the SIP text to match the blog. 
> 
> i.e. 
> 
> * At least 80 million Stacked STX must vote at all to activate this SIP.
> 
> * Of the Stacked STX that vote, at least 80% of them must vote "yes."
>

yawayl:
\

PR #155, opened 2023-10-02
SIP: Nakamoto v1 

jcnelson:
This is the SIP that describes the Nakamoto v1 system.  It implements fast and reliable blocks, and must be implemented before sBTC.

I've styled this SIP-021.  In my mind, it supersedes the [existing SIP-021](https://github.com/stacksgov/sips/pull/113).


sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1051 @@
...
+times.**
+1. **Require stackers to collaborate before the next block can be produced.**
+PoX Stacksrs will need to collectively validate, store, sign, and replicate each
+Stacks block the miner produces before the next block can be produced to be
+selected by cryptographic sortition. Stackers must do this in order to earn
+their PoX payouts and unlock their STX (i.e. PoX is now treated as compensation
+from the miner for playing this essential role).  In the proposed system, a
+cryptographic sortition only selects a new miner; it does not give the miner the
+power to orphan confirmed transactions as it does today. This will ensure that
+miners **do not produce fork and are able to confirm all prior Stacks blocks
```
jferrant, line 144: typo. Should be "do not produce forks" or "do not produce a fork"

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1051 @@
...
+payout in Bitcoin blocks it wins, but does not compete in other blocks, will
+_never_ win sortition unless the Bitcoin miner also has the majority of Bitcoin
+mining power.
+    1. The probability of winning is inversely proportional to the larger of
+either the system's total commits in the current block, or the system's median
+total commits over the past ten blocks.  This means that a sudden decrease in
+the Bitcoin block's total PoX payout, such as due to the exclusion of other
+Stacks miners' `block-commit`s by a Bitcoin MEV miner, does _not_ increase the
+likelihood that the MEV miner's _de minimis_ `block-commit` wins.
+    1. The probability of winning is absolute, not relative. The sum of each
```
jferrant, line 185: there is an issue with the numbers here. Both are 1. and 1. instead of 1. and 2. for these bullet points.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1051 @@
...
+Absent from this header is the VRF proof, because it only needs to be included
+once per tenure.  Instead, this information will be put into the Nakamoto
+`Coinbase` transaction, which has a different wire format than the current
+`Coinbase` transaction.
+
+All existing Stacks transactions, with the exception of `PoisonMicroblock`
+transactions and the current `Coinbase` transaction, will continue to be
+supported.  However, the anchor mode byte will be ignored.
+
+### Tenure Changes
```
jferrant, line 390: We reference tenure change a few times prior to this section. I feel this should either be included in the glossary or defined (at least briefly) prior to its first use in the doc.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1051 @@
...
+    * The 33-byte signed x-coordinate of the signature curve point
+    * The 32-byte scalar derived from the hash of the generator raised to the
+      _k_th power and the message
+  * A 33-byte **compressed secp256k1 public key**, consisting of its signed
+    x-coordinate
+
+To verify the transaction signature, the node both verifies that the public
+key's hash160 is equal to the spending condition's 20-byte hash, and verifies
+that the Schnorr signature over the `presign-sighash` value of the transaction
+was generated by corresponding private key.
```
jferrant, line 579: should say "by a corresponding private key". Also don't you usually verify against a pulbic key? not a private key?

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1051 @@
...
+
+### Example
+
+The need for this additional tweak becomes apparent when considering the
+consequences for a real Bitcoin MEV miner who is active on the Stacks network
+today: F2Pool.
+
+Consider what happens to F2Pool, who spends 200 sats on PoX and zero sats on
+transaction fees for their block-commit. Suppose the median total BTC spend over
+the last ten Bitcoin blocks was 500,000 sats (about what it is right now).  With
```
jferrant, line 680: semantics... but I would change "about what it is right now" to "about what it is at the time of writing" to be clear to the reader that this doc may have an out of date number

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1051 @@
...
+doing so, StackerDB-aware nodes eventually learn about all reachable nodes'
+StackerDB replicas as they walk the peer graph.
+
+StackerDB-aware nodes set up overlay networks on top of the Stacks peer-to-peer
+network to replicate StackerDB chunks.  Nodes that replicate a particular
+StackerDB will periodically exchange version vectors for the list of chunks in
+the replica.  If one node discovers that another node has a newer version of the
+chunk, it will download it and forward it to other StackerDB-aware nodes in the
+overlay that also need it.  In doing so, every StackerDB-aware node in the
+overlay eventually receives an up-to-date replica of the StackerDB chunks.
```
jferrant, line 989: Might be worthwhile to mention how long this data is stored. I think its for at most 2 reward cycles right?

jferrant: It also might be worthwhile to link to the stackerdb code docs if we have any (pretty sure we do. Could swear Jude sent me a link to them before)

jferrant:
Very minor comments on my part :D Great to see this update!

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+To achieve these goals this proposal makes the following changes to the stacks
+protocol:
+
+1. **Decouple Stacks tenure changes from Bitcoin block arrivals.** A miner may
+create many Stacks blocks per Bitcoin block, and the next miner must confirm
+_all of them_. There are no more microblocks or Bitcoin-anchored blocks;
+instead, there are only Nakamoto Stacks blocks. **This will achieve fast block
+times.**
+1. **Require stackers to collaborate before the next block can be produced.**
+PoX Stacksrs will need to collectively validate, store, sign, and replicate each
```
setzeus, line 126: Typo here with "Stacksrs."

setzeus (commented):
Still only ~halfway through but very detailed & informative. Only came across one typo so far.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+prior to this proposal's activation will be usable after this proposal
+activates.  Of particular note is that the `block-height` Clarity variable will
+increment at a much faster pace, since blocks will be produced much more
+quickly.
+
+A separate SIP will be written to describe any changes to Clarity that may take
+effect with the activation of this SIP.
+
+Some design decisions of this SIP were undertaken to support sBTC, which is
+described in a separate SIP.  In particular, the sBTC system requires a notion
```
diwakergupta, line 790: [todo] link to sbtc SIP?

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+settled until it is impossible for a chain reorg to remove it from the Stacks
+history.
+
+# Related Work
+
+This new system bears superficial similarity to proof-of-stake (PoS) systems.
+However, there are several crucial differences that place the Nakamoto Stacks
+system in a separate category of blockchains from PoS:
+
+* Anyone can produce blocks in Stacks by pending BTC.  How they get their BTC is
```
diwakergupta, line 801: [typo] s/pending/spending/

diwakergupta (commented):
Most of my comments are non-blocking. Thanks for writing this up Jude!!

sips/sip-021/sip-021-nakamoto.md
jcnelson: Added a definition in `Introduction/Design`

sips/sip-021/sip-021-nakamoto.md
jcnelson: Will rewrite to "...verifies the Schnorr signature over the `presign-sighash` value with the given public key`.

sips/sip-021/sip-021-nakamoto.md
jcnelson: It's indefinite -- as long as the slot exists across reconfiguration, it will continue to be preserved

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+
+# Introduction
+
+## Glossary
+
+| Term | Definition |
+|:-:|:-|
+|MEV| **Miner Extractable Value:** It is any extra amount of value that a miner can extract from the network by deviating from expected behavior: including, excluding, or reordering transactions in a block. |
+| cryptographic sortition | A process of randomly selecting one or more entities from a set using cryptography. This is a decentralized and verifiable way to select participants for a variety of tasks, such as consensus protocols, lotteries, and auctions. Further details are in [SIP-001][SIP-001-LINK]. |
+| stacker | Someone who locks up their STX tokens in order to support the network and earn Bitcoin rewards. Read more about [how stacking helps the network][HOW-STACKING-HELPS-THE-NETWORK-GIST]. |
```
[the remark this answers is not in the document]
jcnelson: Not that I know of

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+The cryptographic sortition process selects at most one `block-commit` in the
+subsequent burnchain block, which entitles the submitter to propagate their
+block and earn a block reward.
+
+## Problem Statement
+
+
+Over the last three years the stacks community has identified several issues:
+
+1. **Forks and missing blocks are disruptive to on-chain applications.** The act
```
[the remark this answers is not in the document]
jcnelson: Will reword

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+resolving key areas of improvement for the stacks protocol.
+
+# Specification
+
+## Overview
+
+Stackers subsume an essential role in the Nakamoto system that had previously
+been the responsibility of miners.  Before, miners both decided the contents of
+blocks, and decided whether or not to include them in the chain (i.e. by
+deciding whether or not to confirm them).  In this system, miners only decide
```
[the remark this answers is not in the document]
jcnelson: "Miners" is an older and more established term in the blockchain scene than "block producers" and "validators."

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+both canonical and valid, they collectively sign it and replicate it to the rest
+of the Stacks peer network.  Only at this point do nodes append the block to
+their chain histories.
+
+This new behavior prevents forks from arising.  If a miner builds a block atop a
+stale tip, Stackers will refuse to sign the block.  If Stackers cannot agree on
+the canonical Stacks tip, then no block will be appended in the first place.
+While this behavior creates a new failure mode for Stacks -- namely, the **chain
+can halt indefinitely** if Stackers cannot agree on the chain tip -- this is
+mitigated by having a large and diverse body of Stackers such that enough of
```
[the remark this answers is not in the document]
jcnelson: > Is there a specific number below will the protocol will halt indefinitely?

Yes -- one.  As long as one Stacker is online, the system will produce blocks.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+### Tenure Changes
+
+A _tenure change_ is an event in the Stacks blockchain when one miner assumes
+responsibility for confirming transactions from another miner.  The miner's
+_tenure_ lasts until the next cryptographic sortition.  Today, a tenure change
+in Stacks occurs when a Stacks block is discovered from a cryptographic
+sortition.
+
+In this proposal, the Stackers themselves carry out a tenure change by creating
+a specially-crafted `TenureChange` transaction which a miner must include in its
```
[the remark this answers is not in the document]
jcnelson: Miners fetch the transaction from Stackers via their StackerDB

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+(something that is possible today with microblocks, and one of the reasons why
+they are not adequate for addressing this proposal's concerns).
+
+![Untitled presentation
+(11)](https://github.com/stacks-network/stacks-blockchain/assets/459947/98381aa0-f8ff-4842-b15b-881f4c4424ab)
+
+<em>Figure 2: Tenure change overview.  When a new Bitcoin block arrives,
+Stackers begin the process of deciding the last block they will sign from miner
+N.  When they reach quorum, they make this data available for download by
+miners, and wrap it in a WSTS-signed specially-crafted data payload.  This
```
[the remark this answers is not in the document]
jcnelson: Miners would subscribe to the Stackers' StackerDB, which would ensure that this information is propagated by Stackers to their local nodes.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+transaction as its first transaction.
+3. It begins mining transactions out of the mempool to produce Stacks blocks.
+
+If miner _N_ cannott obtain or observe the `TenureChange` transaction, then it
+will keep producing blocks.  However, Stackers will not sign them, so as far as
+the rest of the network is concerned, these blocks never materialized.  If miner
+_N+1_ does not see the `TenureChange` transaction, it does not start mining; a
+delay in obtaining the `TenureChange` transaction can lead to a period of chain
+inactivity.  This can be mitigated by the fact that the miner can learn the set
+of Stackers' IP addresses in advance, can directly query them for the data.
```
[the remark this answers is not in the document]
jcnelson: No, it assumes that at least one public Stacks node exists which hosts a replica of the Stacker's StackerDB.  All other Stackers and miners can run behind NATs.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+each and every tenure, even if the tenure is empty.
+
+### Empty Tenures
+
+Sometimes, a tenure can be empty.  The miner and Stackers may fail to sign any
+Stacks block.  This can happen for banal reasons, such as (but not limited to):
+
+* The next Bitcoin block arrived before the miner could produce a block (e.g. a
+  flash block occurred)
+* The miner crashed before it could mine a block, and remained offline for the
```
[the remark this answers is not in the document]
jcnelson: I don't understand the question?

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+Internally, the Stacks node will call a _private_ function to load the reward
+set into the `.pox-4` data space after it identifies the PoX anchor block.  This
+is required for other dependent SIPs, such as the sBTC SIP.
+
+## Dynamic Runtime Budgets
+
+The time between cryptographic sortitions (and thus tenure changes) depends on
+the time between two consecutive burnchain blocks.  This can be highly variable,
+which complicates the task of sustaining a predictable transaction confirmation
+latency while also preventing a malicious miner from spamming the network with
```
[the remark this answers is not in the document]
jcnelson: The concern being addressed here is the need to bound the amount of computing resources a block can consume while being processed.  If we did not have such a limit, then miners could DoS the system.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+too many high-resource transactions.
+
+Today, each miner receives a tenure block budget, which places hard limits on
+how much CPU, RAM, and I/O their block can consume when evaluated.  In this
+proposal, each miner begins its tenure with a fixed budget, but Stackers may opt
```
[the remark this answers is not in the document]
jcnelson: This would be equivalent to the proposed system.  Either way, Stackers need to be able to throttle the miner's compute usage -- either by explicitly granting the miner an extension, or by explicitly preventing the miner from consuming resources too quickly (such as by NACKing their block or individual transactions within them).  But in both cases, the system must impose a fixed rate of resource consumption.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+This by itself is not a significant improvement -- F2Pool would be required to
+go from paying 200 sats to 1410 sats.  However, with this proposal's tweek,
+F2Pool would be required to _additionally_ win five Bitcoin blocks in a row in
+order to mine this cheaply.  Given that they have 13.89% of the mining power
+today, the odds of this happening by chance are only 0.005%.  Since this is
+unlikely -- about once every 20,000 Bitcoin blocks (once every 138.9 days) --
+F2Pool would instead be required to send legitimate `block-commit` transactions
+in at least 50% of the Bitcoin blocks.  In 87.11% of those, they would be paying
+the same transaction fees as every other Stacks miner.  This alone would cost
+them $106.13 USD/day.  With the additional _de minimis_ PoX payout, this rises
```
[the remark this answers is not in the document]
jcnelson: It's what Jesse used in his MEV report.

sips/sip-021/sip-021-nakamoto.md
jcnelson: Not in-scope.  sBTC needs Nakamoto, but Nakamoto does not need sBTC.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+major version would be bumped from 2 to 3.  The first Nakamoto release would be
+3.0.0.0.0.
+
+# Introduction
+
+## Glossary
+
+| Term | Definition |
+|:-:|:-|
+|MEV| **Miner Extractable Value:** It is any extra amount of value that a miner can extract from the network by deviating from expected behavior: including, excluding, or reordering transactions in a block. |
```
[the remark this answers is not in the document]
jcnelson: Our system has miners, however.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+coinbase and transaction fees before the MEV mitigation is applied.
+
+In the ATC-C algorithm, Nakamoto would use the document's recommended assumed
+total commitment function: the median total PoX spend across all miners for the
+past ten Bitcoin blocks.  It would additionally use the document's recommended
+carryforward function for missed sortitions' coinbases:  the coinbase for a
+Bitcoin block without a sortition would be available to winning miners across
+the next ten tenures.  That is, if a miner whose tenure begins during the next
+ten tenure-changes manages to produce a Stacks block with a `Coinbase`, then
+they receive a 20% of the coinbase that was lost.
```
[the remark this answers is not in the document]
jcnelson: Yes, it should be 10%. Thanks!

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+
+The reason ATC (and ATC-C) were not considered as viable anti-MEV strategies
+before is because a decrease in the PoX total spend can lead to a Bitcoin block
+with no sortition. This is a deliberate design choice in ATC-C, because it has
+the effect of lowering the expected return of MEV mining.  In ATC-C, the
+probability of a miner winning a sortition is _equal to_ (i.e. no longer
+proportional to) the miner's BTC spend, divided by the maximum of either the
+assumed total commit (median total BTC spend in the last 5 blocks) or the total
```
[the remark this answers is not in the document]
jcnelson: Thanks for the call-out.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+BTC spend in this Bitcoin block.  This means that the sum of each miners'
+winning probabilities is not guaranteed to be 1.  The system deals with this by
+creating a virtual "nul" miner that participates in each sortition, such that
```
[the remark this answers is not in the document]
jcnelson: Yup, thanks

sips/sip-021/sip-021-nakamoto.md
jcnelson: Yup, thanks

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+      _k_th power and the message
+  * A 33-byte **compressed secp256k1 public key**, consisting of its signed
+    x-coordinate
+
+To verify the transaction signature, the node both verifies that the public
+key's hash160 is equal to the spending condition's 20-byte hash, and verifies
+that the Schnorr signature over the `presign-sighash` value of the transaction
+was generated by corresponding private key.
+
+The address of a principal who signs a transaction with the WSTS Schnorr
```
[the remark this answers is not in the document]
jcnelson: I'm on the fence about this, tbh.  On the one hand, keeping the signing algorithm a secret is a privacy feature, much like how a p2sh or p2tr address in Bitcoin does not reveal anything about the spending conditions.  This goes one step further: the user can choose whether or not to sign with either ECDSA or Schnorr, but the address would be the same, thereby granting the user a degree of cryptographic agility.  This is because with ECDSA, we recover the original public key from the signature, and like in the described the Schnorr scheme, we verify that its hash160 equals the spending condition's 20-byte hash.

On the other hand, this means wallets would need to choose the signing algorithm when signing a transaction.  They'd want to use [RFC 6979](https://datatracker.ietf.org/doc/html/rfc6979) for their _k_ calculation, but I'm pretty sure they all do this these days.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1040 @@
...
+miner to produce blocks (Figure 1).
+
+As with the system today, miners submit their candidacy to produce blocks by
+sending `block-commit` transactions on the Bitcoin chain.  This proposal calls
+for altering the semantics of the `block-commit` in one key way: the
+`block_header_hash` field is no longer the hash of a proposed Stacks block (i.e.
+its `BlockHeaderHash`), but instead is the index block hash (i.e.
+`StacksBlockId`) of the _previous_ miner's first-ever produced block.
+
+![Untitled presentation
```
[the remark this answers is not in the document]
diwakergupta: Happy to help, should I just update this PR and send a separate PR that builds on this one?

sips/sip-021/sip-021-nakamoto.md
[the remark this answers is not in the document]
diwakergupta: yeah, nvm - poorly worded question and I think I got the answer. Basically I was trying to understand the recovery process for empty tenures.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1053 @@
...
+To support tenure changes, this proposal calls for a new PoX contract, `.pox-4`.
+The `.pox-4` contract would be altered over the current PoX contract (`.pox-3`)
+to meet the following requirements:
+
+* Stackers register a WSTS signing key when they call `stack-stx` and optionally
+  `delegate-stx`.
+    * In `delegate-stx`, Stackers may instead elect for the delegate to supply
+      the signing key.  In this case, the delegate carries out the Stacker's
+responsibility for signing blocks.
+* The stacking minimum is abolished.  If there is at least one stacker that is
```
diwakergupta, line 525: BTW this sentence is misleading: "the stacking minimum is abolished". I thought this was talking about the minimum threshold for someone to win a slot, but it's actually talking about PoX activation.

sips/sip-021/sip-021-nakamoto.md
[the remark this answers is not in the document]
diwakergupta: That doesn't seem like a good rationale to use jargon if simpler alternatives exist.

sips/sip-021/sip-021-nakamoto.md
diwakergupta: I don't follow: you already allude to 'sBTC, which is described in a separate SIP' - wouldn't linking to that SIP provide more context to the reader?

sips/sip-021/sip-021-nakamoto.md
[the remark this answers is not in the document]
jcnelson: A PR against this one, please :pray:

sips/sip-021/sip-021-nakamoto.md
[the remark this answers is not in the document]
jcnelson: This might be a @xoloki question.  Is there a risk of supporting two separate signing methods (ECDSA and Schnorr) for the same address, such that the user can choose whichever method they want at any time?  Namely, I'm concerned about the user becoming a signing oracle.

sips/sip-021/sip-021-nakamoto.md
jcnelson: I'm reluctant to have this SIP depend on the sBTC SIP if at all possible, because there's no guarantee that sBTC activates along with Nakamoto.  sBTC may even undergo significant revisions between Nakamoto's activation and sBTC's activation.  I'd almost rather drop this sentence, and instead describe how Nakamoto makes it easier to represent Bitcoin state on Stacks (which sBTC needs, but is by itself beneficial for other tech such as Stacks-wrapped ordinals).

sips/sip-021/sip-021-nakamoto.md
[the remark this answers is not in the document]
jcnelson: It seems to be a term of art in economics, which I think is entirely appropriate for a MEV discussion and for a conversation about blockchains in general.  A _de minimis_ PoX spend might be thought of similarly to a _de minimis_ tax or a _de minimis_ exchange of value.  The ideal mining system would ensure that _de minimis_ PoX spends yield _de minimis_ STX rewards.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1053 @@
...
+## Design
+
+To achieve these goals this proposal makes the following changes to the Stacks
+protocol:
+
+1. **Decouple Stacks tenure changes from Bitcoin block arrivals.**  In both
+   today's system and this proposal, miners take turns appending blocks to the
+Stacks blockchain -- the next miner is selected by cryptographic sortition,
+and the miner has the duration of the Bitcoin block (its _tenure_) to announce
+new block state.  This proposal calls for allowing a miner to
```
obycode, line 127:
```suggestion
new block state.  This proposal calls for allowing a miner to mine
```

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1085 @@
...
+about 250 vbytes, so that's 5500 sats, or about 1.41 USD with today's BTC price.
+So, F2Pool would lose money by MEV mining at their current rate if prices stay
+the same over those 4 months -- they'd forfeit about 3,525 USD in Bitcoin
+transaction fees (lost by excluding other Bitcoin transactions in order to
+include their block-commit) for a Stacks coinbase worth 500 USD. They'd have to
+pay about 1410 sats per block-commit just to break even, and they'd only recoup
+their investment on average once every 4 months.
+
+This by itself is not a significant improvement -- F2Pool would be required to
+go from paying 200 sats to 1410 sats.  However, with this proposal's tweek,
```
CharlieC3, line 725:
```suggestion
go from paying 200 sats to 1410 sats.  However, with this proposal's tweak,
```

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1085 @@
...
+to continue Stacking.  Furthermore, Stackers may delegate their signing
+authority to a third party if they feel that they cannot participate directly in
+block signing.
+
+That all said, the security budget of the chain is considerably larger in this
+proposal than before.  In order to reorg the Stacks chain, someone must take
+control of at least 70% of the STX that are currently Stacked.  If acquired at
+market prices, then at the time of this writing, that amounts to spending about
+$191 million USD.  By contrast, Stacks miners today spend a few hundred USD per
+Bitcoin block to mine a Stacks block.  Reaching the same economic resistence to
```
CharlieC3, line 776:
```suggestion
Bitcoin block to mine a Stacks block.  Reaching the same economic resistance to
```

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1085 @@
...
+* At least 80 million Stacked STX must vote _at all_ to activate this SIP.  This
+  number is chosen because it is more than double the amount of STX locked by
+the largest Stacker at the time of this writing (reward cycle 69).
+
+* Of the Stacked STX that vote, at least 80% of them must vote "yes."
+
+The act of not voting is the act of siding with the outcome, whatever it may be.
+We believe that these thresholds are sufficient to demonstrate interest from
+Stackers -- Stacks users who have a long-term interest in the Stacks
+blockchain's succesful operation -- in performing this upgrade.
```
CharlieC3, line 881:
```suggestion
blockchain's successful operation -- in performing this upgrade.
```

CharlieC3 (commented):
Only minor things, great job on this Jude as usual!

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1085 @@
...
+
+SIP Number: 021
+
+Title: Fast and Reliable Blocks through PoX-assisted Block Propagation
+
+Authors:
+* Aaron Blankstein <aaron@hiro.so>
+* Brice Dobry <brice@hiro.so>
+* Jacinta Ferrent <jacinta@trustmachines.co>
+* Jude Nelson <jude@Stacks.org>
```
csgui, line 11:
```suggestion
* Jude Nelson <jude@stacks.org>
```

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1085 @@
...
+      the signing key.  In this case, the delegate carries out the Stacker's
+responsibility for signing blocks.
+* The stacking minimum is abolished.  If there is at least one stacker that is
+  able to claim one reward slot, then PoX activates.
+* The `.pox-4` contract will expose each reward cycle's full reward set,
+  including the signing keys for each stacker, via a public function.
+Internally, the Stacks node will call a _private_ function to load the reward
+set into the `.pox-4` data space after it identifies the PoX anchor block.  This
+is required for other dependent SIPs, such as the sBTC SIP.
+
```
friedger, line 532: * Stacks token holders can't vote to deactivate PoX. Function `reject-pox` is removed in `.pox-4`.

friedger, line 532:
```suggestion
* Stacks token holders can't vote to deactivate PoX. Function `reject-pox` is removed in `.pox-4`.
```

friedger, line 532:
```suggestion
* Solo stacking is removed to simplify the stacking protocol. The functions `stack-stx`, `stack-extend`, `stack-increase` are removed.
* Decrease of stacked amount is added. The function `delegate-stack-decrease` is added.
```

friedger (commented):
Added more changes to pox-4 contract.

sips/sip-021/sip-021-nakamoto.md
314159265359879: Solo stacking is removed? does it mean someone who wants to stack on their own uses the pool-operater route instead (a pool with a single delegate)?

setzeus: Negative. Solo stacking is not removed afaik.

friedger: This is my proposal. It would simplify things.

friedger: The relevant issue on stacks-core is https://github.com/stacks-network/stacks-core/issues/4182

Solo stacking is used less and less and for the solo stackers security shouldn't be used at all. Hot wallet as pool operator makes more sense and is adopted by whales and institutional holders already.

setzeus: Can you clarify "the solo stackers security shouldn't be used at all?" A bit unclear to parse through that.

friedger: The security in general is increased the less the wallet holding large amounts is used.
Currently, solo stackers can use desktop Leather wallet with limited features or browser extensions with higher risk of being tricked into bad txs.
Using a solo pool would reduce the risk and increase the number of features that can be used.

friedger: Maybe delegate-stx-decrease might be out of scope for v1

jcnelson: The reason we didn't try to add a `-decrease` method for stacking and delegate-stacking is because it significantly complicates the auto-unlock logic.  Without careful design it could introduce a DoS vector, whereby someone could stack and then decrease to 1 uSTX over many, many iterations in order to blow up the number of unlocks each node would have to process.  Can you post a more detailed design proposal for `-decrease`?

friedger: Update: I removed the suggestion to add `-decrease`

Hero-Gamer:
- Was just looking at the Activation Criteria text, seems **Non-Stackers** AC have not been yet specified in the text. 
- Recommend adding same/similar Activation Criteria from SIP-015 to this Nakamoto SIP, used it last year didn't seem to have a problem, and seem reasonable bar to pass.
- Mike Cohen seems to be getting the Nakamoto Upgrade's Voting platform together quite well, all Stackers and Non-Stackers will be able to vote on a single website.
image

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,541 @@
...
+  - **(NEW)** The first transaction after the `TenureChange` transaction is a `Coinbase`.
+- All transactions either run to completion, or fail due to runtime errors. That is:
+  - The transaction is well-formed
+  - All transactions' senders and sponsors are able to pay the transaction fee
+  - The runtime budget for the tenure is not exceeded
+  - **(NEW)** The total runtime budget is equal to the runtime budget for one tenure, multiplied by the number of valid `TenureExtension` transactions mined in this tenure.
+  - No expression exceeds the maximum nesting depth
+  - No supertype is too large
+  - **(NEW)** The `PoisonMicroblock` transaction variant is not supported anymore
+  - **(NEW)** The current `Coinbase` transaction variant is not supported anymore
```
obycode, line 290:
```suggestion
  - **(NEW)** The `PoisonMicroblock` transaction variant is no longer supported
  - **(NEW)** The current `Coinbase` transaction variant is no longer supported
```

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,541 @@
...
+
+Importantly, this transaction replay feature is directed exclusively by Stacker and miner policy logic. It is not consensus-critical, and in fact cannot be because not all miners or Stackers may have even seen the orphaned Bitcoin state (which precludes them from independently identifying replay transactions; they must instead work together to do so off-chain). Therefore, this feature's implementation can be deferred until after this SIP is ratified.
+
+# Backwards Compatibility
+
+This proposal is a breaking change. However, all smart contracts published prior to this proposal's activation will be usable after this proposal activates. Of particular note is that the block-height Clarity variable will increment at a much faster pace, since blocks will be produced much more quickly.
```
obycode, line 379:
```suggestion
This proposal is a breaking change. However, all smart contracts published prior to this proposal's activation will be usable after this proposal activates.
```

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,541 @@
...
+
+A separate SIP will be written to describe any changes to Clarity that may take effect with the activation of this SIP.
```
obycode, line 381: I believe the only changes to Clarity to activate with this SIP are already documented in this SIP. I think this sentence can be removed.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,541 @@
...
+
+#### Solo Stacking
+
+The user must send a minimal amount of BTC from their PoX reward address to one of the following Bitcoin addresses:
+
+- For **"yes"**, the address is `11111111111111X6zHB1bPW6NJxw6`. This is the base58check encoding of the hash in the Bitcoin script `OP_DUP` `OP_HASH160` `000000000000000000000000007965732d332e30` `OP_EQUALVERIFY` `OP_CHECKSIG`. The value `000000000000000000000000007965732d332e30` encodes "yes-3.0" in ASCII, with 0-padding.
+
+For **"no"**, the address is `1111111111111117Crbcbt8W5dSU7`. This is the base58check encoding of the hash in the Bitcoin script OP_DUP OP_HASH160 `00000000000000000000000000006e6f2d332e30` `OP_EQUALVERIFY` `OP_CHECKSIG`. The value `00000000000000000000000000006e6f2d332e30` encodes "no-3.0" in ASCII, with 0-padding.
+
+From there, the vote tabulation software will track the Bitcoin transaction back to the PoX address in the .pox-3 contract that sent it, and identify the quantity of STX it represents. The STX will count towards a "yes" or "no" based on the Bitcoin address the PoX address sends to.
```
obycode, line 451:
```suggestion
From there, the vote tabulation software will track the Bitcoin transaction back to the PoX address in the .pox-3 contract that sent it, and identify the quantity of STX it represents. The STX will count towards a "yes" or "no" based on the Bitcoin address to which the PoX address sends.
```

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,541 @@
...
+Activating Nakamoto will include two epochs:
+
+- **Epoch 2.5:** Pox-4 contract is booted up but no Nakamoto consensus rules take effect.
+- **Epoch 3:** Nakamoto consensus rules take effect.
+
+There will be at least one reward cycle between the initiation of Epoch 2.5 and Epoch 3, but the exact duration will need to be determined prior to activation.
+
+### Updated Block Limits
+
+Block limits in the initial system described in [SIP-001][SIP-001-link] were designed conservatively such that a low compute threshold was required to run a functional node. These cost limits have been updated twice before, as improvements have been made to the stacks-node. This SIP calls for another block limit update to  reflect recent optimizations in the reference Clarity virtual machine.
```
obycode, line 416: I think we want to say something here about the block limits also changing due to the new mechanism in Nakamoto, in which the block creation is no longer a race with Bitcoin.

sips/sip-021/sip-021-nakamoto.md
AshtonStephens: Deleting.

sips/sip-021/sip-021-nakamoto.md
AshtonStephens: What if we changed the wording of the last sentence to this:

> This SIP calls for another block limit update to reflect recent optimizations in the reference Clarity virtual machine and the new mechanism in Nakamoto in which block creation is no longer a race with Bitcoin.

AshtonStephens: I plan on changing this section to this:

> Block limits in the initial system described in [SIP-001][SIP-001-link] were designed conservatively such that a low compute threshold was required to run a functional node. These cost limits have been updated twice before, as improvements have been made to the stacks-node. This SIP calls for another block limit update to reflect several changes:
> 
> 1. Contract calls are less computationally expensive due to practical optimizations in the Clarity virtual machine.
> 2. Mining a Stacks block is not a race against the next Bitcoin block under Nakamoto.
> 3. The block limit is now spread over multiple Stacks blocks within a miner's tenure.

AshtonStephens: Committed in [358a42c](https://github.com/stacksgov/sips/pull/155/commits/358a42cc188c776fb3559d0a837f708c44abcd2b)

Hero-Gamer:
This SIP affects MEV incentives, would it be better if we ask Economic CAB to be part of the sign-off as well?

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1051 @@
...
+In this proposal, a block is valid if and only if the following are true:
+
+* The block is well-formed
+    * It has the correct version and mainnet/testnet flag
+    * **(NEW)** Its header contains the same Bitcoin block hash as the Bitcoin
+      block that contains its tenure's block-commit transaction
+    * The transaction Merkle tree root is consistent with the transactions
+    * The state root hash matches the MARF tip root hash once all transactions
+      are applied
+    * **(NEW)** the block header has a valid ECDSA signature from the miner
```
dantrevino: Should the validation method be specified here?

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,1051 @@
...
+    * **(NEW)** the block header has a valid WSTS Schnorr signature from the set
```
dantrevino: Should the validation method be specified here?

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,542 @@
...
+4. **Stacks forks arise due to poorly-connected miners.** If a set of miners has a hard time learning the canonical Stacks chain tip when they submit block-commits, then they will collectively orphan other miners who are better-connected. This has happened in practice.
+5. **Some Bitcoin miners run their own Stacks miners and deliberately exclude other Stacks miners' `block-commits` from their Bitcoin blocks.** Once the STX block reward became sufficiently large this allowed them to pay a trivial PoX payout while guaranteeing that they would win the cryptographic sortition in their Bitcoin block. This was anticipated in the original design but the regularity with which it happens today is greater than the original protocol accounted for, and thus must be addressed now.
+
+## Proposed Solution
+To address these shortcomings, this proposal calls for four fundamental changes to the way Stacks works.
```
csgui, line 76: That phrase is mentioning **four** fundamental changes. But only **three** changes are described:

1. Fast blocks;
2. Bitcoin finality;
3. Bitcoin Miner MEV Resistance.

Is it missing the description of the fourth change?

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,542 @@
...
+|cryptographic sortition|A process of randomly selecting one or more entities from a set using cryptography. This is a decentralized and verifiable way to select participants for a variety of tasks, such as consensus protocols, lotteries, and auctions.|
+|miner selection (aka. sortition)|Also known as “leader selection,” “miner election” or simply “sortition,” miner selection is the weighted cryptographic sortition process by which a miner candidate is selected as the next miner (leader). Details of this process are in [SIP-001][SIP-001-link] with mechanism alterations in [SIP-007][SIP-007-link].<br><br>Within this SIP, we will refer to the miner selection process as simply “the sortition.”|
```
csgui, line 45:
```suggestion
|miner selection (aka. sortition)|Also known as “leader selection”, “miner election” or simply “sortition”, miner selection is the weighted cryptographic sortition process by which a miner candidate is selected as the next miner (leader). Details of this process are in [SIP-001][SIP-001-link] with mechanism alterations in [SIP-007][SIP-007-link].<br><br>Within this SIP, we will refer to the miner selection process as simply “the sortition".|
```

obycode: I believe this is just American (as-is) vs British (your suggestion) standards. Either could be correct for this doc.

csgui: I should admit... never realized that!!!  Thanks @obycode Now I am seeing `.,"` everywhere! 👀

obycode: @csgui and I were just chatting about this, and the lack of an Oxford comma in this sentence 😱 😄. At some point, it would be awesome if there was a team of technical writers just focused on this part of these docs, at which point we could pick some common style guide and stick with it. A tech writer I worked with previously liked Microsoft's style guide - https://learn.microsoft.com/en-us/style-guide/

Regardless, this is not an urgent issue at the moment. Clarity is the only requirement for now.

sips/sip-021/sip-021-nakamoto.md
csgui: Got you! Thanks.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,542 @@
...
+
+The weighted threshold signature is a Schnorr signature generated through a variation of the FROST protocol [1]. Each Stacker generates a signing key pair, and they collectively generate an aggregate public key for nodes to use to verify signatures computed through a distributed signing protocol. This signing protocol allocates shares of the associated aggregate private key to Stackers proportional to the number of reward slots they clinch. No Stacker learns the aggregate private key; Stackers instead compute shares of the private key and use them to compute shares of a signature, which can be combined into a single Schnorr signature.
+
+When a miner produces a block, Stackers execute a distributed signing protocol to collectively generate a single Schnorr signature for the block. Crucially, the signing protocol will succeed only if at least X% of the reward slots are accounted for in the aggregate signature. This proposal calls for a 70% signing threshold -- at least 70% of the reward slots (by proxy, 70% of the stacked STX) must sign a block in order to append it to the Stacks blockchain.
```
kantai, line 110: I think this paragraph as a whole needs to be rephrased and clarified. My understanding of the system is the following:

1. There is no minimum percentage of total STX participating for initializing the signing set. If only 1 person "stacked" and cleared the minimum threshold for stacking, they would get 1 slot, and that 1 slot would be sufficient for the signing set.
2. The signing threshold is a percentage of the reward slots actually assigned. If there are 2000 assigned reward slots, then the signing threshold is 70% of the 2000 slots.

This paragraph in the SIP doesn't agree with that.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,547 @@
...
+
+If the Stacks address holder votes for both "yes" and "no" by the end of the vote period, the vote will be discarded.
+
+## For Non-Stackers
+
+If the user is _not_ Stacking, then they can still vote with their liquid STX.
+***TODO: Details need to be added here***
+
+### For Miners
+There is only one criterion for miners to activate this SIP: they must mine the Stacks blockchain up to and past the end of the voting period. In all reward cycles between cycle 75 and the end of the voting period, PoX must activate.
```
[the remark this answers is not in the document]
friedger: Condition for 75 is true already. It does not change anything for miner voting.

sips/sip-021/sip-021-nakamoto.md
obycode: I addressed this in #168

sips/sip-021/sip-021-nakamoto.md
AshtonStephens: Validation is specified in detail below, I don't think it makes sense to add details here.

sips/sip-021/sip-021-nakamoto.md
AshtonStephens: I actually think not - the WSTS Schnorr signature has a defined validity that already is encompassed by this line. What would you make clearer?

obycode (dismissed):
The Technical CAB has approved this SIP. See meeting minutes and vote results [here](https://github.com/stacksgov/sips/blob/main/considerations/minutes/technical-cab/2024-01-31-sip-021.md). 👍

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,549 @@
...
+
+1. Contract calls are less computationally expensive due to practical optimizations in the Clarity virtual machine.
+2. Mining a Stacks block is no longer a race against the next Bitcoin block.
+3. The block limit is now spread over multiple Stacks blocks within a miner's tenure.
+
+A new block limit will be set based on benchmarks of the reference implementation such that a tenure can have its entirety executed over the course of some period less than 10 minutes on reasonable hardware.
+
+### Signer Liveness Enforcement
+
+Once epoch 3.0 goes live, Stackers will be required to actively participate in the system by signing and rejecting valid blocks from the currently tenured miner. This will be enforced by the withholding of PoX payouts during periods in which any non-participating Stackers are inactive. In this way Stackers cannot benefit from the system without participating.
```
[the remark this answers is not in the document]
AshtonStephens: First read was correct - made it clearer PR #171.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,549 @@
...
+
+# Activation
+
+## Determinations Prior to SIP Activation
+
+Some aspects of this proposal are best decided after a reference implementation approaches code completion. The following will be solidified at a later date but will strictly remain within the bounds described below.
+
+### Block Reward Distribution and MEV
+
+The Nakamoto system will use a variation of the Assumed Total Commitment with Carryforward (ATC-C) MEV mitigation strategy described in [this document][MEV-analysis-link] to allocate block rewards to miners. The probability a miner will win the sortition and be granted the current tenure will be based on a function that accounts for the total block commit spend on the blocks leading up to the current sortition.
```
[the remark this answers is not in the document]
AshtonStephens: Added reference in PR #171. Not including license because it doesn't need one.

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,549 @@
...
+## Signer Delegation
+
+Users who Stack may not want to run a 24/7 signing daemon. If not, then they can simply report some other signer's public key when calling `stack-stx` or `delegate-stack-stx`. Then, this other entity would run the signing daemon on their behalf.
+
+While this does induce some consolidation pressure, we believe it is the least-bad option. Some users will inevitably want to outsource the signing responsibility to a third party. However, trying to prevent this programmatically would only encourage users to find work-arounds that are even more risky. For example, requiring users to sign with the same key that owns their STX would simply encourage them to trust a 3rd party to both hold and stack their STX on their behalf, which is worse than just outsourcing the signing responsibility.
+
+# References
+
+- [1] https://eprint.iacr.org/2020/852.pdf
+- [2] https://trust-machines.github.io/wsts/wsts.pdf
```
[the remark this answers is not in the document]
AshtonStephens: Fixed in #171

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,549 @@
...
+
+## Signer Delegation
+
+Users who Stack may not want to run a 24/7 signing daemon. If not, then they can simply report some other signer's public key when calling `stack-stx` or `delegate-stack-stx`. Then, this other entity would run the signing daemon on their behalf.
+
+While this does induce some consolidation pressure, we believe it is the least-bad option. Some users will inevitably want to outsource the signing responsibility to a third party. However, trying to prevent this programmatically would only encourage users to find work-arounds that are even more risky. For example, requiring users to sign with the same key that owns their STX would simply encourage them to trust a 3rd party to both hold and stack their STX on their behalf, which is worse than just outsourcing the signing responsibility.
+
+# References
+
+- [1] https://eprint.iacr.org/2020/852.pdf
```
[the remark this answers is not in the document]
AshtonStephens: Fixed in #171

obycode (approved):

whoabuddy:
The Governance CAB has approved this SIP. See meeting minutes and vote results [in this PR](#172). :+1:

jiga (approved):
Adding SIP Editors acceptance.

jiga (approved):

jiga (approved):

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,556 @@
...
+|-|-|-|
+|version|Version number to describe how to validate the block.|1 byte|
+|chain length|The total number of StacksBlock and NakamotoBlocks preceding this block in this block's history.|8 bytes, big-endian|
+|burn spent|Total amount of BTC spent producing the sortition that selected the miner whose tenure produced this block.|8 bytes, big-endian|
+|consensus hash|The consensus hash of the burnchain block that selected this tenure.  The consensus hash uniquely identifies this tenure, including across all Bitcoin forks.|20 bytes|
+|parent block ID|The index block hash of the immediate parent of this block. This is the hash of the parent block's hash and consensus hash.|32 bytes|
+|transaction Merkle root|The SHA512/256 root hash of the binary Merkle tree calculated over the sequence of transactions in this block.|32 bytes|
+|state index Merkle root|The SHA512/256 root hash of the MARF once all of the contained transactions are processed.|32 bytes|
+|miner signature|Recoverable ECDSA signature from the tenure's miner.|65 bytes|
+|stacker signature|A Schnorr signature collectively generated by the set of Stackers over this block|65 bytes|
```
netrome, line 172: Nit: Isn't a Schnorr signature 64 bytes?

```suggestion
|stacker signature|A Schnorr signature collectively generated by the set of Stackers over this block|64 bytes|
```

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,556 @@
...
+|previous tenure consensus hash|Consensus hash of the previous tenure.  Corresponds to the sortition of the previous winning block-commit.|20 bytes|
+|burn view consensus hash|Current consensus hash on the underlying burnchain.  Corresponds to the last-seen sortition.|20 bytes|
+|previous tenure end|The index block hash of the last Stacks block from the previous tenure.|32 bytes|
+|previous tenure blocks|The number of blocks produced since the last sortition-linked tenure.|4 bytes, big-endian|
+|cause|A flag to indicate the cause of this tenure change<br>- `0x00` indicates that a sortition occurred, and a new miner should begin producing blocks.<br>- `0x01` indicates that the current miner should continue producing blocks. The current miner’s tenure execution budget is reset upon processing this transaction.|1 byte|
+|pubkey hash|The ECDSA public key hash of the current tenure.|20 bytes|
+|signature|The Stacker signature.|65 bytes|
```
netrome, line 209: Same as above (Schnorr signatures are 64 bytes)

```suggestion
|signature|The Stacker signature.|64 bytes|
```

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,556 @@
...
+|chain length|The total number of StacksBlock and NakamotoBlocks preceding this block in this block's history.|8 bytes, big-endian|
+|burn spent|Total amount of BTC spent producing the sortition that selected the miner whose tenure produced this block.|8 bytes, big-endian|
+|consensus hash|The consensus hash of the burnchain block that selected this tenure.  The consensus hash uniquely identifies this tenure, including across all Bitcoin forks.|20 bytes|
+|parent block ID|The index block hash of the immediate parent of this block. This is the hash of the parent block's hash and consensus hash.|32 bytes|
+|transaction Merkle root|The SHA512/256 root hash of the binary Merkle tree calculated over the sequence of transactions in this block.|32 bytes|
+|state index Merkle root|The SHA512/256 root hash of the MARF once all of the contained transactions are processed.|32 bytes|
+|miner signature|Recoverable ECDSA signature from the tenure's miner.|65 bytes|
+|stacker signature|A Schnorr signature collectively generated by the set of Stackers over this block|65 bytes|
+
+**Total bytes:** 263
```
netrome, line 174:
```suggestion
**Total bytes:** 262
```

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,556 @@
...
+
+## Stacker Transaction Inclusion
+
+Some transactions, like `TenureChange`, are generated by Stackers for inclusion in the blockchain. These transactions will have a 0-STX fee, so that Stackers are not in a position of needing to pay STX to do their jobs. They compel miners to include them in their blocks by keeping track of any such pending transactions in their StackerDB, and refusing to sign blocks unless the miner includes them.
+
+The miner obtains these transactions by querying the StackerDB instance used by Stackers. Miners' nodes may subscribe to the StackerDB instance in order to maintain an up-to-date replica.
+
+## Future Work: Guaranteed Transaction Replay
+
+In the event of a Bitcoin fork, Stackers would maintain the list of transactions which miners must reply to in their StackerDB as well. The miner would read this information from the Stackers' StackerDB in order to re-mine them into new Stacks blocks after the Bitcoin fork resolves.
```
netrome, line 532:
```suggestion
In the event of a Bitcoin fork, Stackers would maintain the list of transactions which miners must replay to in their StackerDB as well. The miner would read this information from the Stackers' StackerDB in order to re-mine them into new Stacks blocks after the Bitcoin fork resolves.
```

sips/sip-021/sip-021-nakamoto.md
obycode: This looks to be 160 bytes in the code. We also have added the `signer_bitvec`, 32 bytes.

sips/sip-021/sip-021-nakamoto.md
obycode: The `signature` and the `signers` bitmap have been removed.

obycode:
This SIP was also approved by the Economics CAB. See meeting minutes [here](https://github.com/stacksgov/sips/blob/main/considerations/minutes/economic-cab/2024-02-01-sip-021.md)

sips/sip-021/sip-021-nakamoto.md
AshtonStephens: The Stacker Signature in the [block header](https://github.com/stacks-network/stacks-core/blob/next/stackslib/src/chainstate/nakamoto/mod.rs#L307) is of type [`MessageSignature`](https://github.com/stacks-network/stacks-core/blob/master/stacks-common/src/util/secp256k1.rs#L63) which has the following definition:

```rust
pub struct MessageSignature(pub [u8; 65]);
```

But it acutally seems that the serialization foregoes the 65'th byte [here](https://github.com/stacks-network/stacks-core/blob/master/stacks-common/src/util/secp256k1.rs#L105)?

AshtonStephens: Quickly confirmed with @jcnelson, it's 65 bytes.

sips/sip-021/sip-021-nakamoto.md
AshtonStephens: Removed.

sips/sip-021/sip-021-nakamoto.md
AshtonStephens: Updated because of the bitvec addition to 267 + ceil(num_stackers / 8)

obycode (approved):

sips/sip-021/sip-021-nakamoto.md
obycode: I think that this probably should have been included so that it was explicit, but I think it's too late to add something like this now. We only want to add/change inconsequential things in the doc after approval.

obycode:
Gina, @jcnelson, or @MarvinJanssen -- can we get a sing-off from the steering committee added, then we should be ready to merge!

sips/sip-021/sip-021-nakamoto.md
netrome: That's not the same type though. The `MessageSignature` struct you linked seems to represent a 65 byte recoverable ECDSA signature. This table states that it is a Schnorr signature, which unless I'm mistaken has no 65 byte recoverable counterpart but instead should be just 64 bytes.

netrome: From the Rust Bitcoin library https://docs.rs/bitcoin/0.32.0/bitcoin/key/constants/constant.SCHNORR_SIGNATURE_SIZE.html
```
pub const SCHNORR_SIGNATURE_SIZE: [usize](https://doc.rust-lang.org/nightly/std/primitive.usize.html) = 64; // 64usize
```

netrome: Oh nevermind I see that the `ThresholdSignature` struct serializes to 65 bytes ([implementation link](https://github.com/stacks-network/stacks-core/blob/9775102297632bd07a87d15a80a9d24cd9e40d19/stackslib/src/chainstate/stacks/transaction.rs#L156)). I'm curious to know where the extra byte comes from. Even BIP-340 specifies that the signature is a 64-byte array.

https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#verification

netrome: Does this imply that the Nakamoto signatures are Bitcoin incompatible? I guess we don't need them to be for the purpose of Nakamoto consensus.

netrome: Also looking at the WSTS paper, the only difference is that a BIP-340 signature only stores the x coordinate of the sum of the public nonces whereas this signatures stores the full point in compressed form - so making this bitcoin compatible is trivial.

jiga (approved):
approval from SIP Editors

obycode (approved):

jiga (approved):

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -0,0 +1,562 @@
...
+### Examples
+
+#### Voting "yes" as a solo Stacker
+
+Suppose Alice has stacked 100,000 STX to `1LP3pniXxjSMqyLmrKHpdmoYfsDvwMMSxJ` during at least one of the voting period's reward cycles. To vote, she sends 5500 satoshis for **yes** to `11111111111111X6zHB1bPW6NJxw6`. Then, her 100,000 STX are tabulated as "yes".
+
+#### Voting "no" as a pool Stacker
+
+Suppose Bob has Stacked 1,000 STX in a Stacking pool and wants to vote "no", and suppose it remains locked in PoX during at least one reward cycle in the voting period. Suppose his Stacks address is `SP2REA2WBSD3XMVMYS48NJKS3WB22JTQNB101XRRZ`. To vote, he sends 1 uSTX from `SP2REA2WBSD3XMVMYS48NJKS3WB22JTQNB101XRRZ` for no to `SP00000000000000DSQJTCSE63RMXHDP`. Then, his 1,000 STX are tabulated as "no."
+
```
Hero-Gamer, line 499: Need to add "Activation Status" to log the voting results, similarly to [SIP-015 Stacks 2.1](https://github.com/stacksgov/sips/blob/main/sips/sip-015/sip-015-network-upgrade.md).
Below text is what I recommend: 

**Activation Status**
At the end of cycle 79, the following vote was calculated. A total of 177,166,148 STX participated.

* For solo stacking, 100% voted 'Yes.' Total voting power is 159,087,747 STX balance with votes cast from 28 accounts.
(Of the 159 million solo stacking voting power, StackingDAO’s internal voting contributed over 19 million STX in voting power.
    _-Total number of STX balance used to vote Solo Stacking by StackingDAO: 19,370,000.
    -6,827 StackingDAO accounts took part in the vote: 6,681 accounts voted in favor, 97.86%, 146 StackingDAO accounts voted against, 2.14%._)

* For pool stacking, 100% voted 'Yes.' Total voting power is 16,524,829 STX balance with votes cast from 370 accounts.

* For non-stackers, 99.98% voted 'Yes.' Total voting power of Yes is 1,553,301 STX balance with votes cast from 2,303 accounts. Total voting power of No is 271 STX balance from 16 accounts.



All voting criteria from STX holders have been met. A breakdown of the transactions can be found [here](https://stx.eco/dao/proposals/SP3JP0N1ZXGASRJ0F7QAHWFPGTVK9T2XNXDB908Z.bdp001-sip-021-nakamoto/results).

A copy of the scripts used to tabulate the solo and pool stacking can be found [here](https://github.com/stacksgov/sips/tree/main/sips/sip-021/nakamoto-votes.zip).

Hero-Gamer (commented):
Logging Nakamoto Community Voting Results

sips/sip-021/sip-021-nakamoto.md
Hero-Gamer: If this gets approved, is it ready to merge into main repo? @whoabuddy @wileyj

obycode: I just added your text to the bottom of the file 👍

wileyj (approved):

jiga (approved):

Hero-Gamer (approved):

PR #168, opened 2024-01-26
docs: minor updates to improve the SIP

obycode:
@kantai I addressed that paragraph we discussed here, as well as cleaned up some other minor things. I added a TODO to fill in the voting process for non-stackers.

I also have one remaining question about whether cycle 75 is intended to be the number used in the "For Miners" section.


friedger (approved):
LGTM

PR #178, opened 2024-05-07
Update for get-block-info in SIP-021

friedger:
This PR adds an Addendum to SIP 021-Nakamoto v1 that describes the impact of Fast Blocks to the clarity function `get-block-info?`

The function should be replaced by `get-stacks-block-info?` and `get-tenure-info?`

### DIFF
--- a/sips/sip-021/sip-021-nakamoto.md
+++ b/sips/sip-021/sip-021-nakamoto.md
@@ -37,7 +37,7 @@ In this proposal, Stacks block production would no longer be tied to miner elect
 This proposal, dubbed the "Nakamoto" release, represents a substantial architectural change to the current Stacks blockchain. If adopted, the Stacks major version would be bumped from 2 to 3. The first Nakamoto release would be 3.0.0.0.0.
 
 # Addendum
-_The following was added after this SIP was accepted, where some version number changes were necessary. The following section addresses these changes **without** changing the ratified text_
+_The following was added after this SIP was accepted, where some clarification about Clarity specifications were necessary. The following section addresses these changes **without** changing the ratified text_
 
 The introduction of Fast Blocks and of the new Clarity variables `tenure-height` and `stacks-block-height` in this SIP requires that the existing Clarity function `get-block-info?` is changed. The function should be removed in Clarity 3, instead two new functions should be added that retrieves data for stacks blocks and for tenure blocks:
 * `get-stacks-block-info?`
--- a/sips/sip-021/sip-021-nakamoto.md
+++ b/sips/sip-021/sip-021-nakamoto.md
@@ -44,7 +44,7 @@ The introduction of Fast Blocks and of the new Clarity variables `tenure-height`
   * `id-header-hash`: equivalent to Clarity 2's `(get-block-info? id-header-hash block-height)`
   * `header-hash`: equivalent to Clarity 2's `(get-block-info? header-hash block-height)`
   * `time`: **new** in Clarity 3, this property returns a `uint` value matching the time field in the Stacks block header
-* `(get-tenure-info? property-name tenure-height)`, where `property-name` is one of:
+* `(get-tenure-info? property-name block-height)`, where `property-name` is one of:
   * `burnchain-header-hash`: equivalent to Clarity 2's `(get-block-info? burnchain-header-hash block-height)`
   * `miner-address`: equivalent to Clarity 2's `(get-block-info? miner-address block-height)`
   * `time`: equivalent to Clarity 2's `(get-block-info? time block-height)`, returns the value of the burn chain block header time field of the tenure block
--- a/sips/sip-021/sip-021-nakamoto.md
+++ b/sips/sip-021/sip-021-nakamoto.md
@@ -36,6 +36,22 @@ In this proposal, Stacks block production would no longer be tied to miner elect
 
 This proposal, dubbed the "Nakamoto" release, represents a substantial architectural change to the current Stacks blockchain. If adopted, the Stacks major version would be bumped from 2 to 3. The first Nakamoto release would be 3.0.0.0.0.
 
+# Addendum
+_The following was added after this SIP was accepted, where some version number changes were necessary. The following section addresses these changes **without** changing the ratified text_
+
+The introduction of Fast Blocks and of the new Clarity variables `tenure-height` and `stacks-block-height` in this SIP requires that the existing Clarity function `get-block-info?` is changed. The function should be removed in Clarity 3, instead two new functions should be added that retrieves data for stacks blocks and for tenure blocks:
+* `get-stacks-block-info?`
+  * id-header-hash: as `(get-block-info? id-header-hash bh)`.
+  * header-hash: as `(get-block-info? header-hash bh)`.
+  * time: **new** in Clarity 3, this property returns a `uint` value of the stacks block header time field. 
+* `get-tenure-info?`
+  * burnchain-header-hash: as `(get-block-info? burnchain-header-hash bh)`.
+  * miner-address: as `(get-block-info? miner-address bh)`.
+  * time: as `(get-block-info? time bh)`, returns the value of the burn chain block header time field of the tenure block.
+  * block-reward: as `(get-block-info? time bh)`.
+  * miner-spend-total: as `(get-block-info? miner-spend-total bh)`.
+  * miner-spend-winner: as `(get-block-info? miner-spend-winner bh)`.
+
 # Introduction
 
 ## Glossary
--- a/sips/sip-021/sip-021-nakamoto.md
+++ b/sips/sip-021/sip-021-nakamoto.md
@@ -39,19 +39,19 @@ This proposal, dubbed the "Nakamoto" release, represents a substantial architect
 # Addendum
 _The following was added after this SIP was accepted, where some clarification about Clarity specifications were necessary. The following section addresses these changes **without** changing the ratified text_
 
-The introduction of Fast Blocks and of the new Clarity variables `tenure-height` and `stacks-block-height` in this SIP requires that the existing Clarity function `get-block-info?` is changed. The function should be removed in Clarity 3, instead two new functions should be added that retrieves data for stacks blocks and for tenure blocks:
-* `get-stacks-block-info?`
-  * id-header-hash: as `(get-block-info? id-header-hash bh)`.
-  * header-hash: as `(get-block-info? header-hash bh)`.
-  * time: **new** in Clarity 3, this property returns a `uint` value of the stacks block header time field. 
-* `get-tenure-info?`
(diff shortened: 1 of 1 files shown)


wileyj (dismissed):
looks good to me, i don't see any logical issues with what's proposed as addendum here

friedger:
Is `id-header-hash` still meaningful on the stacks block level as the stack chain does not fork anymore on that level?

friedger:
Is `id-header-hash` still meaningful on the stacks block level as the stack chain does not fork anymore on that level?

wileyj:
> Is `id-header-hash` still meaningful on the stacks block level as the stack chain does not fork anymore on that level?

very good question - @obycode any insight here?

obycode:
Since this add-on is for minor changes to Clarity, let's also add removal of `index-of` and `element-at`, forcing the use of the variants ending in `?`. What do you think?

obycode:
> very good question - @obycode any insight here?

Copying the responses from Slack here:
friedger: The block hash is the stacks block hash. The consensus hash from tenure?
Aaron: yep -- it's the tenure's consensus hash
Aaron: so, even if the tenure spanned multiple bitcoin blocks (due to a tenure-extend), the same consensus hash would be used for the id-header-hash

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -36,6 +36,23 @@ In this proposal, Stacks block production would no longer be tied to miner elect
...
 This proposal, dubbed the "Nakamoto" release, represents a substantial architectural change to the current Stacks blockchain. If adopted, the Stacks major version would be bumped from 2 to 3. The first Nakamoto release would be 3.0.0.0.0.
 
+# Addendum
+_The following was added after this SIP was accepted, where some clarification about Clarity specifications were necessary. The following section addresses these changes **without** changing the ratified text_
+
+The introduction of Fast Blocks and of the new Clarity variables `tenure-height` and `stacks-block-height` in this SIP requires that the existing Clarity function `get-block-info?` is changed. The function should be removed in Clarity 3, instead two new functions should be added that retrieves data for stacks blocks and for tenure blocks:
+* `get-stacks-block-info?`
+  * id-header-hash: as `(get-block-info? id-header-hash bh)`.
+  * header-hash: as `(get-block-info? header-hash bh)`.
+  * time: **new** in Clarity 3, this property returns a `uint` value of the stacks block header time field.
```
obycode, line 46:
```suggestion
* `(get-stacks-block-info? property-name block-height)`, where `property-name` is one of:
  * id-header-hash: identical to Clarity 2's `(get-block-info? id-header-hash block-height)`
  * header-hash: identical to Clarity 2's `(get-block-info? header-hash block-height)`
  * time: **new** in Clarity 3, this property returns a `uint` value matching the time field in the Stacks block header
```

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -36,6 +36,23 @@ In this proposal, Stacks block production would no longer be tied to miner elect
...
+  * header-hash: as `(get-block-info? header-hash bh)`.
+  * time: **new** in Clarity 3, this property returns a `uint` value of the stacks block header time field. 
+* `get-tenure-info?`
+  * burnchain-header-hash: as `(get-block-info? burnchain-header-hash bh)`.
+  * miner-address: as `(get-block-info? miner-address bh)`.
+  * time: as `(get-block-info? time bh)`, returns the value of the burn chain block header time field of the tenure block.
+  * block-reward: as `(get-block-info? time bh)`.
+  * miner-spend-total: as `(get-block-info? miner-spend-total bh)`.
+  * miner-spend-winner: as `(get-block-info? miner-spend-winner bh)`.
+  * vrf-seed: as `(get-block-info? vrf-seed bh)`
```
obycode, line 54:
```suggestion
* `(get-tenure-info? property-name tenure-height)`, where `property-name` is one of:
  * burnchain-header-hash: equivalent to Clarity 2's `(get-block-info? burnchain-header-hash block-height)`.
  * miner-address: equivalent to Clarity 2's `(get-block-info? miner-address block-height)`.
  * time: equivalent to Clarity 2's `(get-block-info? time block-height)`, returns the value of the burn chain block header time field of the tenure block.
  * block-reward: equivalent to Clarity 2's `(get-block-info? time block-height)`.
  * miner-spend-total: equivalent to Clarity 2's `(get-block-info? miner-spend-total block-height)`.
  * miner-spend-winner: equivalent to Clarity 2's `(get-block-info? miner-spend-winner block-height)`.
  * vrf-seed: equivalent to Clarity 2's `(get-block-info? vrf-seed block-height)` 
```

sips/sip-021/sip-021-nakamoto.md
```diff
@@ -36,6 +36,23 @@ In this proposal, Stacks block production would no longer be tied to miner elect
 
 This proposal, dubbed the "Nakamoto" release, represents a substantial architectural change to the current Stacks blockchain. If adopted, the Stacks major version would be bumped from 2 to 3. The first Nakamoto release would be 3.0.0.0.0.
 
+# Addendum
```
obycode, line 39: I think this addendum section can go at the bottom instead of at the top.

obycode:
At today's WG sync, we discussed the value in minimizing changes to the function as `get-block-info?` instead of removing it and adding `get-stacks-block-info?`, with the goal of minimizing unnecessary changes for developers. Let's continue that discussion here.

If we left `get-block-info?` as-is, we would just be modifying the `time` property to return the new nakamoto time instead of the burn-block time. The rest of the properties could still return the values from the current tenure. If we did that, I would still want to add `get-tenure-info?` which takes in a tenure height and includes the properties mentioned in this PR.

friedger:
My concern is that newly onboarded developers (after Nakamoto) get confused about the properties and what they mean. The Nakamoto upgrade should be celebrated as the better solution.
We can provide a tool/add-on for clarinet that converts Clarity 2 code to Clarity 3 code.

obycode:
There was some more discussion about this and there seems to be a preference to not add new functions, but instead just add a new property to `get-block-info?` for the new timestamp.

Just to have everything in one place for ease of discussion, here are the current properties for `get-block-info?`:
- `burnchain-header-hash`
- `id-header-hash`
- `header-hash`
- `miner-address`
- `time`
- `vrf-seed`
- `block-reward`
- `miner-spend-total`
- `miner-spend-winner`

We have one new property to add, which is a timestamp included in the header of each Nakamoto block.

I can see how some of these properties could be confusing to users if we leave `get-block-info?` as-is. For example, they may assume that the `block-reward`, `miner-spend-total`, and `miner-spend-winner` are meaningful for each block. Two options to avoid this bad assumption:
1. only include non-zero values for those properties in the block with the tenure change
2. move them to a separate, tenure-specific function as suggested in this PR

Thinking through all of this, my current preference is:
1. Add a `stacks-time` property to `get-block-info?` to return the new timestamp
2. Modify the way `block-reward`, `miner-spend-total`, and `miner-spend-winner` are set in Nakamoto so that they are only non-zero on blocks with a tenure change.

I still have a concern about `vrf-seed`. Might users think they can expect a unique VRF seed with each block? Should we return a `none` if it's not a tenure change block? If we return `none`, then you can't differentiate between a bad `block-height` and a block that is not a tenure change.

@friedger, @jcnelson, @kantai please add your thoughts.

friedger:
> I still have a concern about vrf-seed. Might users think they can expect a unique VRF seed with each block? 

Yes, because that was the case until now (if you ignore microblocks). Can we create something random out of the signers signatures? 

I like the solution that the function returns none if not tenure height. To use the function I would need to use `tenure-height`, or `(- tenure-height N)` for the Nth previous vrf seed, correct ?

obycode:
Just had another chat about this, and I think we came full circle back around to your original proposal 😅.

obycode:
Do you think calling `(get-stacks-block-info? time 1234)` where `1234` is a pre-epoch 3.0 block should return the burnchain time or should it return `none`?

friedger:
Returning burnchain time sounds correct

jcnelson:
Agree with @friedger -- the timestamp of a Stacks block pre-Nakamoto is the burnchain time.

obycode:
I think we may also need to add a new function `(get-tenure-for-block? block-height)` which returns an `(optional uint)` which is the tenure height of the specified block if `block-height <= tip-height`. The other option would be `(get-tenure-info-for-block? block-height)` which functions like `get-tenure-info?` except it takes in a block height instead of a tenure height and has an additional property, `height`, which returns the tenure height at the specified block.

obycode:
Based on discussion from today's call, I will change the implementation of `get-tenure-info?` to take in a burn block height as its parameter. `tenure-height` can be a new additional property.

obycode:
Another change here... `get-tenure-info?` will take a Stacks block height as its parameter. Implemented in stacks-network/stacks-core#4846

314159265359879 (approved):
Valuable minor clarifications. LGTM!

AcrossfireX:
Acrossfire (SIP Editor) - signing off for SIP Editors on this change. This has been discussed with the community in relevant community conversations already so the community should be well positioned to understand this.

sips/sip-021/sip-021-nakamoto.md
friedger: The only Addendum I could find was on top of the document. I am fine with both.

jiga: I recommend moving addendum to the end of ratified SIP.

whoabuddy:
I missed this when approving/merging SIP-021 in #155.

Should we point the change here to the main branch now to include the changes?

Should this be revised to the end of the document per @jiga's comment?
https://github.com/stacksgov/sips/pull/178/files#r1662687245

Anything else that needs to be done to close this out?

wileyj:
> Should we point the change here to the main branch now to include the changes?

that probably makes the most sense and can be done without any change from PR submitter. 

i think the location is preference only, but there is precedence for adding it after the abstract section: https://github.com/stacksgov/sips/blob/main/sips/sip-022/sip-022-emergency-pox-fix.md?plain=1

obycode:
I changed the target branch. I do think it makes more sense to put this addendum at the end. Putting it at the beginning seems more appropriate for something that is more significant and deserves more attention.

wileyj:
> I changed the target branch. I do think it makes more sense to put this addendum at the end. Putting it at the beginning seems more appropriate for something that is more significant and deserves more attention.

@friedger is that a change  you think you can make here?

sips/sip-021/sip-021-nakamoto.md
obycode: I just pushed a commit moving this to the end.

wileyj (approved):

obycode (approved):

PR #181, opened 2024-05-16
SIP voting results

radicleart:
This PR adds the tabulated Nakamoto community voting results as zip of csv files. There are three csv files enclosed for solo, pool and dao voting results respectively. The counts were tabulated from the pox-3 clarity contract data and event stream.


obycode (approved):
LGTM. Thanks!

PR #183, opened 2024-06-05
SIP-025: Iterating towards WSTS

jcnelson:
This PR proposes SIP-025, which is a simpler and more robust but less space-efficient signature scheme for Nakamoto block headers.  It is a rider SIP, meaning that it will only activate if SIP-021 activates and some additional criteria are met.


BowTiedSirJonathan:
We understand the necessity of this change, having integrated the WSTS implementation for signing in decentralized mining pools. In our case, a faulty coordinator means no participation in the mining process on that pool, allowing users to switch to another pool. However, in the PoX mechanism, a faulty coordinator significantly impacts block finality, as no signatures are produced during that cycle. Addressing this issue is crucial. The two-iteration proposal is practical: the first iteration allows for quicker implementation with minimal impact aside from block size. The second iteration, while the ideal solution, may take too long for Nakamoto 3.0 at this stage. Therefore, implementing it after the first iteration makes sense.

Additionally, although it may be outside the scope of this SIP, it seems necessary to penalize faulty coordinators more than faulty signers, as they disrupt the flow significantly. Penalties could include suspending their participation for a certain number of cycles or allowing them to stack but burning their rewards instead of transferring them. The severity of penalties should escalate with repeated faults. Regarding @crypt0jan point about the ranking system, there is concern that someone with a consistently higher ranking may not want to be the coordinator in most cycles.

BowTiedSirJonathan (commented):
We understand the necessity of this change, having integrated the WSTS implementation for signing in the decentralized mining pools. In our case, a faulty coordinator means no participation in the mining process on that pool, allowing users to switch to another pool. However, in the PoX mechanism, a faulty coordinator significantly impacts block finality, as no signatures are produced during that cycle. Addressing this issue is crucial. The two-iteration proposal is practical: the first iteration allows for quicker implementation with minimal impact aside from block size. The second iteration, while the ideal solution, may take too long for Nakamoto 3.0 at this stage. Therefore, implementing it after the first iteration makes sense.

Additionally, although it may be outside the scope of this SIP, it seems necessary to penalize faulty coordinators more than faulty signers, as they disrupt the flow significantly. Penalties could include suspending their participation for a certain number of cycles or allowing them to stack but burning their rewards instead of transferring them. The severity of penalties should escalate with repeated faults. Regarding @crypt0jan point about the ranking system, there is concern that someone with a consistently higher ranking may not want to be the coordinator in most cycles.

AcrossfireX:
This SIP is clear on its reasons and also the context of the plan to implement full WSTS in the future. 

Acrossfire (SIP Editor) - putting forward SIP Editor approval

saralab:
This SIP has been approved by the 
- Technical CAB: https://github.com/stacksgov/sips/blob/main/considerations/minutes/technical-cab/2024-06-25-sip-WSTS.md?plain=1
- All public Signers: https://stacks.org/nakamoto-signer-vote

This can be merged now. 
@crypt0jan @BowTiedSirJonathan : Can you verify and approve?

jcnelson:
Need to switch the text to "Ratified" first.

jcnelson:
@obycode can you approve this when you get a moment? :pray:

diwakergupta (approved):

jcnelson:
cc @obycode need your approval as well before I can merge

obycode (approved):

PR #203, opened 2024-12-06
SIP-03x: Fix `burn-block-height` keyword inside of `at-block`

jbencin:
### Description

This SIP fixes the behavior of `burn-block-height` inside of an `at-block` context

### Related Issues

- Bug report: hirosystems/clarinet#1615
- Implementation: stacks-network/stacks-core#5524


314159265359879:
I am closing this because `at-block` was removed in epoch 3.4 with SIP042 https://github.com/stacksgov/sips/pull/262

PR #211, opened 2025-04-04
SIP-032: Improved stacking

friedger:
This PR adds a SIP specifying 
* how to remove the cooldown cycle.
* how to streamline solo stacking and delegated stacking
* how to improve the relationship between signers and stackers


sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,148 @@
...
+
+The prepare cycle length defined in PoX contract has be changed to
+
+```
+(define-constant PREPARE_CYCLE_LENGTH (if is-in-mainnet u10 u5))
+```
+
+## Relationship Between Stackers and Signers
+
+The signature provided by a signers to a stackers shall use one of the following topics
```
314159265359879, line 126: do you mean?:
...a signers... --> signers
...a stackers... ---> stackers

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,148 @@
...
+- delegate
+- delegate-increase
+- delegate-extend
+- agg-commit
+- agg-increase
+
+## Transition to PoX-5
+
+A new PoX contract requires that all stacked Stacks tokens are unlocked and Stackers need to lock their Stacks token again using the new PoX-5 contract. The process shall be similar to the previous upgrades of the PoX contract. PoX-4 contract shall be deactivated and PoX-5 contract shall be activated at the beginning of epoch 3.2 All locked Stacks tokens shall be unlocked automatically 1 block after the beginning of epoch 3.2.
+
```
314159265359879, line 137: I think this section should be a bit more elaborate because it should cover potential complexities/risks related to sBTC and its dependence on PoX.

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,148 @@
...
+# Related Work
+
+The previous PoX process is described in [SIP 007](https://github.com/stacksgov/sips/blob/main/sips/sip-007/sip-007-stacking-consensus.md).
+
+# Activation
+
+This SIP requires are hard fork and shall be activated on Stacks 3.2 as defined by the SIP for epoch 3.2.
```
314159265359879, line 144: requires are hard fork --> requires a hard fork

314159265359879 (commented):
Thanks for adding this Friedger! I look forward to the proposed changes going live.

Added some minor comments about typos. 

In the SIP I would also like to read how a transition from PoX-4 to PoX-5 would influence the sBTC protocol and how potential risks or other complexities are mitigated and/or dealt with.

sips/sip-x/sip-x-improved-stacking.md
friedger: sBTC is not involved here.

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Semantic Change of Delegation
+
+When signers verify and accept proposed blocks by miners, their voting power corresponds to the amount of stacked Stacks tokens (see SIP-021). The new stacking process changes the delegation flow as follows:
+
+- `delegate-stx` shall be renamed to `delegate`. The function takes the arguments: `amount`, `delegate-to`, `signature`, optional bitcoin `block-height` defining the end of stacking and auto extending, optional `max-amount`. The `amount` defines how many Stacks tokens are locked immediately. The `max-amount` the maximum of Stacks tokens that can be locked in the future through auto extending. If omitted the whole balance is locked. The argument `delegate-to` describes the signer. The argument `signature` is a signature of the signer indicating that the signer accepted the delegation of voting power.
+
+- `revoke-delegate-stx` shall be renamed to `revoke-delegate`. After calling this functions, auto extension is stopped and Stacks tokens are unlocked for the user at the end of the current cycle.
+
+- `delegate-increase` is replaced by `delegate`. Users can simply call delegate with a higher amount.
```
friedger: We don't want to allow decrease, do we? As it would potentially imply, unlocks for tiny amounts.

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,205 @@
...
+# Abstract
+
+This SIP proposes a change to the stacking process so that stackers can change their stacking settings without the so-called cooldown cycle. Furthermore, the relationship between stackers
+and signers is strengthened and stacking overall is simplified.
+
+The specification defines that
+
+- solo stackers and delegating stackers have to follow the same flow and use the same contract functions: All users delegate block voting power to signers.
+- stackers can change their stacking settings for the next cycle before the prepare phase, including the chosen signer (switch pools).
+- locked Stacks token are locked for 1 cycle at a time and that the locking period is extended by another cycle if the user does not request to unlock tokens.
```
314159265359879: locked Stacks tokens...

plural (tokens instead of token)

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,205 @@
...
+- delegated stacking tokens are locked immediately.
+- stacking rewards are received by the Bitcoin address specified by the signer.
+- locking Stacks tokens is protected by a new type of post conditions, enabling stacking through a contract in a single transaction.
+
+# Introduction
+
+## Current Situation
+
+Currently, the stacking protocol has a few aspects that make using and integration stacking harder than it could be:
```
314159265359879: ...make using and integrating...

(integrating instead of integration)

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,205 @@
...
+The current stacking process has two different paths, one for solo stackers, one for delegated stackers,
+resulting in a more complex user experience and more complex code.
+
+The process is defined by a prepare phase of 100 bitcoin blocks that is used to find an anchor block
+for the next stacking cycle. Stackers must lock
+their Stacks tokens before the
+prepare phase, i.e. before the 2000th block of the current stacking cycle. Furthermore, Stacks
+tokens are locked for a locking period that always ends at the beginning of a cycle, i.e. after the
+prepare phase. Therefore, the current implementation of Stacking includes a period where Stacks
+tokens are unlocked and cannot earn stacking rewards (Cooldown Cycle).
```
314159265359879: ...includes a period during which Stacks tokens are unlocked and ineligible for stacking rewards (Cooldown Cycle).

(suggestion to improve clarity)

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,205 @@
...
+
+Furthermore, users cannot change their
+stacking settings (decrease amount, change PoX reward address, etc.) without an unlocked cycle during which users cannot earn
+stacking rewards.
+
+## Proposed Solution
+
+This SIP proposes a new Proof of Transfer (PoX) contract without
+the flow for solo stacking and moves responsibilities from pool operators to signers who were introduced in the Nakamot upgrade. It also
+defines a new locking mechanism that allows users to switch from one signer to another with a single contract call and without a cool down period.
```
314159265359879: cooldown

(instead of _cool down_)

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,205 @@
...
+
+- `delegate-stx` shall be renamed to `delegate`. The function takes the arguments: `amount`, `delegate-to`, `signature`, optional bitcoin `block-height` defining the end of stacking and auto extending, optional `max-amount`. The `amount` defines how many Stacks tokens are locked immediately. The `max-amount` the maximum of Stacks tokens that can be locked in the future through auto extending. If omitted the whole balance is locked. The argument `delegate-to` describes the signer. The argument `signature` is a signature of the signer indicating that the signer accepted the delegation of voting power.
+
+- `revoke-delegate-stx` shall be renamed to `revoke-delegate`. After calling this functions, auto extension is stopped and Stacks tokens are unlocked for the user at the end of the current cycle. If the signer the user delegated to did not aggregated enough Stacks tokens to receive at least one reward slot then the user's Stacks token are unlocked immediately through this call.
```
314159265359879: ...did not aggregate enough...

(instead of _aggregated_)

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,205 @@
...
+
+1. one for signing blocks and delegation approval;
+2. one for PoX reward address and reward distribution;
+3. one for aggregation on the Stacks blockchain.
+
+The Stacks address of the first key is used by Stacks holders during delegation for `delegate-to`.
+
+The encoded bitcoin address of the second key and the Stacks address of the third key are announced in a registration contract call. The new function of the PoX contract is named `signer-register` and just takes the encoded bitcoin address, and the stacks address as parameters.
+
+Note, for solo stackers these three keys can be just a single key. For more complex setup, the keys can be handled by different independant entities. Also, the PoX reward address can be a deposit address for sBTC.
```
314159265359879: ...independent...

(instead of _independ**a**nt_)

314159265359879 (commented):
I just added a couple nits, mostly typos and two suggestion to improve clarity/readability.

sips/sip-x/sip-x-improved-stacking.md
friedger: :+1:

sips/sip-x/sip-x-improved-stacking.md
friedger: :+1:

sips/sip-x/sip-x-improved-stacking.md
friedger: :+1:

sips/sip-x/sip-x-improved-stacking.md
friedger: :+1:

sips/sip-x/sip-x-improved-stacking.md
friedger: :+1:

sips/sip-x/sip-x-improved-stacking.md
friedger: :+1:

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,205 @@
...
+the switching costs are too high for users to switch (two weeks' worth of yield).
+
+Furthermore, users cannot change their
+stacking settings (decrease amount, change PoX reward address, etc.) without an unlocked cycle during which users cannot earn
+stacking rewards.
+
+## Proposed Solution
+
+This SIP proposes a new Proof of Transfer (PoX) contract without
+the flow for solo stacking and moves responsibilities from pool operators to signers who were introduced in the Nakamot upgrade. It also
```
rafaelcr: Small typo
```suggestion
the flow for solo stacking and moves responsibilities from pool operators to signers who were introduced in the Nakamoto upgrade. It also
```

nieldeckx:
Really liking this upgrade!

One ask from StackingDAO: please give plenty of time before the PoX-5 change goes live. We’ll need to update our code and get an audit, which takes a while. A clear schedule for testnet and mainnet would be nice.

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,206 @@
...
+
+- `delegate-stx` shall be renamed to `delegate`/`designate`. The function takes the arguments: `amount`, `signature`, optional bitcoin `block-height` defining the end of stacking and auto extending, optional `pox-addr` defining that pox-address that the signer must use, optional `max-amount`. The `amount` defines how many Stacks tokens are locked immediately. The `max-amount` defines the maximum of Stacks tokens that can be locked in the future through auto extending. If provided the minimum of the user's stx balance and `max-amount` is locked. If omitted, the whole user's balance is locked. Users can use sponsored transactions to call revoke-delegate-stx in case there is no unlocked stx in the account. The argument `signature` is a signature of the signer indicating that the signer accepted the delegation of voting power. The signing follows the structured message signing with the parameters and topic `designate` as message. The public key of the signature can be used to identify the signer similar to the Stacks address of the pool operator in the previous PoX design.
```
djordon: It wasn't immediately clear why we were having both and `amount` and a `max-amount`. Here is my understanding:

The design of `amount` and `max-amount` allows me to make a `delegate` contract call now and have my STX frozen for stacking later. It also makes the arguments of the `delegate-update-amount` function consistent with the `delegate` function. The purpose of the `amount` variable is to make it more difficult for me to _not_ have enough STX in my account when it becomes "officially" frozen through stacking. Is this a fair characterization of their purpose here?

djordon: One thing of note with this design is that it makes it difficult to know how much was stacked by a particular user. I can make a `delegate` contract call with an amount of zero (or some other low value) and a max-amount of None and it becomes tough to know how much was actually stacked by the user. The stacks node can figure it all out, but this SIP doesn't propose that the node emits events when the amount gets finalized.

I'm thinking that it would be better to remove the `max-amount` parameter from this function, but still have a `max-amount` around in the background, the same one used in `delegate-update-amount`. The `delegate` function would just set the `max-amount` to `amount`. If the user wants to change the amount during the appropriate time periods, then they call `delegate-update-amount` and the corresponding amounts are locked or unlocked immediately upon confirmation of a successful contract call.

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,206 @@
...
+The following function shall be removed:
+
+- `delegate-stack-extend`
+
+In addition, the stacking settings for each Stacker (user with delegation) shall be applied for the next stacking cycle if the user did not signal the end of stacking by calling `revoke-delegate-stx` 200 blocks before the start of the next cycle.
+
+This results in the following structure of the stacking cycle:
+
+- bitcoin block 1-1900: stacking as usual, user can signal change of stacking settings.
+- bitcoin block 1901-2000: stacking as usual, signers can still aggregate stacking changes.
```
djordon: If a user tries to signal a change in their stacking settings, are is it possible for it to apply to the next cycle? Whatever the answer is, can we add it to this SIP?

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,206 @@
...
+
+- `revoke-delegate-stx` shall be renamed to `revoke-delegate`/`revoke-designate`. After calling this function, auto extension is stopped and Stacks tokens are unlocked for the user at the end of the current cycle. If a signer does not aggregate enough Stacks tokens to receive at least one reward slot (or does not commit at all) then the user's Stacks token are unlocked immediately through this call.
+
+- `delegate-increase` is replaced by `delegate-update-amount`/`designate-update-amount`. It sets max-amount to the new value. The amount can be smaller than the currently locked amount. As tokens are only locked for 1 cycle at a time, handling decreasing is now easy enough in comparison to the previous system with locking periods of up to 12 cycles.
```
djordon: Can we change the name here to `update-delegated-amount`?

djordon: Also, one thing that isn't specified here is whether, on an increase of the `max-amount` from the previous value, whether the STX gets locked immediately.

sips/sip-x/sip-x-improved-stacking.md
```diff
@@ -0,0 +1,206 @@
...
+
+- `delegate-stack-extend`
+
+In addition, the stacking settings for each Stacker (user with delegation) shall be applied for the next stacking cycle if the user did not signal the end of stacking by calling `revoke-delegate-stx` 200 blocks before the start of the next cycle.
+
+This results in the following structure of the stacking cycle:
+
+- bitcoin block 1-1900: stacking as usual, user can signal change of stacking settings.
+- bitcoin block 1901-2000: stacking as usual, signers can still aggregate stacking changes.
+- bitcoin block 2001-2100 (prepare phase): no rewards for Stackers, changes to stacking are applied to the cycle after next.
```
djordon, line 131: Would it be harmful to allow the signers to aggregate stacking changes in the prepare phase?

sips/sip-x/sip-x-improved-stacking.md
djordon: > One thing of note with this design is that it makes it difficult to know how much was stacked by a particular user.

This may be equally true for the current pox-4 design, so maybe one of my concerns above isn't such a big deal.

friedger: The idea was that users don't have to interact every cycle and that they still can autocompound when the distribution contract swaps sbtc to stx.

I'll add a section about events. 

The difficulty to find out the stacked amount in a contract should be addressed as well. The max amount does not really change it.

sips/sip-x/sip-x-improved-stacking.md
friedger: My idea for max amount only applies to the auto extend

sips/sip-x/sip-x-improved-stacking.md
friedger: Yes, because they need to find the anchor block.

My initial proposal to reduce the length of the prepare phase was rejected. The change would be too big.

sips/sip-x/sip-x-improved-stacking.md
friedger:
```suggestion
- bitcoin block 1901-2000: stacking as usual, signers can still aggregate stacking changes, signalled changes by users will be applied only for the next cycle.
```

sips/sip-x/sip-x-improved-stacking.md
friedger: I have added a section about how to find the stacked amount.

sips/sip-x/sip-x-improved-stacking.md
friedger: A better name would be `update-max-amount`. I added that and clarified when the excess amount is unlocked in case of a decrease.

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+prepare phase, i.e. before the 2000th block of the current stacking cycle. Furthermore, Stacks
+tokens are locked for a locking period that always ends at the beginning of a cycle, i.e. after the
+prepare phase. Therefore, the current implementation of Stacking includes a period where Stacks
+tokens are unlocked and ineligible for stacking rewards (Cooldown Cycle).
+
+The current PoX contract allows users to lock Stacks tokens during the prepare phase. This is too late for the next cycle
+and the tokens are locked without earning stacking rewards for one cycle.
+
+The cooldown cycle for unstacking from a pool presents a problem for network decentralization.
+As a Stacker, when I delegate to a signer and that signer does not perform and gets low yield,
```
jcnelson: Nit: SIPs do not use the first-person.

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+
+This SIP proposes a new Proof of Transfer (PoX) contract without
+the flow for solo stacking and moves responsibilities from pool operators to signers who were introduced in the Nakamoto upgrade. It also
+defines a new locking mechanism that allows users to switch from one signer to another with a single contract call and without a cooldown period.
+
+# Specification
+
+Applying these upgrades to the Stacks blockchain requires a consensus-breaking network upgrade,
+in this case, a hard fork. Like other such changes, this will require a new Stacks epoch.
+In this SIP, we will refer to this new epoch as Stacks 3.2.
```
jcnelson: It'll need to be Stacks 3.3 now that SIP-031 has activated.

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Delegated Stacking only
+
+The following PoX contract functions shall be removed:
+
+- stack-stx
+- stack-increase
+- stack-extend
```
jcnelson, line 109: Instead of removing them, could they instead wrap calls to the delegation functions?  For example, `stack-stx` could be re-implemented to delegate STX to the caller, and have the caller stack and aggregate-commit them.

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Locking Post Conditions
+
+The following type of post conditions shall be added to the current definition in SIP-005 (`TransactionPostCondition`)
+
+- `LockingLimit(PostConditionPrincipal, FungibleConditionCode, u64)`
+
+A transaction (using Deny mode) with this post condition will abort if the locked Stacks tokens of the principal do not satisfy the provided conditions. The logic for the conditions follows that of STX transfer.
```
jcnelson, line 117: Depending on the order in which this SIP and the Clarity 4 SIP (#218) get ratified, it may be useful for #218 to add an asset guard special form for PoX locking (cc @obycode).

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+    2100 : End of cycle
+```
+
+That means the locking period is 1 cycle, with automatic extension for another cycle until the user decides to end stacking.
+
+The amount of stacking STX can increase during the automatic extension up to the minimum of the balance and a user specified maximum amount.
+
+The amount of stacked STX does not decrease during the automatic extension. If the user requested a decrease of the stacked amount, this difference is unlocked at the end of the stacking period.
+
+## Semantic Change of Delegation
```
jcnelson, line 151: I'm generally fine with renaming things.  Would it also be possible to preserve the existing function names and signatures, and have them call the new functions?  The reasoning here is that it would keep existing tool and scripts working.

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+
+- `delegate-stx` shall be renamed to `designate`. The function takes the arguments: `amount`, `signature`, optional bitcoin `block-height` defining the end of stacking and auto extending, optional `pox-addr` defining that pox-address that the signer must use, optional `max-amount`. The `amount` defines how many Stacks tokens are locked immediately. The `max-amount` defines the maximum of Stacks tokens that can be locked in the future through auto extending. If provided the minimum of the user's stx balance and `max-amount` is locked. If omitted, the whole user's balance is locked. Users can use sponsored transactions to call revoke-delegate-stx in case there is no unlocked stx in the account. The argument `signature` is a signature of the signer indicating that the signer accepted the delegation of voting power. The signing follows the structured message signing with the parameters and topic `designate` as message. The public key of the signature can be used to identify the signer similar to the Stacks address of the pool operator in the previous PoX design.
```
jcnelson: Is this the function signature?

```clarity
(define-public (designate
   ;; The amount of uSTX to be locked immediately
   (amount uint)
   ;; Signature of the signer indicating that the signer accepted the delegation of voting power
   (signature (buff 65))
   ;; The Bitcoin block height at which this designation ends, if any.
   ;; N.B. the text says `block-height`, but this is a reserved word.
   ;; I'm using `end-block-height` in its place.
   (end-block-height (optional uint))
   ;; PoX address that the signer must use, to which PoX yield accrues
   (pox-addr (optional { version: (buff 1), hashbytes: (buff 32) }))
   ;; Maximum amount of uSTX that can be locked through future auto-extending
   (max-amount (optional uint))
   
   ;; code goes here
)
```

jcnelson: A few other questions:

* Can you describe the SIP-018 payload that must be signed by the signer?
* How does this method identify the signer public key to use, in order to authenticate `signature`?  Can it be an argument to the function, like it is for `stack-aggregation-commit` and `stack-aggregation-increase`?
* How does this method ensure that signatures cannot be reused?  The current PoX code uses an `auth-id`, which must be included in the signature and which cannot be re-used across signatures from the same signer key.

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+
+- `revoke-delegate-stx` shall be renamed to `revoke-designation`. After calling this function, auto extension is stopped and Stacks tokens are unlocked for the user at the end of the current cycle. If a signer does not aggregate enough Stacks tokens to receive at least one reward slot (or does not commit at all) then the user's Stacks token are unlocked immediately through this call.
+
+- `delegate-increase` is replaced by `update-max-amount`. It sets max-amount to the new value. This value is applied when the user's designation is extended for the next cycle. The amount can be smaller than the currently locked amount. As tokens are only locked for 1 cycle at a time, handling decreasing is now easy enough in comparison to the previous system with locking periods of up to 12 cycles.
+
+- `delegate-extend` is removed because the locking period is automatically extended each cycle.
+
+## Amount of Stacked STX
+
+When a transaction locks or unlocks STX for the user a corresponding event shall be emitted. The following events used in PoX 4 are used:
```
jcnelson: PoX 4 uses the function's name as the event's name (i.e. in the event tuple's `name` field).  Is this mapping from a PoX 5 event to the old PoX 4 event correct?

* `designate` --> `delegate-stx`
* `designate-increase` --> `stacks-aggregation-increase`
* `designate-decrease` --> (this is new, and has no PoX 4 analogue)
* `designate-extend` --> similar to `delegate-stx-extend`, but is now automatic?
* `signer-accept` --> `stacks-aggregation-commit`

Can you also explain what the event payloads will be?

jcnelson: Ah, I see that `stacks-aggregate-commit` is due to be replaced with `signer-register`.  Is `signer-accept` a new function in PoX 5?  I don't see it defined in this document.

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+
+The other topics are no longer used.
+
+Signers handle two or three private keys:
+
+1. one for signing blocks,
+2. one for delegation approval and register transactions on the Stacks blockchain;
+3. optionally, one for PoX reward address and reward distribution.
+
+Note, for solo Stackers these keys can be just a single key. For more complex setup, the keys can be handled by different independent entities. Also, the PoX reward address can be a deposit address for sBTC of a Stacks smart contract. In this case, the signer does not hold a private key for the PoX reward address.
```
jcnelson: In all likelihood, solo stackers also have separate keys for signing, for cold storage STX, and for Bitcoin rewards.

sips/sip-032/sip-032-improved-stacking.md
obycode: Ah, yes. I had that thought at some point, but then forgot to include it. Thanks! I'll add it.

sips/sip-032/sip-032-improved-stacking.md
jcnelson: > Also, the PoX reward address can be a deposit address for sBTC of a Stacks smart contract.

This is very interesting.  Is the taproot address guaranteed to be stable for long enough that it can be used as a PoX address?  My understanding of the sBTC minting process is as follows:

* The user creates a taproot address from two scripts -- a deposit script, and a reclaim script
   * The deposit script contains the WSTS aggregate public key's X coordinate, the maximum BTC fee, and the Stacks principal to which to mint the sBTC (it's `<[8-byte-max-fee][up-to-159-byte-SIP-005-principal]> OP_DROP <signers-x-public-key> OP_CHECKSIG`).
   * The reclaim script, which allows the sender to reclaim the BTC after a timeout (it's `<locktime> OP_CSV <user-supplied-reclaim-script>`).
* The user creates a Bitcoin transaction which sends BTC to the taproot address
* The user informs the Emily API service of the transaction 
* The Emily API service forwards the transaction to the sBTC signers
* The sBTC signers, if possible, will spend the taproot UTXO via the deposit script.  If the signers can't do this in time, then the user can spend it with the reclaim script.

The risk to users in supplying an sBTC deposit taproot address is that their maximum fee value and locktime value will need to remain set for at least one stacking cycle.  The user would need to change their PoX address each time they want to change the maximum fee or advance the locktime.  Can this SIP advise users on how to construct their sBTC deposit addresses for them to be suitable for their desired number of reward cycles?

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Voting Power
+
+The voting power for a signer is the sum of the accepted stacked STX, not the sum of the currently stacked STX because users can have descreased the stacking amount.
+
+## Transition to PoX-5
+
+A new PoX contract requires that all stacked Stacks tokens are unlocked and Stackers need to lock their Stacks token again using the new PoX-5 contract. The process shall be similar to the previous upgrades of the PoX contract. The PoX-4 contract shall be deactivated and the PoX-5 contract shall be activated at the beginning of epoch 3.2. All locked Stacks tokens shall be unlocked automatically one block after the beginning of epoch 3.2. Nevertheless, these tokens will earn rewards until the end of the cycle.
```
jcnelson: Tagging the epoch -- it's now 3.3.

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+A reference implementation for creating signatures is available at ... (TODO). Note, that signers (or previously pool operators) no longer extend the locking period for each Stacker on-chain, instead signers provide a signature for each Stacker off-chain.
+
+A reference implementation for a sidecar for signer node is available at ... (TODO).
+
+# Related Work
+
+The previous PoX process is described in [SIP 007](https://github.com/stacksgov/sips/blob/main/sips/sip-007/sip-007-stacking-consensus.md).
+
+# Activation
+
```
jcnelson, line 307: Tagging the epoch -- it's now 3.3

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+This SIP requires a hard fork and shall be activated on Stacks 3.2, as defined by the SIP for epoch 3.2.
```
jcnelson: I think it's fine if you want to reuse the activation criteria used in SIP-031, btw.

sips/sip-032/sip-032-improved-stacking.md
friedger: > In all likelihood, solo stackers also have separate keys for signing, for cold storage STX, and for Bitcoin rewards.

I inserted "theoretically".

> Also, the PoX reward address can be a deposit address for sBTC of a Stacks smart contract.

My intention was to highlight that this is possible with the new setup. However, explaining details how to implement this, should be part of a different resource, e.g. https://github.com/friedger/clarity-trustless-pool/

sips/sip-032/sip-032-improved-stacking.md
friedger: Updated for 3.3. Waiting for a sip that defines 3.3.

sips/sip-032/sip-032-improved-stacking.md
friedger: fixed

sips/sip-032/sip-032-improved-stacking.md
friedger: fixed

sips/sip-032/sip-032-improved-stacking.md
friedger: fixed

sips/sip-032/sip-032-improved-stacking.md
friedger: fixed

sips/sip-032/sip-032-improved-stacking.md
friedger: `signer-accept` is defined in ## Amount of Stacked STX as event name. The function is `signer-register`.

friedger: I understand, I'll change signer-accept to a function name.

friedger: The events in Pox 5 are similar (-->) in Pox 4
*    designate --> delegate-stx + delegate-stx-stack
*    designate-increase --> delegate-stx-increase but by user
*    designate-decrease --> (this is new, and has no PoX 4 analogue)
*    designate-extend --> similar to delegate-stx-extend, but is now automatic? Correct, however, there is no call from the user
*    signer-accept --> stacks-aggregation-commit

Payloads should be similar to pox-4, I'll spend some time to specify them as well.

sips/sip-032/sip-032-improved-stacking.md
friedger: Signatures should be created in the same way as Pox-4 using public key and auth-id, etc. I'll extend the section about signatures.

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+
+**Title:** Improved Stacking Protocol
+
+**Authors:**
+
+- Friedger Müffke ([friedger@ryder.id](mailto:friedger@ryder.id))
+
+**Consideration:** Technical
+
+**Type:** Consensus
```
wileyj:
```suggestion
**Type:** Consensus (hard fork)
```

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+
+**Created:** 2025-04-01
+
+**License:** BSD 2-Clause
+
+**Sign-off:**
+
+**Discussions-To:**
+
+- [Stacks Forum Discussions](https://forum.stacks.org/t/remove-cool-down-cycle-in-stacking/17899)
```
wileyj:
```suggestion
- https://forum.stacks.org/t/remove-cool-down-cycle-in-stacking/17899
```
style nit, this is not defined in sip-000 so it may be accepted or rejected and resolved.

sips/sip-032/sip-032-improved-stacking.md
wileyj: +1 for backwards compatibility here

sips/sip-032/sip-032-improved-stacking.md
hstove: I think it would be best to include the actual function signature in the SIP. Additionally, we will need `signer-key` as an argument

sips/sip-032/sip-032-improved-stacking.md
```diff
@@ -0,0 +1,232 @@
...
+
+- `delegate-stx` shall be renamed to `designate`. The function takes the arguments: `amount`, `signature`, optional bitcoin `block-height` defining the end of stacking and auto extending, optional `pox-addr` defining that pox-address that the signer must use, optional `max-amount`. The `amount` defines how many Stacks tokens are locked immediately. The `max-amount` defines the maximum of Stacks tokens that can be locked in the future through auto extending. If provided the minimum of the user's stx balance and `max-amount` is locked. If omitted, the whole user's balance is locked. Users can use sponsored transactions to call revoke-delegate-stx in case there is no unlocked stx in the account. The argument `signature` is a signature of the signer indicating that the signer accepted the delegation of voting power. The signing follows the structured message signing with the parameters and topic `designate` as message. The public key of the signature can be used to identify the signer similar to the Stacks address of the pool operator in the previous PoX design.
```
hstove: > The `amount` defines how many Stacks tokens are locked immediately. The `max-amount` defines the maximum of Stacks tokens that can be locked in the future through auto extending. If provided the minimum of the user's stx balance and `max-amount` is locked. If omitted, the whole user's balance is locked. 

If `max-amount` is omitted, the whole user's balance is locked? Or if `amount` is omitted? Does this mean `amount` is now an `optional`?

Also, grammatically, it should be "user's whole balance", and "If provided, ..." (note the comma)

hstove: I totally understand the usage of `max-amount` here, but this differs from the current usage of `max-amount`. Currently, `max-amount` is used for the signer to say "you can't stack any more than this, because I don't want to have too much signing power". In this proposal, `max-amount` is for the _stacker_ to say "you can only lock at most this much".

These are two different things, so we'll need to reconcile this fact. Either we add a new argument for the _stacker's_ intention, or we remove the _current_ usage of `max-amount`.

Additionally, from a UX perspective, we need to think about how `max-amount` is used in signature generation. When referring to the stacker's intention of "you can only lock at most X", we should probably just omit it from signature generation. Otherwise, signers will need to get this information ahead of time for generating signatures.

hstove: After further consideration, I think we need to keep both the old param _and_ the new param, meaning we need to figure out what to call these params. My recommendation is:

- `max-amount`: the same as before. This represents "the max amount a signer will allow a stacker to stack towards their signer key". For most signers, this will remain `u128::max`
- `allowance`: the maximum amount that a stacker is allowing the signer to stack.

This would essentially mean replacing "max-amount" with "allowance" in this paragraph.

hstove: After further reading, I actually don't understand what `max-amount` means here. If I call `designate` with an `amount` of 1000, and a `max-amount` of 2000, what happens? It doesn't seem like `signer-accept` takes an `amount` or anything, so where does `max-amount` ever come into play?

friedger: I have updated this section to clarify what it means when `max-amount` is omitted. In this case, `amount` is used as `max-amount`.

I also renamed `max-amount` to `amount-allowance` to avoid confusion about `max-amount` used during signature generation.

`max-amount`/`amount-allowance` is used during `auto-extend` where the stacking amount can be increased automatically as well.

friedger: I clarified what happens when max-amount is omitted and fixed the grammar.

PR #218, opened 2025-08-14
SIP-033: Clarity 4: high-demand new builtins

obycode:
The Clarity smart contract language powers decentralized applications across a
wide range of domains. Over years of real-world use, developers have encountered
several pain points when interacting with other, sometimes untrusted, contracts.
Securely calling unknown contracts unlocks powerful use cases, but demands
careful safeguards.

This SIP addresses common feedback and requests from builders in the ecosystem.
It proposes new Clarity features to make it easier for developers to write
secure and composable smart contracts. Specifically, it proposes:

1. **A new Clarity function to fetch the hash of a contract's code body.** This
   enables on-chain contract code validation, for example allowing contract A to
   validate that contract B follows a specific template and is therefore safe to
   interact with. This is especially useful for enabling bridges and
   marketplaces to safely and trustlessly support a dynamic set of assets.
2. **A new set of Clarity functions to allow a contract to set post-conditions
   to protect its assets.** These allow a contract to safely call arbitrary
   external contracts (e.g. passed in as traits) while ensuring that if the
   executed code moves assets beyond those specified, the changes will be rolled
   back.
3. **A new Clarity function to convert simple values into `string-ascii`
   values.** This function will enable developers to easily convert values like
   `bool`s and `principal`s into their ASCII string representations,
   facilitating the generation of string-based messages for interacting with
   cross-chain protocols.
> [!NOTE]
> Additional feature added 👇 
4. **A new Clarity keyword to get the timestamp of the _current_ block.** This
   keyword will allow developers to easily access the timestamp of the block
   currently being processed, enabling time-based logic and features in their
   smart contracts. This is especially important for DeFi applications.

### DIFF
--- a/sips/sip-033/sip-033-clarity4.md
+++ b/sips/sip-033/sip-033-clarity4.md
@@ -26,6 +26,7 @@ License: CC0-1.0
 
 Sign-off:
 - Jesse Wiley <jesse@stacks.org>, Chairperson, Technical CAB: https://github.com/stacksgov/sips/blob/main/considerations/minutes/technical-cab/2025-10-06-sip-033.md
+- werner.btc <werner@trustmachines.co>, SIP Editor, Editors
 
 Discussions-To:
 https://forum.stacks.org/t/clarity-4-proposal-new-builtins-for-vital-ecosystem-projects/18266


sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,378 @@
...
+
+**Status:** Draft
+
+**Created:** 2025-06-20
+
+**License:** CC0-1.0
+
+**Sign-off:**
+
+**Discussions-To:** https://github.com/stacksgov/sips
```
wileyj: Probably better to link to a discussion vs the current repo? i think there were some good issues in clarity-lang, or even the forum post: https://forum.stacks.org/t/clarity-4-proposal-new-builtins-for-vital-ecosystem-projects/18266
```suggestion
**Discussions-To:** https://forum.stacks.org/t/clarity-4-proposal-new-builtins-for-vital-ecosystem-projects/18266
```

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,378 @@
...
+**Authors:**
+
+- Adriano Di Luzio <adriano@bitcoinl2labs.com>
+- Brice Dobry <brice@hiro.so>
+- Marvin Janssen <marvin@ryder.id>
+- Jude Nelson <jude@stacks.org>
+
+**Consideration:** Technical
+
+**Type:** Consensus
```
wileyj:
```suggestion
**Type:** Consensus (hard fork)
```

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,378 @@
...
+
+# Introduction
+
+The Clarity smart contract language powers decentralized applications across a
+wide range of domains. Over years of real-world use, developers have encountered
+several pain points when interacting with other, sometimes untrusted, contracts.
+Securely calling unknown contracts unlocks powerful use cases, but demands
+careful safeguards.
+
+This SIP addresses common feedback and requests from builders in the ecosystem.
```
wileyj:
```suggestion
This SIP addresses common feedback and requests from developers in the ecosystem.
```
nit - this is the only use of `builders`. the rest of the document refers to `developers` -  recommended change for consistency.

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,378 @@
...
+It proposes new Clarity features to make it easier for developers to write
+secure and composable smart contracts. Specifically, it proposes:
+
+1. **A new Clarity function to fetch the hash of a contract's code body.** This
+   enables on-chain contract code validation, for example allowing contract A to
+   validate that contract B follows a specific template and is therefore safe to
+   interact with. This is especially useful for enabling bridges and
+   marketplaces to safely and trustlessly support a dynamic set of assets.
```
wileyj:
```suggestion
   interact with. This is especially useful for enabling services, such as bridges and
   marketplaces. to safely and trustlessly support a dynamic set of assets.
```

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,378 @@
...
+4. **A new Clarity keyword to get the timestamp of the _current_ block.** This
+   keyword will allow developers to easily access the timestamp of the block
+   currently being processed, enabling time-based logic and features in their
+   smart contracts. This is especially important for DeFi applications.
+
+# Specification
+
+## Fetching the hash of a contract body: `contract-hash?`
+
+Originally proposed [here](https://github.com/clarity-lang/reference/issues/88).
```
wileyj: meta question, but should hyperlinks belong in a tech spec like this sip?
i would advocate for leaving links unchanged in the source, 
example with  github expanding the link based on the url
```
Originally proposed here: https://github.com/clarity-lang/reference/issues/88
```

or with the md hyperlink title as the full url
```
Originally proposed here: [https://github.com/clarity-lang/reference/issues/88](https://github.com/clarity-lang/reference/issues/88)
```

this can be resolved either way and has no bearing on the text in question - just a style comment that i felt needed to raised. 

edited to show the markdown and not the rendered output

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,378 @@
...
+principal against the granted allowances. If any allowance is exceeded,
+`restrict-assets?` returns an error; otherwise, it returns `ok` with the result
+of the last body expression.
+
+The most important use case of these new builtins is to allow a contract to
+protect its own assets. To ensure that the contract's assets are safe by
+default, the new `as-contract?` expression is similar to the existing
+`as-contract` expression, but will implicitly behave as though there is a
+`restrict-assets?` expression around its body, specifying the contract principal
+as the asset owner. The old `as-contract` is no longer available in Clarity 4.
```
wileyj, line 126: This is interesting  - i think this should be added to the backwards compatibility section. 
specifically that `as-contract` is not available in Clarity 4

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,378 @@
...
+    (define-public (foo)
+      (restrict-assets? tx-sender ()
+        (try! (stx-transfer? u1000000 tx-sender 'ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM))
+      )
+    ) ;; Returns (err -1)
+    (define-public (bar)
+      (restrict-assets? tx-sender ()
+        (+ u1 u2)
+      )
+    ) ;; Returns (ok u3)
```
radubahmata, line 200: ~As restrict-assets? output is (response A int), the function will return (ok 3)~

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,378 @@
...
+    token defined in `contract-id` with name `token-name` from the `asset-owner`
+    of the enclosing `restrict-assets?` or `as-contract?` expression. Note,
+    `token-name` should match the name used in the `define-fungible-token` call
+    in the contract.
+  - **Example**:
+    ```clarity
+    (restrict-assets? tx-sender
+      ((with-ft (contract-of token-trait) "stackaroo" u50))
+      (try! (contract-call? token-trait transfer u100 tx-sender 'ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM none))
+      ) ;; Returns (err 0)
```
radubahmata:
```suggestion
    ) ;; Returns (err 0)
```

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,378 @@
...
+    ) ;; Returns (ok true)
+    ```
+
+- `with-nft`
+
+  - **Input**:
+    - `contract-id`: `principal`: The contract defining the NFT asset.
+    - `token-name`: `(string-ascii 128)`: The name of the NFT or `"*"` for any
+      NFT defined in `contract-id`.
+    - `identifier`: `T`: The identifier of the token to grant access to.
```
radubahmata: Subjective suggestion: an explanation on how `*` for `token-name` in a contract with 2 or more defined NFTs works along with `identifier` would be valuable. If it's included in any linked discussion please ignore this comment.

sips/sip-033/sip-033-clarity4.md
obycode: The `int` is for the `err` case, the `ok` case is type `A`, which is dependent on the body, `uint` in the case of this example.

radubahmata: Correct. I mixed up the ok and err cases 👍

sips/sip-clarity4/sip-clarity4.md
obycode: Added in 573917b50ed47cdb49ba2887c765c97918b16beb.

sips/sip-clarity4/sip-clarity4.md
obycode: Fixed in 573917b50ed47cdb49ba2887c765c97918b16beb.

sips/sip-033/sip-033-clarity4.md
obycode: Added in 573917b50ed47cdb49ba2887c765c97918b16beb.

sips/sip-clarity4/sip-clarity4.md
obycode: Yeah, that makes sense. Changed in 573917b50ed47cdb49ba2887c765c97918b16beb.

sips/sip-clarity4/sip-clarity4.md
obycode: Updated in 573917b50ed47cdb49ba2887c765c97918b16beb.

sips/sip-clarity4/sip-clarity4.md
obycode: Good call. Updated in 573917b50ed47cdb49ba2887c765c97918b16beb.

sips/sip-clarity4/sip-clarity4.md
obycode: Added in 573917b50ed47cdb49ba2887c765c97918b16beb

sips/sip-clarity4/sip-clarity4.md
obycode: Updated in 573917b50ed47cdb49ba2887c765c97918b16beb

sips/sip-clarity4/sip-clarity4.md
radubahmata: Thanks! One thing I think is still worth clarifying is what should be passed as the `identifier` when using `"*"`. Since `identifier` is of type `T`:

For example, if a contract defines `nft1` and `nft2`, and we want to issue an allowance for all NFTs defined by the contract, we’d write:

```clarity
(with-nft (contract-of nft-trait) * <something>)
```

What exactly should `<something>` be in this case? Is it still a `uint` as in the examples, meaning the same id would issue an allowance for both `nft1` u123 and `nft2` u123? And if so, shouldn’t `T` just be `uint`?

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,384 @@
...
+principal against the granted allowances. If any allowance is exceeded,
+`restrict-assets?` returns an error; otherwise, it returns `ok` with the result
+of the last body expression.
+
+The most important use case of these new builtins is to allow a contract to
+protect its own assets. To ensure that the contract's assets are safe by
+default, the new `as-contract?` expression is similar to the existing
+`as-contract` expression, but will implicitly behave as though there is a
+`restrict-assets?` expression around its body, specifying the contract principal
+as the asset owner. The old `as-contract` is no longer available in Clarity 4.
```
MarvinJanssen, line 126: Removing the old `as-contract` introduces a very annoying pattern. Right now, the (only) way to get the current contract's principal is the following structure:

```clarity
(as-contract tx-sender)
```

If this SIP is implemented as it is currently written, to achieve the same:

```clarity
(unwrap-panic (as-contract? () tx-sender))
```

My suggestion: finally introduce a keyword for the current contract principal, alongside `tx-sender`, `contract-caller`, `tx-sponsor?`.

```clarity
current-contract
```

sips/sip-clarity4/sip-clarity4.md
wileyj: thanks! i'll add that idea to the retro doc as well to standardize on it

sips/sip-033/sip-033-clarity4.md
obycode: I like it!

obycode: Added 2f5479ae7f1c83ed40672ba4b69c5b26fa7bd8de

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,404 @@
...
+
+With this new `as-contract?`, the expression needed to get the principal of the
+current contract, a common pattern, becomes ugly:
+
+```clarity
+(unwrap-panic (as-contract? () tx-sender))
+```
+
+To avoid that, a new builtin keyword. `current-contract` is added which will
+return the principal for the current contract.
```
MarvinJanssen, line 139: I would personally put it in its own section with a rationale.

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,404 @@
...
+  (if (> block-time 1755820800)
+    (print "after 2025-07-22")
+    (print "before 2025-07-22"))
+  ```
+
+# Related Work
+
+Not applicable.
+
+# Backwards Compatibility
```
MarvinJanssen, line 503: I'm not sure if it is worth including, but `current-contract` will become a keyword in Clarity4. Any contract code predating it needs to be slightly refactored if it is to be redeployed as Clarity4. Luckily, it does appear that `current-contract` is [not used as a variable or constant name often](https://github.com/search?q=repo%3Aboomcrypto%2Fclarity-deployed-contracts%20current-contract&type=code).

(Disregarding @LNow's black magic that allows you to override keywords in certain situations.)

sips/sip-clarity4/sip-clarity4.md
obycode: Oh, that's an interesting case... if the two NFTs defined in the contract have different identifier types, what should we do? I think that should just cause a type error, basically making "*" not usable for that contract.

radubahmata: Agreed, I think we definitely shouldn't allow using "*" in that case. Unsure yet if that could lead to more complex issues. If not, probably a type error is enough.

MarvinJanssen: Why not permit multiple `with-nft` entries in such cases?

```clarity
(with-nft (contract-of nft-trait) * u14)
(with-nft (contract-of nft-trait) * "string-identifier")
```

It will be a rare occurrence but at least it will not prevent such structures.

MarvinJanssen:
I gave this yet another read after going over some LISA code. What is unclear from the SIP is whether an "allow mode"  as-contract call is still possible at all.

I agree with the rationale that we want to be safe by default, but there are still use-cases to forwarding the contract `tx-sender` to allow broad access. In fact, this mechanism is the cornerstone of ExecutorDAO (a popular DAO template these days) and protocols that build on it. I can give two examples of what we did in LISA:

https://github.com/lisalab-io/liquid-stacking/blob/31e4efaad057999e662272410c2c34e74b3ea368/contracts/extensions/treasury.clar#L30-L35

The LISA treasury has a `proxy-call` construct that can only be invoked in privileged context. These proxies can then arbitrarily implement new features that can include moving tokens. You can see the deployed proxies here:

https://github.com/lisalab-io/liquid-stacking/tree/31e4efaad057999e662272410c2c34e74b3ea368/contracts/proxies

Such a mechanism is not possible without allow mode as-contract calls.

Another example is the LISA vault contract:

https://github.com/lisalab-io/liquid-stacking/blob/main/contracts/extensions/lqstx-vault.clar#L16-L21

The `fund-strategy` and `refund-strategy` functions forward the contract `tx-sender` so that the strategies can utilise any and all tokens in the vault.

You can see the mechanism being utilised in the public pool strategy:

https://github.com/lisalab-io/liquid-stacking/blob/31e4efaad057999e662272410c2c34e74b3ea368/contracts/strategies/public-pools/public-pools-strategy.clar#L28-L54

I think the design is quite elegant, it would be unfortunate if it is made impossible in Clarity4.

sips/sip-clarity4/sip-clarity4.md
obycode: I think that's unnecessary, since the "*" is just a convenience addition -- they can still specify the names of both tokens. It seems cleaner to stick with a type error.

sips/sip-033/sip-033-clarity4.md
obycode: How does this sound?

> All new keywords introduced in Clarity 4 can no longer be used as identifiers in
> a Clarity 4 smart contract. This means that some existing contracts would need
> to be modified if published as a Clarity 4 contract.

sips/sip-clarity4/sip-clarity4.md
obycode: How do the changes in 4a0c614ff019892ec592e4661f2b86d68b658ff5 look?

sips/sip-033/sip-033-clarity4.md
obycode: I added this in 4a0c614ff019892ec592e4661f2b86d68b658ff5. Let me know what you think.

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+`string-ascii` representations.
+
+- **Input**: `int` | `uint` | `bool` | `principal` | `(buff 524284)` |
+  `(string-utf8 262144)`
+- **Output**: `(response (string-ascii 1048571) uint)`
+- **Signature**: `(to-ascii? value)`
+- **Description**: Returns the `string-ascii` representation of the input value
+  in an `ok` response on success. The only error condition is if the input type
+  is `string-utf8` and the value contains non-ASCII characters, in which case,
+  `(err u1)` is returned.
```
jcnelson: Do you want a different error for "value is too big to encode?"

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+The old `as-contract` builtin function is no longer available in Clarity 4,
+since it has been replaced with the safer `as-contract?`.
+
+All new keywords introduced in Clarity 4 can no longer be used as identifiers in
+a Clarity 4 smart contract. This means that some existing contracts would need
+to be modified if published as Clarity 4 contracts.
+
+# Activation
+
+TBD
```
jcnelson: You can probably copy/paste and adapt the past coin vote hard forks we've done in the past.  But also please add a caveat that none of the SIP-031 tokens can vote (per SIP-031)

sips/sip-clarity4/sip-clarity4.md
obycode: That's not possible, since it is enforced by the type system.

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,466 @@
+# Preamble
+
+**SIP Number:** XYZ
```
314159265359879: Lets make it SIP033, that is not used yet and it makes it easier to reference this.

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+
+**Sign-off:**
+
+**Discussions-To:**
+https://forum.stacks.org/t/clarity-4-proposal-new-builtins-for-vital-ecosystem-projects/18266
+
+# Abstract
+
+This SIP details Version 4 of the Clarity smart contract language, which
+introduces new operations to make it easier to build secure smart contracts on
```
314159265359879: the following replacement would increase precision.
... to make it easier to build secure...
replace with: ...to provide mechanisms for building more secure...

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+
+This SIP is made available under the terms of the CC0 (Creative Commons Zero)
+license, available at https://opensource.org/licenses/CC0-1.0. This SIP’s
+copyright is held by the Stacks Open Internet Foundation.
+
+# Introduction
+
+The Clarity smart contract language powers decentralized applications across a
+wide range of domains. Over years of real-world use, developers have encountered
+several pain points when interacting with other, sometimes untrusted, contracts.
```
314159265359879: instead of "pain points" use "security and interoperability challenges" to be consistent with the technical tone of this SIP.

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+   facilitating the generation of string-based messages for interacting with
+   cross-chain protocols.
+4. **A new Clarity keyword to get the timestamp of the _current_ block.** This
+   keyword will allow developers to easily access the timestamp of the block
+   currently being processed, enabling time-based logic and features in their
+   smart contracts. This is especially important for DeFi applications.
+
+# Specification
+
+## Fetching the hash of a contract body: `contract-hash?`
```
314159265359879: Fetching the Hash of a Contract Body: `contract-hash?`

Use consistent heading style across all subsections (capitalizing each main word).

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+  - `(ok 0x<hash>)`, where `<hash>` is the SHA-512/256 hash of the code body, on
+    success
+  - `(err u1)` if the principal is not a contract principal
+  - `(err u2)` if the specified contract does not exist
+- **Example**:
+  ```clarity
+  (contract-hash? 'SP2QEZ06AGJ3RKJPBV14SY1V5BBFNAW33D96YPGZF.BNS-V2) ;; Returns (ok 0x9f8104ff869aba1205cd5e15f6404dd05675f4c3fe0817c623c425588d981c2f)
+  ```
+
+## Limiting asset access: `restrict-assets?`
```
314159265359879: Limiting Asset Access: `restrict-assets?`

Use consistent heading style across all subsections (capitalizing each main word).

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+      )
+    ) ;; Returns (err -1)
+    (define-public (bar)
+      (as-contract? ((with-stx u1000000))
+        (try! (stx-transfer? u1000000 tx-sender recipient))
+      )
+    ) ;; Returns (ok true)
+    ```
+
+## Built-in keyword: `current-contract`
```
314159265359879: Built-in Keyword: `current-contract`

Use consistent heading style across all subsections (capitalizing each main word).

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+  `(err u1)` is returned.
+- **Example**:
+  ```clarity
+  (to-ascii? true) ;; Returns (ok "true")
+  (to-ascii? 42) ;; Returns (ok "42")
+  (to-ascii? 'SP2QEZ06AGJ3RKJPBV14SY1V5BBFNAW33D96YPGZF) ;; Returns (ok "SP2QEZ06AGJ3RKJPBV14SY1V5BBFNAW33D96YPGZF")
+  (to-ascii? 0x12345678) ;; Returns (ok "0x12345678")
+  ```
+
+## Timestamp for current block: `block-time`
```
314159265359879: Timestamp for Current Block...

consistent heading style

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+- `restrict-assets?`
+
+  - **Input**:
+    - `asset-owner`: `principal`: The principal whose assets are being
+      protected.
+    - `((with-stx|with-ft|with-nft|with-stacking)*)`: The set of allowances to
+      grant during the evaluation of the body expressions.
+    - `AnyType* A`: The Clarity expressions to be executed within the context,
+      with the final expression returning type `A`, where `A` is not a
+      `response`
```
314159265359879, line 170: ..., where `A` is not a `response`
Add rationale for why nested `response` is disallowed; the "error-prone" note is brief, consider expanding it?

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+
+- `with-all-assets-unsafe`
+
+  - **Input**: None
+  - **Output**: Not applicable
+  - **Signature**: `(with-all-assets-unsafe)`
+  - **Description**: Grants unrestricted access to all assets of the contract to
+    the enclosing `as-contract?` expression. Note that this is not allowed in
+    `restrict-assets?` and will trigger an analysis error, since usage there
+    does not make sense (i.e. just remove the `restrict-assets?` instead). **_⚠️
```
314159265359879: consider using "**Security Warning**" (bold) instead of ⚠️emoji. Emojis are not standard in SIP texts.

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+    ) ;; Returns (ok true)
+    ```
+
+## Built-in keyword: `current-contract`
+
+In many smart contracts, there is a common pattern for retrieving the current
+contract's principal:
+
+```clarity
+(as-contract tx-sender)
```
314159265359879, line 382: clarify this is about Clarity 3, not Clarity 4?

or it contradicts the next sentence...removal of `as-contract`,...

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+
+## Conversion to `string-ascii`: `to-ascii?`
+
+Originally proposed here: https://github.com/clarity-lang/reference/issues/82
+
+`to-ascii?` is a new Clarity function that converts simple values into their
+`string-ascii` representations.
+
+- **Input**: `int` | `uint` | `bool` | `principal` | `(buff 524284)` |
+  `(string-utf8 262144)`
```
314159265359879, line 411: `(buff 524284)` |  `(string-utf8 262144)`
Consider adding reasoning why these specific limits are chosen. This can help implementers and reviewers.

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+  ```
+
+## Timestamp for current block: `block-time`
+
+`block-time` is a new Clarity keyword that returns the timestamp of the current
+block in seconds since the Unix epoch. This is the same timestamp that is in the
+block header, which is verified by the signers to be:
+
+- Greater than the timestamp of its parent block
+- At most 15 seconds into the future
```
314159265359879, line 441: Reference consensus rules source (Stacks blockchain protocol) for verification instead of leaving as a bare fact.

sips/sip-clarity4/sip-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+- **Example**:
+  ```clarity
+  (if (> block-time 1755820800)
+    (print "after 2025-07-22")
+    (print "before 2025-07-22"))
+  ```
+
+# Related Work
+
+Not applicable.
```
314159265359879: Better than leaving blank I think we can mention related SIPs:

Clarity version 1
https://github.com/stacksgov/sips/blob/main/sips/sip-002/sip-002-smart-contract-language.md
https://github.com/stacksgov/sips/blob/main/sips/sip-006/sip-006-runtime-cost-assessment.md
https://github.com/stacksgov/sips/blob/main/sips/sip-008/sip-008-analysis-cost-assessment.md

Clarity version 2 introduced with SIP015
https://github.com/stacksgov/sips/blob/main/sips/sip-015/sip-015-network-upgrade.md
https://github.com/stacksgov/sips/blob/main/sips/sip-020/sip-020-bitwise-ops.md

Clarity v3 introduced with SIP021
https://github.com/stacksgov/sips/blob/main/sips/sip-021/sip-021-nakamoto.md

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,466 @@
...
+change. A contract that uses one of these new operators would be invalid before
+this SIP is activated, and valid after it is activated.
+
+The old `as-contract` builtin function is no longer available in Clarity 4,
+since it has been replaced with the safer `as-contract?`.
+
+All new keywords introduced in Clarity 4 can no longer be used as identifiers in
+a Clarity 4 smart contract. This means that some existing contracts would need
+to be modified if published as Clarity 4 contracts.
+
```
314159265359879, line 517: Expand with migration considerations. What tools or warnings will guide developers upgrading contracts or using older (clarity 1, 2, 3) contracts?

It may be obvious for people who have worked with Clarity for longer but not for newer users. How to use `at block` or Clarity 1, 2 and 3 contracts after Clarity 4 is live.

sips/sip-033/sip-033-clarity4.md
obycode: I'm having a little trouble coming up with good wording for this. I'll come back to it later, or please let me know if you have a suggestion.

sips/sip-clarity4/sip-clarity4.md
obycode: How do we plan to enforce this in the future once those tokens start moving?

sips/sip-033/sip-033-clarity4.md
obycode: I added one sentence here, but I'm not sure what else is needed.

sips/sip-clarity4/sip-clarity4.md
wileyj: hmm, i see the intent but i think this could lead to a case where each subsequent clarity change should reference all other sips that precede it. 

for clarity-5 and on, i think it's fine to mention previous sips, but no need to enumerate them all.

sips/sip-clarity4/sip-clarity4.md
wileyj: 🤔 good question. that might require some changes as to how votes are tallied. i have some ideas, but they might be controversial and out of scope for this PR.

sips/sip-clarity4/sip-clarity4.md
obycode: so leave this or revert it?

wileyj: since it was already added, i say keep it. just noting for future sips so we don't get to a place where we're always adding several previous sips as related work.

sips/sip-clarity4/sip-clarity4.md
pjklein: Do the vote contracts need to be implementing one of two traits to create an on-chain demarcation:

- one trait for pre-movement SIP votes - to mark that the vote is happening before movement has become possible
- another trait for post-movement SIP votes that implements the possible changes @wileyj surmises above

There is no mechanism for mutexing a pair of traits, is there?

cc @obycode @jcnelson

pjklein: @wileyj , since "out of scope for this PR" where should the discussion of enforcement be had? Would it be additional, non-concensus-related implementation in support of SIP-031?

obycode:
For the `restrict-assets?` and `as-contract?` expressions, I was attempting to design it such that we could avoid nested responses, but when actually finishing the implementation and writing examples, I realized that this is not a reasonable restriction, because it is too limiting on the body operations. That means that the return type from one of these expressions could be `(response (response x y) int)`. Hopefully this is the last change needed to the SIP. Let me know if anyone has any alternative ideas, but I think it's either this, or make the allowance violation error not catchable.

```clarity
(as-contract? ((with-stx u1000))
  (try! (stx-transfer? u200 tx-sender recipient))
  (contract-call? .other do-something x y)
)
```

`as-contract?` returns a `(response T int)`. Let's say `.other::do-something` returns `(response uint uint)`. So that would require that `as-contract?` returns a `(response (response uint uint) int)`. The only way to avoid that would be to force the body expressions to only throw `(err int)`, or to panic (e.g. with an `unwrap-panic`). I think it would be okay to have the nested response , as long as Clarity enforces that both are handled (which I don't think it does currently). Again, the alternative is that the allowance violation is just a runtime error which is not catchable.

sips/sip-clarity4/sip-clarity4.md
wileyj: > @wileyj , since "out of scope for this PR" where should the discussion of enforcement be had? Would it be additional, non-concensus-related implementation in support of SIP-031?

well, i don't think individual sips are a good place to enforce that rule - but for now, it's necessary and needs to be accounted for in the CABS and steering committee before coming to a vote. 

i think enforcement should require updates to the text in sip-000 (already being discussed, PR with proposed updates and methods on how to update sip-000 is being worked on). i suspect any method to enforce that sip-031 tokens cannot vote will require a bit of scrutiny.

sips/sip-033/sip-033-clarity4.md
wileyj: I beleive this is answered:
```
Note that the limitation on the maximum sizes of
  `buff` and `string-utf8`
```

sips/sip-033/sip-033-clarity4.md
wileyj: link added

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,581 @@
...
+Type: Consensus
+
+Layer: Consensus (hard fork)
+
+Created: 2025-06-20
+
+License: CC0-1.0
+
+Sign-off:
```
wileyj, line 27:
```suggestion
Sign-off:
- Jesse Wiley <jesse@stacks.org>, Chairperson, Technical CAB
```

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,563 @@
...
+
+Title: Clarity Smart Contract Language, version 4
+
+Author(s):
+
+- Adriano Di Luzio <adriano@bitcoinl2labs.com>
+- Brice Dobry <brice@hiro.so>
+- Marvin Janssen <marvin@ryder.id>
+- Jude Nelson <jude@stacks.org>
+
```
Hero-Gamer, line 14: I'd suggest adding @Rapha-btc as one of authors for his contribution for advocating for `secp256r1` inclusion on this SIP. 

I would not have been aware the impact of this primitive  if it wasn't for his thread.

Discussion-to: 
1) [Adding WebAuthn (P-256) Support to Clarity for Native Passkey Integration](https://forum.stacks.org/t/adding-webauthn-p-256-support-to-clarity-for-native-passkey-integration/17886)
2) [Clarity 4 Proposal: New builtins for vital ecosystem projects](https://forum.stacks.org/t/clarity-4-proposal-new-builtins-for-vital-ecosystem-projects/18266)

Rapha also came on [weekly SIP call](https://forum.stacks.org/t/adding-webauthn-p-256-support-to-clarity-for-native-passkey-integration/17886/13?u=herogamer) with Brice to share context and details relating to this feature.

brice-stacks: Added

sips/sip-033/sip-033-clarity4.md
obycode: It seems like this change should come from the technical steering committee chair directly.

wileyj: i can send a PR to your branch if you'd prefer. https://github.com/stacksgov/sips/blob/main/considerations/technical.md

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,564 @@
...
+  ```clarity
+  (if (> stacks-block-time 1755820800)
+    (print "after 2025-07-22")
+    (print "before 2025-07-22"))
+  ```
+
+## Secp256r1 Functions
+
+Clarity 4 introduces a new function to verify signatures for the secp256r1
+elliptic curve, which is widely used for cryptographic operations.
```
jcnelson: Might want to add an example use-case, such a WebAuthn.

whoabuddy: tagging @Rapha-btc if he has something to add here

sips/sip-033/sip-033-clarity4.md
obycode: Oh, should this say "Technical CAB Chair"? I thought this was for sign-off from the steering committee. Yes, the technical CAB sign-off should come from you. We still need to define a standard process for how that happens. For now, I'm happy to just commit a suggestion, but can you update the text?

sips/sip-033/sip-033-clarity4.md
obycode: Is this clear now? Can we resolve this?

sips/sip-033/sip-033-clarity4.md
obycode: Any other changes recommended to clarify this further or can we resolve this?

sips/sip-033/sip-033-clarity4.md
Rapha-btc: Passkey-Native Wallet, smart wallet contract principal governed by passkey. Wallet created in 3 seconds, Face ID done. No seed phrase "onboard your mom" style.

Rapha-btc:
I'm looking for examples of `contract-hash?` usage to verify DEX templates before they can self-register for token transfers. This would enforce royalties when users sell tokens to DEX contracts by creating an auto-allowlist of compliant exchanges that implement a required royalty-collecting function. My first implementation is at https://github.com/Rapha-btc/pizza-square/blob/main/contracts/pizza-faktory.clar copy @obycode

sips/sip-033/sip-033-clarity4.md
wileyj: updated to use the current format from recent SIPs until a specific format is codified

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,567 @@
...
+
+Author(s):
+
+- Brice Dobry <brice@stackslabs.com>
+- Marvin Janssen <marvin@ryder.id>
+- Jude Nelson <jude@stacks.org>
+- Rapha.btc <ra@fak.fun>
+- Adriano Di Luzio <adriano@bitcoinl2labs.com>
+
+Status: Draft
```
jcnelson: Can you advance this to Activation-in-Progress @obycode?

sips/sip-033/sip-033-clarity4.md
```diff
@@ -0,0 +1,567 @@
...
+Type: Consensus
+
+Layer: Consensus (hard fork)
+
+Created: 2025-06-20
+
+License: CC0-1.0
+
+Sign-off:
+- Jesse Wiley <jesse@stacks.org>, Chairperson, Technical CAB
```
jcnelson: Can you add the governance sign-off as well?

sips/sip-033/sip-033-clarity4.md
obycode: [c2e7310](https://github.com/stacksgov/sips/pull/218/commits/c2e7310ac34f30a8fa33b6805e705b2368a58cad)

sips/sip-033/sip-033-clarity4.md
obycode: This one required Technical CAB only.

sips/sip-033/sip-033-clarity4.md
314159265359879: @obycode I wasn't sure why this is done either, so it depends on what this tries to achieve. 

If the goal is to _avoid reversion ambiguities_, something like this:

Because `restrict-assets?` can revert all body expressions on error, nested `response` values could introduce ambiguity about which layer’s error should trigger a rollback. For clarity and safety, only the outermost response layer may represent the operation’s success or failure.

314159265359879 (approved):
This is looking great to me. I am hereby signing-off on SIP033.

I see one open comment, I would like to see addressed but I consider that non-blocking for this editor sign-off.

mphizi:
Hi @obycode, did the proposal for type-aliases ([https://github.com/clarity-lang/reference/issues/6](https://github.com/clarity-lang/reference/issues/6)) especially for tuples and list definitions get into this SIP or SIP-034?

I think this would really great ergonomic improvement. Thanks for all the great work taking place in the ecosystem!

wileyj (approved):

sips/sip-033/sip-033-clarity4.md
obycode: Let's go ahead and resolve this. The wording in the description was updated to better explain this.

PR #250, opened 2026-01-10
SIP-038: Standard Trait Definition for Commitment-Based Private Metadata (Encrypted NFTs)

proofofgame:
Adds a draft SIP proposing an eNFT trait for commitment-based private layers (non-consensus). Discussion: https://forum.stacks.org/t/encrypted-nfts-on-stacks-a-standard-trait-for-commitment-based-private-layers-enft/18560


sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
+# SIP-XXX: Encrypted NFTs - Standard Trait for Commitment-Based Private Layers (eNFT)
+
+## Preamble
+
+- **SIP**: SIP-XXX
+- **Title**: Encrypted NFTs - Standard Trait for Commitment-Based Private Layers (eNFT)
```
314159265359879: I suggest this Title:
Standard Trait Definition for Commitment-Based Private Layers (Encrypted NFTs)

To be consistent with titles for SIP009 and SIP010. 

"Layers" in crypto may be misunderstood in this context. Would "...Commitment-Based Private Metadata" also be accurate? It would prevent that ambiguity for users just reading the title. 

I think you do a good job of explaining what is meant by "Private Layers" in this context in the SIP. Let's see how other editors think about this.

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+- **Authors**: Xenitron nortinex@gmail.com
+- **Status**: Draft
+- **Consideration**: Technical
+- **Type**: Standard
+- **Layer**: Traits
```
314159265359879: I am not sure we need this "Layer" line. I see it in recent SIPs (31, 33, 34), but not in the older ones, and it looks like it is repeating "type". in all of these, including this one. 
Example from SIP033
````
Type: Consensus
Layer: Consensus (hard fork)
````

This is not a question for the author more for the other editors/CAB's. Perhaps I am misunderstanding the purpose this "Layer" categorization. @wileyj what do you think?

When I see "Layer", I am thinking more along the lines of "Stacks (Bitcoin L2)", or "Stacks subnet (Bitcoin L3)", it would make it a future-oriented categorization field.

File: sip-enft-private-layer.md [hunk 3]
```diff
@@ -0,0 +1,197 @@
...
+- **Created**: 2026-01-10
+- **License**: CC0-1.0
+- **Sign-off**: 
+- **Discussions-To**: https://forum.stacks.org/t/encrypted-nfts-on-stacks-a-standard-trait-for-commitment-based-private-layers-enft/18560
+
+# Abstract
+
+This SIP proposes a simple, composable trait for “Encrypted NFTs” (eNFTs): NFTs that commit on-chain to a private off-chain payload (a “private layer”) while keeping the payload itself off-chain. The standard defines (1) a public commitment interface, (2) an optional owner-gated envelope descriptor for wallet UX, and (3) indexer-friendly event conventions. This enables interoperable private-layer NFT patterns across contracts, wallets, marketplaces, and indexers without requiring consensus changes.
```
314159265359879: I would prefer if this first sentence of the abstract was split in two for readability. 

For example:
This SIP proposes a simple, composable trait for “Encrypted NFTs” (eNFTs). An eNFT commits on-chain to a private off-chain payload (“private layer”) while the payload itself remains off-chain.

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+
+# Copyright
+
+This SIP is made available under the terms of the CC0-1.0 license.
+This SIP’s copyright is waived to the extent possible under law.
+
+# Introduction
+
+Today, “private layer / encrypted metadata” NFTs on Stacks are implemented as bespoke, project-specific patterns. That fragmentation makes it hard for wallets, marketplaces, and indexers to support them generically.
```
314159265359879: “Today,” is acceptable but unnecessary; remove to reduce conversational tone.

“as bespoke, project-specific patterns” is clear but slightly redundant; one modifier can be removed.
i.e. “...as project-specific patterns”

“support them generically” → “support them in a generic way” or “support them uniformly”.

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+
+This SIP standardizes a minimal interface for eNFT private layers:
+- A **public commitment** (hash) to the private payload, enabling anyone to verify integrity after reveal.
+- An optional **owner-gated envelope descriptor** to guide wallet UX (e.g., show “Reveal” only to the owner).
```
314159265359879: UX --> User Experience 

Easier to read for non-technical people.

314159265359879: want to also clarify that verification on "Reveal" is client-side to avoid ambiguity?

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+- Full on-chain confidentiality (not achievable in a public read-only model).
+- Mandating a specific storage network, encryption scheme, or reveal flow.
+- Defining a full game/puzzle “answer verification” protocol (out of scope).
+
+# Specification
+
+## 1. Terminology
+
+- **Payload**: The private off-chain data (JSON, image, encrypted blob, etc.).
+- **Commitment**: A hash over the payload (or ciphertext) recorded on-chain.
```
314159265359879: Consider explicitly stating that the commitment is immutable per version unless updated by contract logic?

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+- **Envelope**: A small off-chain descriptor used for UX/access, e.g., a retrieval URI, encrypted key, or signed URL.
+- **Reveal**: Off-chain delivery of payload/envelope to the user; verification is done by comparing hash(payload) to on-chain commitment.
+
+## 2. Commitment Semantics
+
+Implementations MUST expose a 32-byte commitment for a given token-id.  
+The commitment MUST be computed using the declared algorithm identifier.
+
+This SIP does not mandate a single algorithm, but RECOMMENDS:
```
314159265359879: ..., but recommends:

Avoid all caps except for MUST/SHOULD terminology.

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+  )
+)
+```
+
+## 3.2 Error Codes (Recommended)
+Implementations MAY choose their own error codes. This SIP RECOMMENDS:
+
+`err u404` => token-id not found / no private layer set
+
+`err u100` => not authorized (e.g., tx-sender not owner) for get-owner-envelope
```
314159265359879: Consider formatting as a bullet list instead of inline code blocks for readability.

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+  token-id: token-id,
+  envelope-hash: envelope-hash,
+  algo: algo,
+  envelope-version: envelope-version
+})
+```
+
+Notes:
+
+- Events are intended for indexers/wallets discovery and UI updates.
```
314159265359879: UI --> user interface

Improve readability.

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+This SIP’s copyright is waived to the extent possible under law.
+
+# Introduction
+
+Today, “private layer / encrypted metadata” NFTs on Stacks are implemented as bespoke, project-specific patterns. That fragmentation makes it hard for wallets, marketplaces, and indexers to support them generically.
+
+This SIP standardizes a minimal interface for eNFT private layers:
+- A **public commitment** (hash) to the private payload, enabling anyone to verify integrity after reveal.
+- An optional **owner-gated envelope descriptor** to guide wallet UX (e.g., show “Reveal” only to the owner).
+- A recommended **event format** so indexers can discover updates efficiently.
```
314159265359879: consider alternative:
...discover updates efficiently --> detect updates efficiently.

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Goals
+
+- Define a reusable trait for commitment-based private layers for NFTs.
+- Make private-layer patterns discoverable and composable for wallets/indexers.
+- Keep the standard minimal and non-consensus.
```
314159265359879: I think I might prefer being explicit about scope without using consensus language, consider:
- Keep the standard minimal, additive, and backward compatible.

This also still says "non-consensus" and "no breaking changes" but it does so precisely rather than implicitly.

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+Notes:
+
+- Events are intended for indexers/wallets discovery and UI updates.
+
+- `envelope-uri` SHOULD NOT be treated as confidential by default.
+
+## Backwards Compatibility
+This SIP is additive and does not break existing SIP-009 NFT contracts. Projects can implement this trait alongside existing token URI metadata patterns. Wallets/indexers can support eNFTs incrementally.
```
314159265359879: Consider explicitly stating “no breaking changes” to align with SIP language used elsewhere.

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Security Considerations
```
314159265359879: Very well done, this section!

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+**Read-only visibility and “owner gating”**
+
+`get-owner-envelope` ownership checks are primarily for **wallet UX consistency**, not cryptographic secrecy. Read-only calls can be simulated by node operators; therefore, **confidentiality MUST NOT rely solely on on-chain gating**.
```
314159265359879: Replace “node operators” with “any party running a node” for clarity to non-technical readers.

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+
+If the envelope or payload must remain private, it SHOULD be protected off-chain via:
+
+- encryption (e.g., encrypt payload/key per owner), and/or
+
+- access control (signed messages, bearer tokens, short TTL URLs).
+
+**Integrity and server non-equivocation**
```
314159265359879: “Integrity and server non-equivocation”, consider explicitly stating “assuming collision-resistant hash functions”.

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+- storage mapping token-id => envelope tuple
+
+- setters restricted to contract-defined roles (e.g., token owner or game operator)
+
+- `print` events as specified above
+
+## Related Work
+
+Cryptographic commitments (hash commitments) and commit–reveal mechanics are widely used across blockchains and in on-chain games to prevent equivocation. Separately, some NFT ecosystems support “encrypted metadata” or off-chain private content patterns, but these are typically application-specific and lack a shared interface for wallets/indexers.
+To the best of our knowledge, Stacks currently has no minimal, interoperable trait + event convention that standardizes a commitment-anchored private layer for SIP-009 NFTs. This SIP fills that gap by defining a composable trait and predictable indexer events, while remaining non-consensus and implementation-agnostic.
```
314159265359879: Slight redundancy between the last two sentences; they can be merged or tightened.

For example:
To the best of our knowledge, Stacks lacks a minimal, interoperable trait and event convention for commitment-anchored private layers on SIP-009 NFTs; this SIP defines such a standard in a composable, non-consensus, and implementation-agnostic manner.

Instead of 
````
To the best of our knowledge, Stacks currently has no minimal, interoperable trait + event convention that standardizes a commitment-anchored private layer for SIP-009 NFTs. This SIP fills that gap by defining a composable trait and predictable indexer events, while remaining non-consensus and implementation-agnostic.
````

File: sip-enft-private-layer.md [hunk 17]
```diff
@@ -0,0 +1,197 @@
...
+
+- NFT contracts implement the trait
+
+- wallets/indexers add UI for commitments and optional reveal flows
+
+- marketplaces may display “private layer present” and verified reveal status
+
+## References
+
+[1] SIP-009 (NFT trait standard on Stacks)
```
314159265359879: add link to sip

314159265359879: https://github.com/stacksgov/sips/blob/main/sips/sip-009/sip-009-nft-standard.md

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+
+[2] Forum discussion thread linked in Preamble
```
314159265359879: add link to forum discussion
https://forum.stacks.org/t/encrypted-nfts-on-stacks-a-standard-trait-for-commitment-based-private-layers-enft/18560

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+The commitment MUST be computed using the declared algorithm identifier.
+
+This SIP does not mandate a single algorithm, but RECOMMENDS:
+- `algo = u0` => SHA-256 over raw payload bytes
+
+## 3. Trait Interface (Clarity)
+
+The trait defines two read-only methods: one public (commitment) and one optional/UX-oriented (owner envelope).
+
+### 3.1 Clarity trait definition
```
314159265359879: ...Trait Definition (capitalization)

sip-enft-private-layer.md
```diff
@@ -0,0 +1,197 @@
...
+- setters restricted to contract-defined roles (e.g., token owner or game operator)
+
+- `print` events as specified above
+
+## Related Work
+
+Cryptographic commitments (hash commitments) and commit–reveal mechanics are widely used across blockchains and in on-chain games to prevent equivocation. Separately, some NFT ecosystems support “encrypted metadata” or off-chain private content patterns, but these are typically application-specific and lack a shared interface for wallets/indexers.
+To the best of our knowledge, Stacks currently has no minimal, interoperable trait + event convention that standardizes a commitment-anchored private layer for SIP-009 NFTs. This SIP fills that gap by defining a composable trait and predictable indexer events, while remaining non-consensus and implementation-agnostic.
+
+## Activation
```
314159265359879: With SIP009 and SIP010 we had criteria for minimal adoption by a certain bitcoin block. So that we have some signal for adoption before activating the SIP. I suggest something similar here. 

This SIP is activated when three contracts are deployed on mainnet that adopt the trait specified in this standard before Bitcoin block 1,000,000. 

A trait that follows this specification is available on mainnet as...


From SIP009
````
This SIP is activated if 5 contracts are deployed that use the same trait that follows this specification. This must happen before Bitcoin tip #700,000.

A trait that follows this specification is available on mainnet as SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait.nft-trait.
````

**Review by 314159265359879 — CHANGES_REQUESTED** [review]
@proofofgame Thank you for sending this PR, I think this would be an excellent addition. It is well written and already close to ready for the next step.

Overall I believe:
- The SIP meets the minimum bar for a standards-track proposal.
- Technical content is coherent, minimal, and well justified.
- With minor editorial tightening and one or two clarifying statements, it would be appropriate for formal consideration and community review under the SIP process.

**Minimum additions or clarifications recommended before editor sign-off:**

1. Explicit relationship to SIP-009
- Add one short paragraph explicitly stating that this trait is designed to be implemented alongside SIP-009-compliant NFTs, not as a replacement.
- This avoids ambiguity for reviewers and implementers.

2. Trait optionality clarification
- Explicitly state that get-owner-envelope is optional and that contracts may implement only the commitment interface.
- This is implied but should be stated normatively.

3. Algorithm registry guidance
- While flexibility is good, SIP reviewers may request a short table or registry suggestion for algo values (even if non-binding).
- This can remain informative, not normative.

4. Failure mode clarity
- Clarify expected behavior when no private layer exists for a token-id (e.g., return error vs. empty optional).
- You already recommend an error code; explicitly tie this to trait behavior.

**Some global editorial notes:**
- Capitalization is mostly consistent; keep section headers capitalized, but avoid unnecessary capitalization in prose.
- Reduce parenthetical overload where possible; move clarifications into short sentences.
- Prefer declarative, shorter sentences for readability.
- Replace passive constructions where clarity improves.

I did my best to mark specific ways on how to improve the SIP further, as much as possible with examples so hopefully you don't have to guess what I mean. Feel free to make adjustments if there is anything I misunderstood. 

Overall quality is high. The proposal is clear, technically sound, and already close to SIP-ready. Issues are mostly consistency, minor grammar, and opportunities to simplify language without losing rigor.

File: sip-enft-private-layer.md (hunk 19 above)
Comment (proofofgame) [reply sip-enft-private-layer.md]: Regarding optionality: We decided to split the standard into two separate traits (commitment and envelope). This allows strictly strictly-typed compliance in Clarity without forcing implementers to write stub functions for optional features.

File: sip-enft-private-layer.md (hunk 16 above)
Comment (proofofgame) [reply sip-enft-private-layer.md]: Thanks for the detailed review! I've addressed all comments:
- Added links to References section
- Changed title as suggested
- Split traits into required/optional
- Added Algorithm Registry table
- Clarified relationship to SIP-009

Ready for another look when you have time.

File: sip-enft-private-layer.md (hunk 1 above)
Comment (proofofgame) [reply sip-enft-private-layer.md]: Thanks! Agreed on the ambiguity. I updated the SIP title to use “Commitment-Based Private Metadata” for clarity, and added a brief note in Terminology that “private layer” is a legacy synonym.

File: sip-enft-private-layer.md (hunk 2 above)
Comment (proofofgame) [reply sip-enft-private-layer.md]: Good point. I removed the “Layer:” field for now to match older SIP formatting; happy to re-add if editors prefer keeping it.

File: sip-enft-private-layer.md (hunk 17 above)
Comment (proofofgame) [reply sip-enft-private-layer.md]: Done

File: sip-enft-private-layer.md (hunk 20 above)
Comment (proofofgame) [reply sip-enft-private-layer.md]: Thanks for the feedback on activation criteria.

I've thought about this carefully. SIP-009 required 5 contracts because it's foundational infrastructure - every NFT project needs it. This SIP is a specialized trait for a specific use case (private metadata), so adoption will naturally be more limited.

Proposed approach:
1. Reference trait deployed (we can do this)
2. At least one production/mainnet NFT contract uses the trait(s) exactly as specified in this SIP (we can upgrade/deploy a production contract for our eNFT games to match the final interface + events)
3. At least one ecosystem integration - indexer, wallet, or marketplace demonstrating they can consume the interface

Criteria (3) is key: it proves interoperability without requiring us to convince other projects to implement a Draft SIP. An indexer parsing our events is independent validation that the interface works.

Comment (314159265359879) [reply sip-enft-private-layer.md]: That sounds reasonable to me. I like the approach. I would still like to see a time component. "...before Bitcoin block x."

Comment (proofofgame) [reply sip-enft-private-layer.md]: Makes sense. We can add a time-bounded activation window expressed in Bitcoin blocks: before Bitcoin block height (H₀ + N), where H₀ is the Bitcoin height at the time this SIP is merged/ratified, and N is the window (e.g., 12,960 blocks ≈ ~3 months). This keeps the deadline objective and avoids hardcoding a moving “current height” value.

File: sip-enft-private-layer.md [hunk 21]
```diff
@@ -0,0 +1,250 @@
+# SIP-XXX: Standard Trait Definition for Commitment-Based Private Metadata (Encrypted NFTs)
+
+## Preamble
+
+- **SIP**: SIP-XXX
```
wileyj: Proposing SIP-038
```suggestion
- SIP Number: 038  
```

sip-enft-private-layer.md
```diff
@@ -0,0 +1,250 @@
...
+- **Title**: Standard Trait Definition for Commitment-Based Private Metadata (Encrypted NFTs)
+- **Authors**: Xenitron nortinex@gmail.com
+- **Status**: Draft
+- **Consideration**: Technical
+- **Type**: Standard
+- **Created**: 2026-01-10
+- **License**: CC0-1.0
+- **Sign-off**: 
+- **Discussions-To**: https://forum.stacks.org/t/encrypted-nfts-on-stacks-a-standard-trait-for-commitment-based-private-layers-enft/18560
```
wileyj: Please follow standard formatting
```suggestion
Title: Standard Trait Definition for Commitment-Based Private Metadata (Encrypted NFTs)  
Authors: Xenitron nortinex@gmail.com 
Status: Draft  
Consideration: Technical  
Type: Standard  
Created: 2026-01-10  
License: CC0-1.0  
Sign-off:   
Discussions-To: https://forum.stacks.org/t/encrypted-nfts-on-stacks-a-standard-trait-for-commitment-based-private-layers-enft/18560  
```

proofofgame: Done

sips/sip-038/sip-enft-private-layer.md
```diff
@@ -0,0 +1,261 @@
...
+SIP Number: 038  
+Title: Standard Trait Definition for Commitment-Based Private Metadata (Encrypted NFTs)   
+Authors: Xenitron nortinex@gmail.com, Bogachev ogres2009anton@gmail.com  
+Status: Draft  
+Consideration: Technical  
+Type: Standard  
+Created: 2026-01-10  
+License: CC0-1.0  
+Sign-off:  
+Discussions-To: https://forum.stacks.org/t/encrypted-nfts-on-stacks-a-standard-trait-for-commitment-based-private-layers-enft/18560
```
wileyj, line 15: I think we need to adjust this to `Layer: Trait` if the proposed is adding a new trait. 


```suggestion
Sign-off:  
Layer: Trait  
Discussions-To: https://forum.stacks.org/t/encrypted-nfts-on-stacks-a-standard-trait-for-commitment-based-private-layers-enft/18560  
```

sips/sip-038/sip-enft-private-layer.md
```diff
@@ -0,0 +1,261 @@
...
+Sign-off:  
+Discussions-To: https://forum.stacks.org/t/encrypted-nfts-on-stacks-a-standard-trait-for-commitment-based-private-layers-enft/18560  
+
+# Abstract
+
+This SIP proposes a minimal, composable trait interface for “Encrypted NFTs” (eNFTs): NFTs that commit on-chain to a private off-chain payload (“private metadata”) while keeping the payload itself off-chain. The standard defines:
+
+1) a public commitment interface,  
+2) an optional owner-gated envelope descriptor for wallet User Experience, and  
+3) indexer-friendly event conventions.
```
wileyj:
```suggestion
1) A public commitment interface
2) Optional owner-gated envelope descriptor for wallet user experience.
3) Indexer-friendly event conventions.
```

sips/sip-038/sip-enft-private-layer.md
```diff
@@ -0,0 +1,261 @@
...
+
+
+## Introduction
+
+“Encrypted metadata / private metadata” NFTs on Stacks are currently implemented as project-specific patterns. This fragmentation makes it difficult for wallets, marketplaces, and indexers to support such NFTs uniformly.
+
+This SIP standardizes a minimal interface for commitment-based private metadata on SIP-009 NFTs:
+
+- A **public commitment** (hash) to the payload (or ciphertext), enabling anyone to verify integrity after reveal.
+- An optional **owner-gated envelope descriptor** to guide wallet User Experience (e.g., show “Reveal” only to the owner).
```
wileyj:
```suggestion
- An optional **owner-gated envelope descriptor** to guide wallet user experience (e.g., show “Reveal” only to the owner).
```

Second instance i've seen of `User Experience` - is there a reason these words together are capitalized?

sips/sip-038/sip-enft-private-layer.md
```diff
@@ -0,0 +1,261 @@
...
+## Goals
+
+- Define a reusable trait for commitment-based private metadata for eNFTs.
+- Make private-layer patterns discoverable and composable for wallets/indexers.
+- Keep the standard minimal, additive, and backward-compatible.
+
+## Non-Goals
+
+- Full on-chain confidentiality (not achievable in a public read-only model).
+- Mandating a specific storage network, encryption scheme, or reveal flow.
```
wileyj, line 57: 🤔 i think i disagree with this non-goal as it pertains to advocating for a standard. 

Can you explain more your reasoning why an encryption scheme should not be standardized here? is it acceptable that different eNFT's may all choose different and competing encryption schemes?

if yes, I think there needs to be some additional text on what that would mean in practice, and how it may impact users/devs who would adhere to this proposed trait standard.  for example, would a potential api/wallet implementing this standard need to support some random collection of cryptography libraries?

sips/sip-038/sip-enft-private-layer.md
```diff
@@ -0,0 +1,261 @@
...
+# Specification
+
+## 1. Terminology
+
+- **Payload**: The private off-chain data (JSON, image, encrypted blob, etc.).
+- **Ciphertext**: An encrypted payload (still considered “payload bytes” for commitment purposes).
+- **Commitment**: A hash over the payload bytes (often ciphertext bytes) recorded on-chain.
+- **Envelope**: A small off-chain descriptor used for UX/access (e.g., retrieval URI, encrypted key blob, signed URL descriptor).
+- **Reveal**: Off-chain delivery of payload/envelope to the user; the client verifies integrity by comparing hash(payload_bytes) to the on-chain commitment.
+- **Algo ID**: An integer identifier that indicates how the commitment hash is computed.
```
wileyj, line 71: Unless i misunderstand the value for this key - it's simply an integer as an ID, correct?
I don't think `Algo-` adds any value here, and could be set solely as `ID`.

wileyj: note: if suggestion is accepted to rename to `ID` - be sure to update later references to `algo-id` etc like on line 78

sips/sip-038/sip-enft-private-layer.md
```diff
@@ -0,0 +1,261 @@
...
+- **Scheme Version**: A monotonically increasing version number for commitment/envelope encoding.
```
wileyj: Similar comment as above for ID- what does `Scheme` add to the definition that isn't implied by `Version`?

sips/sip-038/sip-enft-private-layer.md
```diff
@@ -0,0 +1,261 @@
...
+- This SIP does not mandate a single algorithm, but **recommends**:
+- `algo = u0` => SHA-256 over raw payload bytes
+
+### Algorithm Registry (Informative)
+
+| Algo ID | Algorithm | Notes |
+|---------|-----------|-------|
+| u0 | SHA-256 | Recommended default |
+| u1-u99 | Reserved | For future standardization |
+| u100+ | Application-defined | Projects may define custom algorithms |
```
wileyj, line 90: I think this answers my question above - using the ID, the intent is to use that to perform a lookup on which algorithm is used, correct?
Still, question remains about downstream application/users and if they would then be required to implement a varying number of crypto libraries.

sips/sip-038/sip-enft-private-layer.md
```diff
@@ -0,0 +1,261 @@
...
+```clarity
+(define-trait enft-commitment-trait
+  (
+    ;; Public: returns the commitment to the private payload (often ciphertext bytes)
+    (get-private-commitment (uint)
+      (response
+        (tuple
+          (commitment (buff 32))                 ;; hash of payload bytes
+          (algo uint)                            ;; algorithm id (e.g., u0 = sha256)
+          (mime (optional (string-ascii 64)))    ;; e.g. "application/json", "image/png"
```
wileyj, line 119: Are there any restrictions to MIME types here? in theory, could the MIME type be an executable?

sips/sip-038/sip-enft-private-layer.md
```diff
@@ -0,0 +1,261 @@
...
+- Wallets/indexers should treat the presence of `enft-owner-envelope-trait` as an optional capability.
+
+---
+
+### 3.3 Error Codes (Recommended)
+
+Implementations **may** choose their own error codes. This SIP **recommends**:
+
+- `err u404` => token-id not found / no commitment set
+- `err u100` => not authorized for `get-owner-envelope` (only relevant if `enft-owner-envelope-trait` is implemented)
```
wileyj: if above not found is using http status code, why not use `u401` here?

sips/sip-038/sip-enft-private-layer.md
```diff
@@ -0,0 +1,261 @@
...
+- **Scheme Version**: A monotonically increasing version number for commitment/envelope encoding.
+
+### 2. Commitment Semantics
+
+Implementations **must** expose a commitment for a given token-id.
+
+- The commitment **must** be computed exactly according to the declared algorithm id.
+- The commitment **must** be stable: the same payload bytes must produce the same commitment.
+- This SIP does not mandate a single algorithm, but **recommends**:
+- `algo = u0` => SHA-256 over raw payload bytes
```
wileyj, line 82: what is `token-id`?

I think it would help to define the schemas you propose first, with all possible keys

sips/sip-038/sip-enft-private-layer.md
```diff
@@ -0,0 +1,261 @@
...
+
+If the envelope or payload must remain private, it **should** be protected off-chain via:
+- encryption (e.g., encrypt payload/key per owner), and/or
+- access control (signed challenges, bearer tokens, short-TTL URLs).
+
+### Integrity and server non-equivocation
+Assuming collision-resistant hash functions, the on-chain commitment ensures the server cannot swap the underlying private payload after minting without being detected. Clients **must** verify `hash(payload_bytes) == commitment` after reveal.
+
+### Phishing / malicious URIs
+Wallets should treat envelope URIs as untrusted input and apply standard URL safety practices (origin warnings, link previews sandboxing, and user confirmation flows where appropriate).
```
wileyj, line 219: 👍

sips/sip-038/sip-enft-private-layer.md
```diff
@@ -0,0 +1,261 @@
...
+1) A reference trait contract implementing this specification is deployed on Stacks mainnet.
+
+2) At least one production NFT contract on mainnet implements the trait(s) exactly as specified in this SIP.
+
+3) At least one ecosystem integration — indexer, wallet, or marketplace — demonstrates consumption of the interface, 
+   proving interoperability independent of the authors.
+
+A trait that follows this specification will be made available 
+on mainnet upon ratification, and its deployed address will be 
+recorded here.
```
wileyj, line 261:
```suggestion
A trait that follows this specification will be made available 
on mainnet upon status update to activation-in-progress, and its deployed address will be 
recorded here.
```

sips/sip-038/sip-enft-private-layer.md
proofofgame: Accepted - will add Layer: Trait to preamble.

sips/sip-038/sip-enft-private-layer.md
proofofgame: will apply suggested formatting

sips/sip-038/sip-enft-private-layer.md
proofofgame: Good catch, no reason for capitalization

sips/sip-038/sip-enft-private-layer.md
proofofgame: Good point, I agree this needs clearer justification in the spec.

The SIP standardizes interoperability at the contract boundary: wallets and indexers need a stable way to (1) discover that a token has a commitment-based private layer, (2) read the commitment hash, and (3) optionally obtain an owner-gated envelope for a consistent "Reveal" UX. The trait gives that stable interface. Wallets never decrypt anything - they only verify `hash(payload) == on-chain commitment`. Decryption happens off-chain between sender and owner, outside the trait scope.

We avoided mandating a single encryption scheme because:

- Different projects have fundamentally different threat models (server-assisted reveal vs. client-side decryption with embedded key blob vs. proxy re-encryption, etc.)
- Schemes will evolve, and a standards-level breaking change each time crypto improves would be worse than flexibility
- Standardizing too early risks ossifying an implementation before the ecosystem converges


That said, I fully agree this cannot mean "anything goes" with no guardrails. To address fragmentation risk, we'll strengthen the spec by:

1. Making algo/version fields **normative** - implementers using any non-default algorithm id **must** publish a stable, publicly available algorithm specification
2. Expanding the Algorithm Registry with recommended baseline ids and explicit guidance that u0 (SHA-256) is the interoperability default
3. Adding wallet guidance: clients may support only u0 and treat unknown algorithm ids as unsupported for verification while still supporting discovery (similar to how browsers handle unknown MIME types)

In short: one trait, one UX surface, with algorithm/version identifiers to keep multiple schemes composable without breaking wallet support. A future SIP could standardize encryption profiles if the ecosystem converges on common patterns.

sips/sip-038/sip-enft-private-layer.md
proofofgame: I see the point about simplicity. My concern is that bare `ID `is ambiguous - it could be confused with` token-id `in context. Would `hash-id` work as a compromise? Open to your preference

sips/sip-038/sip-enft-private-layer.md
proofofgame: Accepted - will simplify to `Version` and clarify in terminology that it refers to the commitment/envelope encoding scheme.

sips/sip-038/sip-enft-private-layer.md
proofofgame: Downstream applications do not need to support all algorithms. SHA-256 (`u0`) is the recommended default. A wallet can support only `u0` and display a warning or reject tokens with unrecognized algorithm IDs - similar to how browsers handle unknown MIME types. The registry exists so the ecosystem can converge organically without requiring a new SIP for each algorithm. I'll add guidance text making this explicit

sips/sip-038/sip-enft-private-layer.md
proofofgame: Good security question. The `mime` field is informational only - it describes payload content, it does not instruct the client to execute anything. However, I'll add to Security Considerations: "Clients should validate MIME types and should reject executable types (e.g. `application/x-executable`) by default." This follows standard content-security practices.

sips/sip-038/sip-enft-private-layer.md
proofofgame: Agreed - `u401` is consistent with the HTTP-inspired convention already used for `u404`. Will update

sips/sip-038/sip-enft-private-layer.md
proofofgame: Will add `token-id` to Terminology section (the SIP-009 NFT identifier, type `uint`). Will also restructure so data schemas appear before semantic rules

sips/sip-038/sip-enft-private-layer.md
proofofgame: Accepted - will use "activation-in-progress" instead of "ratification" as suggested.

proofofgame:
Thanks a lot - appreciate it. @314159265359879 

You were right: I was still finishing a few wileyj comments. Those are now addressed and pushed, and I’m resolving the completed threads.

When you get a chance, please re-check and update the review state if everything looks good. Thank you.

sips/sip-038/sip-enft-private-layer.md
wileyj: thanks! that addition seems reasonable.

sips/sip-038/sip-enft-private-layer.md
wileyj: Thanks for the update - this makes more sense; i'll re-read the spec to see it's more clear. Resolving

sips/sip-038/sip-enft-private-layer.md
wileyj: it's definitely debatable and entirely subject to preference - i'll resolve this based on the `token-id` comment (which does make `algo-id` etc more apt in the full context). 

my preference is always KISS, but i will ack that can lead to ambiguity.

proofofgame:
Hi all, just a quick follow-up on this PR.

I believe the requested changes have now been addressed in the latest commits. When you have a moment, could you please re-check the current version and update the review state if it looks good for editor sign-off?

Thanks again

sips/sip-038/sip-enft-private-layer.md
```diff
@@ -0,0 +1,270 @@
+# SIP-038: Standard Trait Definition for Commitment-Based Private Metadata (Encrypted NFTs)
+
+## Preamble
+
+SIP Number: 038  
+Title: Standard Trait Definition for Commitment-Based Private Metadata (Encrypted NFTs)   
+Authors: Xenitron nortinex@gmail.com, Bogachev ogres2009anton@gmail.com  
+Status: Draft
```
clairetopalian:
```suggestion
Status: Accepted   
```

314159265359879 (approved):
Sorry for the delay, thank you for making such quick work of addressing the comments. 

Consider this the editor sign-off.

clairetopalian (approved):

314159265359879:
Lets merge this SIP @clairetopalian?

jcnelson:
Why was this merged? It hasn't passed CAB approval yet

clairetopalian:
Apologies, this was merged in error. I've opened a PR to reset the status to Draft until CAB approval is complete: https://github.com/stacksgov/sips/pull/269

proofofgame:
Hi all — a quick update from the SIP-038 authors.

SIP-038 is currently in `main` as Draft after #250 was merged and the status
was subsequently reset pending CAB approval in #269.

Since the original editor review, we have continued working with the eNFT
model in production and have done a much deeper security/cryptographic review
of the proposed standard.

The core motivation has not changed: SIP-038 remains an additive extension to
SIP-009 for interoperable private/encrypted NFT payloads.

However, we have identified several material improvements that we believe
should be incorporated before CAB consideration:

- clarify that public Clarity state cannot provide confidential
  "owner-gated" access; confidentiality must come from cryptography, so the
  envelope model should be framed around owner-authorized issuance rather than
  on-chain access control;

- bind payloads to an exact canonical context and commit to a canonical
  manifest, including version and cipher-suite semantics, rather than leaving
  raw-payload commitment/canonicalization application-defined;

- define interoperable encrypted-envelope semantics while keeping recipient
  key discovery and re-wrapping infrastructure outside the core SIP;

- strengthen Security Considerations around exact-token authorization,
  capability persistence after transfer, low-entropy private payloads, and
  authorization failures at the envelope-issuance layer.

This revision is informed in part by production experience with owner-private
NFT gameplay, including a historical exact-token authorization failure that
helped us make the threat model much more precise.

We are preparing the revised text together with an exact Clarity trait,
canonical serialization rules, a reproducible test vector, and a reference
implementation before asking for renewed review.

Since #250 is already merged, our understanding is that the clean path is to
submit a new PR against `main` updating the existing SIP-038 Draft, and then
request a fresh editor/CAB review of that revision.

@wileyj @314159265359879 @jcnelson — please let us know if you would prefer a
different process before we open that revision PR.

Thanks again for the extensive review work on the original draft.

PR #257, opened 2026-02-06
SIP-040: SIP for improved post-conditions

brice-stacks:
This SIP proposes two new post-condition improvements, `Originator` mode, and a `MAY SEND` condition for NFTs.


sips/sip-post-conds/sip-post-conds.md
```diff
@@ -0,0 +1,186 @@
+# Preamble
+
+SIP Number: TBD
```
wileyj: Proposing SIP-040
```suggestion
SIP Number: 040
```

sips/sip-040/sip-040-post-conds.md
```diff
@@ -0,0 +1,186 @@
...
+
+- Brice Dobry <brice@stackslabs.com>
+
+Status: Draft
+
+Consideration: Governance, Technical
+
+Type: Consensus
+
+Layer: Consensus
```
wileyj:
```suggestion
Layer: Consensus (hard fork)
```

sips/sip-040/sip-040-post-conds.md
```diff
@@ -0,0 +1,186 @@
...
+
+License: BSD-2-Clause
+
+Sign-off:
+
+Discussions-To:
+
+- https://forum.stacks.org/t/improved-post-conditions/18661
+
+# Abstract
```
wileyj, line 29: Is it worth mentioning that this sip is a rider for SIP-039, and if that sip is approved this one follows the same activation?

or is this SIP intended to activate separately?

sips/sip-040/sip-040-post-conds.md
```diff
@@ -0,0 +1,186 @@
...
+`Allow` mode. The reality is that some applications, e.g. DeFi apps, move assets
+between a variety of contracts as part of one contract call. The movements may
+be dynamic and difficult or impossible to predict in advance, so users end up
+reverting to signing `Allow` mode transactions, just to make it work. This is a
+very bad habit to teach our users, and we can do better. What the user really
+cares about is something like, "When I call contract X, at most `m` STX and `n`
+USDCx may be moved from my wallet." They don't care if contract X sends sBTC to
+contract Y, which sends USDCx to contract Z, etc. `Originator` mode allows the
+user to specify only the restrictions on their own assets, allowing any
+movements of assets amongst other principals.
```
wileyj:
```suggestion
reverting to signing `Allow` mode transactions, just to make it work. What users typically
prefer is something like, "When I call contract X, at most `m` STX and `n`
USDCx may be moved from my wallet.", and less so if contract X sends sBTC to
contract Y, which sends USDCx to contract Z, etc. `Originator` mode allows the
user to specify only the restrictions on their own assets, allowing any
movements of assets amongst other principals.
```

I hope i kept the intent in this section, but i felt it was important to not ascribe emotion to what users may or may not want.

sips/sip-040/sip-040-post-conds.md
```diff
@@ -0,0 +1,186 @@
...
+movement upon completion of the transaction. It is always satisfied regardless
+of whether the specified NFT is sent, and it is treated as covering that
+specific NFT instance for the specified principal when enforcing allowlist
+coverage in `Deny` or `Originator` mode. In `Allow` mode, it is redundant but
+valid.
+
+The transaction encoding reserves 1-byte for the non-fungible condition code.
+`MAY SEND` adds a new acceptable value for this byte, `0x12`.
+
+## Updates to the SIP-005 Specification
```
wileyj, line 115: hmmm - i wonder if we need a way to differentiate when a SIP replaces an older one -as well as- updates an existing spec. 🤔

sips/sip-040/sip-040-post-conds.md
```diff
@@ -0,0 +1,186 @@
...
+ Post-conditions are defined in terms of which assets each account sends or
+ does not send during the transaction's execution.  To enforce post-conditions,
+```
+
+# Related Work
+
+- This SIP adds to the existing specification for post-conditions in
+  [SIP-005](../sip-005/sip-005-blocks-and-transactions.md)
+- This SIP will activate in epoch 3.4 together with
+  [SIP-clarity5](../sip-clarity5/sip-clarity5.md)
```
wileyj: Please update with SIP numbers (039).

sips/sip-040/sip-040-post-conds.md
brice-stacks: I have this sentence in the activation section:

> If approved, it should go live together with Clarity 5 in epoch 3.4.

I was trying to avoid linking to the other SIP, since neither is merged yet, but could do so if sip-039 gets merged.

sips/sip-040/sip-040-post-conds.md
brice-stacks: Yeah, I wasn't quite sure about the best way to do this. One option is to replicate the entirety of SIP-005 here, with the required edits. Another option might be to call this SIP-005-1 and include the full body with edits, plus a section that explains the differences from the previous version -- I kinda like that option.

sips/sip-040/sip-040-post-conds.md
brice-stacks: Updated, but the link won't work until that one is merged.

sips/sip-040/sip-040-post-conds.md
brice-stacks: I tweaked it a bit differently. Let me know if you think that works.

sips/sip-040/sip-040-post-conds.md
brice-stacks: I went ahead and added it, with a currently broken link.

wileyj: thanks! i think only a mention is fine here, we can't expect links to unmerged sips to resolve.
but, it has been assigned a sip number so it's searchable in prs (and when merged with be easier to find).

sips/sip-040/sip-040-post-conds.md
wileyj: prefer not go the sip-005-1 route. i think this will be fine, but we should consider how to address this uniformly in the future.

sips/sip-040/sip-040-post-conds.md
wileyj: i think it's fine to only mention the sip number, since it's been assigned. same comment as above : we can't expect links to resolve until merged, so it doesn't make sense to require a link that won't resolve.

sips/sip-040/sip-040-post-conds.md
brice-stacks: 9254c01

sips/sip-040/sip-040-post-conds.md
brice-stacks: 9254c01

sips/sip-040/sip-040-post-conds.md
wileyj: better - only nit is this line in the updated text: `This is a very bad habit to teach our users, and we can do better.`. objectively, you're correct but it's also subjective. it's not a hill i'm willing to die on though.

brice-stacks: > objectively, you're correct but it's also subjective.

🤣 I want to argue that "This is a very bad habit to teach our users, and we can do better," is objectively true and kinda want to leave it.

sips/sip-040/sip-040-post-conds.md
```diff
@@ -0,0 +1,183 @@
...
+In the current ecosystem, `Deny` mode is the safest way to interact with smart
+contracts, and users should be in the habit of never signing a transaction in
+`Allow` mode. The reality is that some applications, e.g. DeFi apps, move assets
+between a variety of contracts as part of one contract call. The movements may
+be dynamic and difficult or impossible to predict in advance, so users end up
+reverting to signing `Allow` mode transactions, just to make it work. This is a
+very bad habit to teach our users, and we can do better. The concern of the user
+is typically about assets moving out of their own wallet, and they are not
+concerned about assets moving in and out of contracts, triggered by their call.
+Contracts that they do care about should be protected internally, using the
```
wileyj:
```suggestion
Contracts that users do care about should be protected internally, using the
```

sips/sip-040/sip-040-post-conds.md
```diff
@@ -0,0 +1,183 @@
...
+
+# Introduction
+
+In the current ecosystem, `Deny` mode is the safest way to interact with smart
+contracts, and users should be in the habit of never signing a transaction in
+`Allow` mode. The reality is that some applications, e.g. DeFi apps, move assets
+between a variety of contracts as part of one contract call. The movements may
+be dynamic and difficult or impossible to predict in advance, so users end up
+reverting to signing `Allow` mode transactions, just to make it work. This is a
+very bad habit to teach our users, and we can do better. The concern of the user
```
wileyj: I still feel this line is very subjective and emotionally driven. what about a change like this:

```suggestion
reverting to signing `Allow` mode transactions, just to make it work, which erodes the security habits we want users to build. The concern of the user
```

wileyj: I think the original text is a bit finger-pointy, that's why it keeps drawing my attention to it.

sips/sip-040/sip-040-post-conds.md
```diff
@@ -0,0 +1,183 @@
...
+
+- This SIP adds to the existing specification for post-conditions in SIP-005
+- This SIP will activate in epoch 3.4 together with SIP-039
+
+# Backwards Compatibility
+
+These new post-condition additions are backwards compatible for transactions
+that do not use the new mode or condition code. Nodes need to be updated to
+accept transactions with these new features, but existing SIP-005
+post-conditions will continue to operate as they do now. Pre-upgrade nodes will
```
wileyj: should `Pre-upgrade` be changed to something like `Prior to activation`?

sips/sip-040/sip-040-post-conds.md
```diff
@@ -0,0 +1,183 @@
...
+
+SIP Number: 040
+
+Title: Improved Post-Conditions
+
+Author(s):
+
+- Brice Dobry <brice@stackslabs.com>
+
+Status: Draft
```
wileyj:
```suggestion
Status: Accepted
```

sips/sip-040/sip-040-post-conds.md
```diff
@@ -0,0 +1,183 @@
...
+
+SIP Number: 040
+
+Title: Improved Post-Conditions
+
+Author(s):
+
+- Brice Dobry <brice@stackslabs.com>
+
+Status: Accepted
```
wileyj:
```suggestion
Status: Activation-In-Progress
```

sips/sip-040/sip-040-post-conds.md
brice-stacks: I wanted to be finger pointy 😅. I'm good with this proposal though.

sips/sip-040/sip-040-post-conds.md
brice-stacks: yeah, that sentence is a little weird. I'm not entirely sure what point I was trying to make originally. Your suggestion sounds good to me.

314159265359879 (approved):
Lets merge this together with SIP039 now that is voted in and live.

clairetopalian (approved):

jcnelson:
@obycode Can you open a PR to switch the activation status to `Ratified` and add a link to it in the `README.md`?  Thanks! :pray:

PR #266, opened 2026-04-14
SIP-043 Identicons for contracts

friedger:
This SIP describes how contracts can be identified by visual identifiers. It used `(clarity-hash? current-contract)` for minidenticons as used in source-of-clarity: 

image

### DIFF
--- a/sips/sip-xxx/sip-xxx-identicon-for-contracts.md
+++ b/sips/sip-xxx/sip-xxx-identicon-for-contracts.md
@@ -30,7 +30,7 @@ Discussions-To:
 
 This SIP specifies a deterministic way to derive a visual identifier (an "identicon") for any Clarity smart contract deployed to the Stacks blockchain. The identicon is a pure function of the contract's source code after canonicalization: two deployments of the same source — on any address, at any height, on any network — render the same icon. Wallets, explorers, and apps that implement the specification allow users to recognize known code at a glance and to notice when contracts that share a name or author address do not share code.
 
-The specification pins three things: the canonical form of the source (the output of `clarinet fmt` with default settings), the hash function (`SHA-512/256` over the UTF-8 bytes of the canonical form), and the rendering library and configuration (`minidenticons`, default options, seeded with the lowercase hex-encoded hash). The output is an SVG suitable for inline rendering.
+The specification pins three things: the canonical form of the source (the output of `clarinet format` with default settings), the hash function (`SHA-512/256` over the UTF-8 bytes of the canonical form), and the rendering library and configuration (`minidenticons`, default options, seeded with the lowercase hex-encoded hash). The output is an SVG suitable for inline rendering.
 
 # Copyright
 
@@ -48,16 +48,16 @@ This SIP proposes that consensus. It does not propose any consensus change to th
 
 ## 1. Canonical source
 
-The canonical form of a contract's source code is the byte-for-byte output of `clarinet fmt` invoked with default settings, against the contract's `.clar` source file.
+The canonical form of a contract's source code is the byte-for-byte output of `clarinet format` invoked with default settings, against the contract's `.clar` source file.
 
 - Formatters MUST normalize line endings to `\n` (LF).
 - Formatters MUST produce a final trailing `\n`.
-- Comments (`;;` and `;;;`) are part of the source and are preserved by `clarinet fmt`. They are hashed.
+- Comments (`;;` and `;;;`) are part of the source and are preserved by `clarinet format`. They are hashed.
 - The canonical form is UTF-8 encoded.
 
-Implementations that do not have access to a Clarinet formatter (for example, browser-only apps fetching source from `/v2/contracts/source`) SHOULD hash the source as returned by the node, assuming the contract author deployed `clarinet fmt`-formatted source. Implementations MAY display a "source not canonicalized" warning next to the identicon when the returned source deviates from a heuristic check (trailing whitespace, mixed line endings, tabs mixed with spaces).
+Implementations that do not have access to a Clarinet formatter (for example, browser-only apps fetching source from `/v2/contracts/source`) SHOULD hash the source as returned by the node, assuming the contract author deployed `clarinet format`-formatted source. Implementations MAY display a "source not canonicalized" warning next to the identicon when the returned source deviates from a heuristic check (trailing whitespace, mixed line endings, tabs mixed with spaces).
 
-Contract authors who wish to participate in the identicon convention MUST run `clarinet fmt` before deploying. Two authors deploying identical logic with different whitespace will produce different identicons.
+Contract authors who wish to participate in the identicon convention MUST run `clarinet format` before deploying. Two authors deploying identical logic with different whitespace will produce different identicons.
 
 ## 2. Hash
 
@@ -120,13 +120,7 @@ This SIP activates once 10 ecosystem participants who display a set of contracts
 
 ## TypeScript / browser
 
-A reference implementation lives in this repository at `src/utils/contractIdenticon.ts` and is wired into the wallet selector at `src/components/wallet-selector/WalletCard.tsx`. The module:
-
-1. fetches contract source via `GET /v2/contracts/source/{address}/{name}` against the appropriate Hiro API host for the network;
-2. hashes the returned source with `sha512_256` from `@noble/hashes/sha2`;
-3. renders via `minidenticonSvg` from the `minidenticons` package.
-
-Canonicalization via `clarinet fmt` is not performed in the browser implementation; see §1 on assumptions and heuristic warnings.
+TODO
 
 ## Clarity (self-declaration)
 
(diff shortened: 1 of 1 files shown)


sips/sip-xxx/sip-xxx-identicon-for-contracts.md
```diff
@@ -0,0 +1,132 @@
...
+
+# Activation
+
+This SIP activates once 10 ecosystem participants who display a set of contracts have implemented this specification.
+
+# Reference Implementations
+
+## TypeScript / browser
+
+TODO
```
314159265359879: SIP-000 §5 requires reference implementations be "complete, functional code" or explicitly marked as "forthcoming with timeline." Replace with a single sentence: JavaScript/TypeScript: Forthcoming (target: May 2026).

sips/sip-xxx/sip-xxx-identicon-for-contracts.md
```diff
@@ -0,0 +1,132 @@
...
+## 1. Canonical source
+
+The canonical form of a contract's source code is the byte-for-byte output of `clarinet format` invoked with default settings, against the contract's `.clar` source file.
+
+- Formatters MUST normalize line endings to `\n` (LF).
+- Formatters MUST produce a final trailing `\n`.
+- Comments (`;;` and `;;;`) are part of the source and are preserved by `clarinet format`. They are hashed.
+- The canonical form is UTF-8 encoded.
+
+Implementations that do not have access to a Clarinet formatter (for example, browser-only apps fetching source from `/v2/contracts/source`) SHOULD hash the source as returned by the node, assuming the contract author deployed `clarinet format`-formatted source. Implementations MAY display a "source not canonicalized" warning next to the identicon when the returned source deviates from a heuristic check (trailing whitespace, mixed line endings, tabs mixed with spaces).
```
314159265359879: "Implementations that do not have access to a Clarinet formatter (for example, browser-only apps fetching source from /v2/contracts/source) SHOULD hash the source as returned by the node, assuming the contract author deployed clarinet format-formatted source."

Issue: Circular assumption. 

Rewrite to separate the normative requirement from the implementation constraint:

Suggestion 1: Implementations without access to a local Clarinet formatter SHOULD attempt to retrieve pre-formatted source and SHOULD emit a warning if the source deviates from the canonical heuristic (trailing whitespace, line-ending consistency, tab usage).

or

Suggestion 2: Implementations MAY skip canonicalization and hash source as-received from nodes, with the caveat that non-canonical source will produce identicons that diverge from the contract author's reference rendering.


Perhaps this is already obvious to the author, to clarify for other readers:
The result of suggestion 1: Implementation knows what to do even when it detects non-canonical source (warn the user, don't silently produce divergent identicons).
The result if going with suggestion 2: No hidden assumption. Implementation makes an explicit choice with known consequences.

Both break the circularity by acknowledging: I don't know if the source is canonical, and here's what I do about it.

sips/sip-xxx/sip-xxx-identicon-for-contracts.md
```diff
@@ -0,0 +1,132 @@
...
+- When passed to the renderer, the hash is lowercase hex-encoded without a `0x` prefix (64 characters).
+
+This is the same hash exposed in Clarity as the `sha512/256` function, so a contract MAY compute and expose its own identicon hash on-chain:
+
+```clarity
+(define-read-only (get-identicon-hash (source (buff 65535)))
+  (sha512/256 source))
+```
+
+Off-chain implementations MUST NOT rely on any on-chain value; the hash is always derived from the deployed source.
```
314159265359879: > Off-chain implementations MUST NOT rely on any on-chain value; the hash is always derived from the deployed source.

I think this is trying to say something like: "Don't use off-chain copy of the contract source or a cached hash that some contract or indexer published on-chain; always recompute from source."
Lets make sure everyone understands "deployed source" to mean the source as it was deployed and not the source after being deployed (the latter can still be changed, the prior is immutable). 

Suggestion 1:

Off-chain implementations MUST compute the identicon hash directly from the canonical source code fetched via /v2/contracts/source, not from any contract copy or on-chain-published or cached hash value. The source, once deployed, is immutable and authoritative.

or 

Suggestion 2: (simpler more direct)

Do not trust on-chain hash declarations. Always recompute the identicon hash from the source code returned by the node.

sips/sip-xxx/sip-xxx-identicon-for-contracts.md
```diff
@@ -0,0 +1,132 @@
...
+
+This SIP is made available under the terms of the Creative Commons CC0 1.0 Universal license, available at https://creativecommons.org/publicdomain/zero/1.0/.
+
+# Introduction
+
+Clarity contracts are addressed by `{deployer}.{name}`. That pair is useful for on-chain references but is a poor user-facing identifier: a user reading `SP2J….transfer-helper` cannot tell from the name alone whether this is the audited, widely-deployed helper they have used before, a fork with a subtle change, or a phishing lookalike deployed to a confusable address.
+
+Hash-based identicons are a well-understood solution to this problem (see `identicon`, `jdenticon`, `blockies`, the GitHub default avatar). What is missing on Stacks is a consensus on *which* bytes get hashed and *which* renderer is used, so that the same contract produces the same icon everywhere it is shown.
+
+This SIP proposes that consensus. It does not propose any consensus change to the Stacks blockchain — the specification is an off-chain convention followed by wallets, explorers, and apps.
```
314159265359879: "This SIP proposes that consensus"

Issue: imprecise use of ocnsensus, conventions are not consensus in the blockchain sense. 

Replace with:
This SIP proposes adoption of this convention as a standard across the Stacks ecosystem.

sips/sip-xxx/sip-xxx-identicon-for-contracts.md
```diff
@@ -0,0 +1,132 @@
...
+
+The identicon is rendered with the **`minidenticons`** library, used in its default configuration:
+
+- Library: `minidenticons` (https://github.com/laurentpayot/minidenticons), version 4.x or later.
+- Function: `minidenticonSvg(seed, saturation, lightness)`.
+- Arguments:
+  - `seed`: the lowercase hex-encoded identicon hash from §2.
+  - `saturation`: default (`50`).
+  - `lightness`: default (`50`).
+- Output: an SVG string, 5×5 symmetric grid.
```
314159265359879: Do you want to specify the dimensions or internal grid structure?

Suggestion
Output: an SVG element (default viewBox) rendering a 5×5 symmetric pixel grid."

sips/sip-xxx/sip-xxx-identicon-for-contracts.md
```diff
@@ -0,0 +1,132 @@
...
+- **Jdenticon**: another open-source identicon library. Produces ~874k distinct icons (vs. ~295k for minidenticons). A future revision of this SIP MAY switch renderers; the hash specification is independent of the renderer.
+- **Sourcify**: verifies EVM contract source against deployed bytecode. Useful precedent for canonical-source conventions. This SIP intentionally skips the verification step because Stacks deploys source directly, not bytecode.
+
+# Backwards Compatibility
+
+This is a new off-chain convention. There is no on-chain consensus change and no breaking impact on existing contracts, wallets, or explorers. Implementations that do not adopt the SIP continue to display contracts as before.
+
+# Activation
+
+This SIP activates once 10 ecosystem participants who display a set of contracts have implemented this specification.
```
314159265359879: Minor change suggestion here:

Once 10 ecosystem participants who display a set of contracts have implemented this specification" is vague. Does "display a set of contracts" mean display any contracts, or display this SIP's identicons? 

Tighten to: 
Once 10 ecosystem participants (wallets, explorers, developer tools or dApps) have publicly implemented and deployed this specification in production.

314159265359879 (changes requested):
Thanks for creating this SIP @friedger this sounds convenient way to verify contract code copies. This is already of high quality and almost ready for editor sign-off. 

I have added some in line editorial comments when I'd like something rewritten I do my best to offer concrete suggestions to choose from or use for your own rewrite. 

I would like to suggest one important design clarification which can be added as a separate section.

------
Design clarification: code identity vs principal identity

The code identicon is advisory and supplements the contract principal. To prevent visual conflation of code identity with address identity, implementations SHOULD:

* Display the code identicon adjacent to the contract principal ({deployer}.{name}) with a clear visual hierarchy or label indicating it represents source code, not contract address.
* Optionally display a separate principal-based identicon (derived from SHA-256(principal)) to make address divergence visually apparent. Two deployments of identical code to different principals would then show: identical code icons, distinct address icons.

This dual approach prevents users from assuming 'same icon = same contract' when they may be interacting with different deployments.

friedger:
Thank you for the feedback. Comments have been incorporated @314159265359879

314159265359879 (approved):
Thanks for addressing the comments. Editor sign-off ✅ 

Can you add the SIP number SIP-043?

Then we can merge as `activation in progress`.

aryzing:
This is great

PR #260, opened 2026-02-11
Use a table for SIPS (draft -> ratified)

wileyj:
Need to verify the links and activation status, but this is currently up to date and gives a nicer 1-pager of sips. 

rather than the standard list of ratified sips - this change adds a markdown table of all sips that have been accepted and a number assigned (no matter the status). 

the caveat is this will need to be updated regularly by sip editors (etc) so it remains current.

PR #241, opened 2025-10-22
SIP-000-1: SIP Process SIP-000 Upgrade

Hero-Gamer:
**SIP-000-1** packs a number of refinement of the SIP Process we've learned over the past few years. 

Below is a "**Change log**" of SIP-000-1 
1. Update Abstract to mention this replaces SIP-000 and the rationale
2. Added “## Backwards Compatibility” but it is NA
3. In "## SIP Format", reordered some fields sequence according to SIP_Template.md 
4. Moved 2 ## SIP Flow subsections together, renamed the first to SIP Status
5. Moved “## SIP Types”  to logical order, after “## Layer”
6. New subsection “## SIP Layers” created by moving the explanations from Preambles
7. Preambles’ spec’s ordering of fields reordered to match the SIP_TEMPLATE
8. Added “### SIP Authoring Rules”
9. Added “### Expectation of Minimum Timelines And Rules”
10. Added “### SIP Versioning and Replacement Update Policy”
11. Added “### Documentation, Repo Maintancne, Merging into Main” 

* This PR should serve as a starting point of SIP-000 improvements we want to add.
* Some great ideas from @whoabuddy in this PR too we can include: https://github.com/stacksgov/sips/issues/145
So please welcome to make changes and add/remove suggestions!

**Some unanswered questions:**
* **Hard Fork SIPs code freeze rules** - Define process for code freeze and creation of dedicated release branches. - Only milestone changes for that release may be merged into the fork branch.

* **SIP-000 folder naming**: I tried both folder naming ideas, whether it's "_1" or "-1" (see images below), might look both little odd and confusing, I'd actually perhaps propose keeping the SIP-000 folder naming as is. Only the the MD files get updated according to the SIP-000 versioning policy, and filed correctly within.

* **Rider SIPs rules of engagement:** How Rider SIPs should be handled approval wise. And documentation wise.  

Screenshot 2025-10-21 at 23 19 52
Screenshot 2025-10-21 at 23 24 02

Issue #228, opened 2025-09-25
SIP Update Policy


Hero-Gamer:
Hi all, 

Putting in a conversation to confirm the updating/replacing method for existing SIP. So we can ahead making proposal for updating SIP-000.

**2 main ideas on the table so far:**

1. **Replacing a SIP by a new SIP, retiring the old.** (Matches currently method indicated in SIP-000)
- Use a new SIP allocation to replace the existing SIP that's ratified already.
- Example: If we want to update SIP-000: SIP process, we would submit SIP-035 and that will be the new SIP process SIP, we would retire SIP-000.
- Note: For this is the approach taken by BIP Editor. https://github.com/bitcoin/bips/pull/1820 - they are replacing BIP-0002 old process with BIP-0003 new process. (I reached out to an active BIP Editor Jon https://github.com/jonatack - to confirm their approach. He said "Yes, BIP2 remains and keeps its number and filename. Only a couple of its headers are updated.BIP2 Status would be updated from Active to Replaced and a header “Superseded-By: 3” would be added")

2. **Keeping same SIP number. Archive the previous version.** (new idea)
- Example: If SIP-000 were to be updated, we would just update the doc, ensure there is a history of the previous version saved in the archive folder (e.g. _**sip-000-stacks-improvement-proposal-process_V1.md**_ or could be just the date created _**sip-000-stacks-improvement-proposal-process_2020-06-23.md**_)
- Same would go to SIP-009 NFT if it were need updating, or a SIP-010 that need updating. Or a SIP-028 about sBTC federation, it can keep the same number just updating the document with our usual SIP process review & approval of course. 

**Context:** 
* As we are in the process of **updating SIP-000** based on all the learnings so far, and potentially updating SIP-009 NFT Standard (if I am not wrong majority of NFT contracts do not adhere to [SIP-009](https://github.com/stacksgov/sips/blob/main/sips/sip-009/sip-009-nft-standard.md) as Gamma's NFT contract is a slight variation on SIP-009.), and any other future updates to existing SIP. 

I asked my GPT for some thoughts you can see her: https://chatgpt.com/share/68d5904b-7558-800e-859c-5b373c30a412
Some food for thoughts: 

Method 1  - Replacement Model
✅ Advantages:
Preserves immutability — no ratified SIP is ever altered.
Matches Bitcoin precedent (familiar for cross-contributors).
Enables clear supersession metadata (Superseded-By).
Easier for tooling to index authoritative SIPs.

[...]

Issue #145, opened 2023-05-19
Documenting the SIP approval process


whoabuddy:
Now that we have more and more people looking at SIPs, we need a consistent workflow that's easy to use but adheres to SIP-000.

Since GitHub is the main medium we're using for the SIPs, I propose the following approach:

1. New SIP is created in `Draft` status via a PR
2. Once ready, request review from SIP Editors, who have at least one person with access to this GitHub repo that can be tagged (s/o to @rafaelcr)
3. SIP Editors will provide feedback until SIP is ready, then review/approve the PR, then make a commit to the PR adding their sign-off and updating the status to `Accepted`
4. Request reviews from relevant CABs, who also have at least one person with access (gov: @whoabuddy, tech: @obycode)
5. CAB will provide feedback and add minutes in a separate PR, then if approved, review/approve the PR, and add their sign-off. Last CAB to add sign-off also updates status to `Recommended`
6. Request review from Steering Committee (likely @jcnelson) and they can review/approve the PR, add their sign-off (is there a spot for this?), and update status to `Activation-in-Progress` and the README.

At that point we should probably merge it as it would update the README on the main branch, which shows both activation-in-progress and ratified SIPs.

Then a separate PR could be created (by SC, or maybe by others in the groups above) that changes the status to `Ratified` and updates the README once it's finalized.

Another way to look at this:

- a PR gets created for the SIP, which tracks it from `Draft` -> `Accepted` -> `Recommended` -> `Activation-in-Progress`
  - includes comments/feedback from SIP editors
  - includes approval/sign-off from CABs
  - includes approval/sign-off from SC
  - merged once it reaches `Activation-in-Progress`
- a PR gets created for each CABs meeting minutes
- a PR gets created to update the status from `Activation-in-Progress` to `Ratified`

Curious if others have thoughts on this, the SIPs below have parts of the process in place, and once we settle on a structure we can document it either through a SIP, somewhere in this repo as a doc, or both.

- #141 
- #131 
- #127

PR #273, opened 2026-07-10
Add CONTRIBUTING.md documenting working practices under SIP-000

314159265359879:
This adds a CONTRIBUTING.md that writes down how we work in this repo: how to submit a SIP, review etiquette, how the SIP-000 statuses map to GitHub steps, what can change after a CAB vote, when a SIP PR merges, and how a ratified SIP gets updated.

Nothing in here changes the SIP process. It documents practice under SIP-000 and says so explicitly, where the two would disagree SIP-000 wins. That is also why this is a repo doc and not a SIP: these are working rules we should be able to improve with a simple PR.

Context: this came out of https://github.com/stacksgov/sips/pull/241 and https://github.com/stacksgov/sips/issues/228. The useful operational ideas from #241 live here now, and the answer to #228 is the last section, the replacement model is already in SIP-000. The workflow mapping builds on what whoabuddy wrote up in #145. The numbering rules and the Withdrawn Draft label come from the discussion in #260.

Feedback welcome, especially from editors and CAB members on whether this matches how we actually work.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,93 @@
+# Contributing to the SIPs Repository
+
+This document describes the working practices for this repository. It documents
+how we apply the SIP process defined in
+[SIP-000](./sips/sip-000/sip-000-stacks-improvement-proposal-process.md).
+It does not add to or override SIP-000. Where this document and SIP-000
+disagree, SIP-000 wins.
+
+These practices can be improved through a regular pull request to this file.
+
+## Submitting a SIP
+
+- Fork this repository and open your SIP as a pull request from your fork.
+  Do not create branches in this repository.
+- Follow the [SIP template](./sips/SIP_TEMPLATE.md) for structure and
+  preamble fields. Your SIP starts in Draft status and without a number.
+  A SIP Editor assigns the number when the SIP moves to Accepted.
+- Write in plain language. Marketing or promotional material does not belong
+  in the SIP text. Supporting material can be added as supplemental documents
+  per SIP-000.
+
+Before submitting, it helps to socialize your idea. A post on the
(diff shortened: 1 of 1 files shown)


314159265359879:
@brice-stacks and @rafaelcr do you think it is okay to approve and merge this. There are no new ideas in this but it tries to capture the unwritten rules and the outcomes of past conversations about how we work in this repo.

CONTRIBUTING.md
```diff
@@ -0,0 +1,99 @@
...
+4. **Activation-in-Progress**: the Steering Committee votes and adds their
+   sign-off. The SIP's own Activation section now governs what happens.
+5. **Ratified** or **Rejected**: the Steering Committee determines the outcome
+   from the Activation criteria and updates the status. For SIPs activated
+   by a vote, the results and how they were verified are recorded with the SIP.
+
+## SIP numbers and status labels
+
+- SIP numbers are assigned once and never reused, even if the SIP never
+  advances past Draft. This prevents one number from referring to two
```
brice-stacks, line 57: Above it says that the number is not assigned until it is accepted. That conflicts with this sentence, which says that the number is never reused, "even if the SIP never advances past Draft."

brice-stacks (commented):
This looks good. Thanks for putting it together! I just noticed some inconsistencies with the draft/number explanation.

CONTRIBUTING.md
314159265359879: good point, I have updated it to "...past Accepted"

314159265359879:
This PR also fixes the status of SIP-021, SIP-033 and SIP-034. The README already showed it correctly but the statuses in here were stale.

brice-stacks (approved):

clairetopalian (approved):

PR #275, opened 2026-07-22
Mark 044 and 045 as ratified

brice-stacks:
The second commit just changes formatting to the standard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,89 +2,91 @@
 
 The SIPs describe the design, implementation, and governance of the Stacks
 blockchain.  
-The SIP process ([SIP-000](./sips/sip-000/sip-000-stacks-improvement-proposal-process.md)) describes how to make a SIP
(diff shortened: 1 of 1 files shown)


brice-stacks:
We'll still also need an update to those SIPs to add the vote results. This just updates the README.

jcnelson:
Can you also transition the SIP-044 and SIP-045 documents to ratified hen you add the vote results?

brice-stacks:
Sure, should we just do that all together in one PR?

jcnelson:
That's fine with me if it's okay with you :) I just thought doing them all in the same PR might be easier, since updating the table and updating the docs are usually done at around the same time either way.

314159265359879 (approved):
Thanks Brice! LGTM

brice-stacks:
SIP activation status added in #276, #277, and #278.

clairetopalian (approved):

